PR #67, opened 2020-04-22
fix(api): omitempty integration responses fields

afiunelw:
Our API Server returns responses that can change 100%, this behavior is
making us omit a number of fields if they are empty so that we don't
present them to the end user.

There is a current bug that this change is fixing where if you try to
create a new Lacework Integration with the Go client, the request
generated will send the following body and the API Server will through a
500 Internal Server Error (in the k8s logs, Schema validation failure)
```json
{
  "NAME": "Azure Activity Log",
  "TYPE": "AZURE_AL_SEQ",
  "ENABLED": 1,
  "STATE": {
    "ok": false,
    "lastUpdatedTime": "",
    "lastSuccessfulTime": ""
  },
  "DATA": {
    "CREDENTIALS": {},
    "TENANT_ID": "ID",
    "QUEUE_URL": "URL"
  }
}
```

We are avoiding adding the `STATE` field if it is empty so that we send
this request body instead:
```json
{
  "NAME": "Azure Activity Log",
  "TYPE": "AZURE_AL_SEQ",
  "ENABLED": 1,
  "DATA": {
    "CREDENTIALS": {},
    "TENANT_ID": "ID",
    "QUEUE_URL": "URL"
  }
}
```

### DIFF
--- a/api/integrations_test.go
+++ b/api/integrations_test.go
@@ -19,6 +19,7 @@
 package api_test
 
 import (
+	"encoding/json"
 	"fmt"
 	"net/http"
 	"strings"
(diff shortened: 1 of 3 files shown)


api/integrations_test.go
```diff
@@ -297,6 +298,48 @@ func TestIntegrationsList(t *testing.T) {
...
+		// @afiune the only rist here is that if the caller tries to access the state
+		// directly before checking that there is actually an state, the go program
+		// will panic
+		assert.Nil(t, resData.State)
+
+		// verify the generate json blob from the response
+		jsonBlob, err := json.Marshal(resData)
+		if assert.Nil(t, err) {
+			assert.Equal(t,
+				"{\"INTG_GUID\":\"MY_MOCKED_ID\",\"NAME\":\"minimal_integration\",\"TYPE\":\"MINIMAL\",\"ENABLED\":0,\"DATA\":{}}",
```
afiunelw, line 336: Here you can see that there is no `STATE` field anymore. 💯

Issue #81, opened 2020-04-27
Add sub-cmds to `integrations` cmd to support CRUD operations


afiunelw:
Today, the `lacework integration` command has only one the `list` sub-command,
this card is about adding additional sub-commands to `create`, `delete`, and perhaps
`update` external integration.

**Jira: ALLY-91**

PR #88, opened 2020-04-30
Add commands to create and delete integrations

afiunelw:
We are adding the ability to create external integration via the CLI, as
of this commit, we can create Docker Hub and AWS Config integrations.
The code has comments of the rest of the integrations available in the
platform, such integrations will be added on further changes.

## Example: Create a Docker Hub integration
```
$ lacework integration create
? Choose an integration type to create:  Docker Hub
▸ Name:  techallylw docker hub from CLI
▸ Username:  techallylw
▸ Password:  ***********************
▸ Limit by Tag:  *
▸ Limit by Label:  *
▸ Limit by Repository:
▸ Limit Number of Images per Repo:  5
The integration was created.
```

This PR is also adding the ability to create container registry integrations
into the `api` Go package, for now, users can add only Docker Hub and
Docker V2 Registry integrations. Here is a basic example of the creation
of a Docker Hub integration:

```go
client, err := api.NewClient("account")
if err != nil {
  return err
}

docker := api.NewContainerRegIntegration("foo",
  api.ContainerRegData{
    Credentials: api.ContainerRegCreds {
      Username: "techally",
      Password: "secret",
    },
    RegistryType: api.DockerHubRegistry.String(),
    RegistryDomain: "index.docker.io",
    LimitByTag: "*",
    LimitByTag: "*",
    LimitByLabel: "*",
    LimitNumImg: "5",
  },
)

client.Integrations.CreateContainerRegistry(docker)
```

Finally, this change includes a new command to delete external integrations
by providing an integration GUID.
```
$ lacework integration delete TECHALLY_F3B88157AE9DEFBB9A33539A1D2ACB0884EEDEAC10124D4
```

To find the list of integration configured on your Lacework account, a user
can use the command:
```
$ lacework int list
```

tenor-51003174

## Unix Support (Linux/Darwin)
Screen Shot 2020-04-30 at 12 05 50 PM

## Windows Support

From Powershell:
Screen Shot 2020-04-30 at 12 42 40 PM

From CMD:
Screen Shot 2020-04-30 at 12 41 19 PM

Closes https://github.com/lacework/go-sdk/issues/81


cli/cmd/integration.go
```diff
@@ -93,11 +113,35 @@ var (
...
 	// integrationDeleteCmd represents the delete sub-command inside the integration command
 	integrationDeleteCmd = &cobra.Command{
-		Use:    "delete",
-		Hidden: true,
-		Short:  "delete an external integrations",
-		Args:   cobra.NoArgs,
-		RunE: func(_ *cobra.Command, _ []string) error {
+		Use:   "delete <int_guid>",
+		Short: "delete an external integrations",
+		Long: `Delete an external integration by providing its integration GUID, to find the
```
scottford-lw, line 118: This is a run on sentence. It should be something to effect of...

```
Delete an external integration by providing its integration GUID. Integration GUIDs can be found by using the 'lacework integration list' command
```

cli/cmd/integration.go
```diff
@@ -71,11 +72,30 @@ var (
...
 	// integrationCreateCmd represents the create sub-command inside the integration command
 	integrationCreateCmd = &cobra.Command{
-		Use:    "create",
-		Hidden: true,
-		Short:  "create an external integrations",
-		Args:   cobra.NoArgs,
+		Use:   "create",
+		Short: "create an external integrations",
+		Args:  cobra.NoArgs,
+		Long: `This command will prompt an interactive session that will help you create
```
scottford-lw, line 78: I think rather than "This command will..." we should just be explicit...

`Creates an external integration in your account through an interactive session. Use the '--noninteractive' flag to pass disable interactive mode`

@afiunelw I also see that `lacework int create --help` does not specify which parameters you need to pass if you use `--noninteractive`

cli/cmd/integration_aws.go
```diff
@@ -0,0 +1,125 @@
...
+	aws := api.NewAwsCloudTrailIntegration(answers.Name,
+		api.AwsIntegrationData{
+			QueueUrl: answers.QueueUrl,
+			Credentials: api.AwsIntegrationCreds{
+				RoleArn:    answers.RoleArn,
+				ExternalId: answers.ExternalID,
+			},
+		},
+	)
+
```
scottford-lw, line 120: Do we want to implement any validation that here as you do with the `lacework configure` command? Might be a good feature down the line if not. Right now I can pass any text to the form and it will try to create.

```
lacework int create                                                                                                                                                                                                                                                                       scottford@scott-fords-mbp
? Choose an integration type to create:  AWS CloudTrail
▸ Name: asdf
▸ Role ARN: sdf
▸ External ID: asdf
▸ SQS Queue URL: asdf
```

afiunelw: Yes, I would love to add validation for every field to avoid hitting the API with silly requests, I go and create a card to add this in further PRs.

cli/cmd/integration.go
afiunelw: I will change the message, it is trying to let the user know that, the command is disabled if the CLI is running in noninteractive mode.

PR #100, opened 2020-05-08
fix(api): use correct types on events response

afiunelw:
This PR is fixing the case where events with type `NewExternalServerDNSConn`
are sending the `bytes` field with a scientific notation format.

The error we are seeing is the following:
```
$ lacework event show 37
ERROR unable to get event details: json: cannot unmarshal number 6.873831E7 into Go struct field eventDnsNameEntity.data.entity_map.DnsName.total_out_bytes of type int32
```

Switching these fields to be `float32` seems to fix the parsing problem.

### DIFF
--- a/api/events.go
+++ b/api/events.go
@@ -150,20 +150,20 @@ type eventContainerEntity struct {
 type eventDnsNameEntity struct {
 	Hostname      string  `json:"hostname"`
 	PortList      []int32 `json:"port_list"`
-	TotalInBytes  int32   `json:"total_in_bytes"`
-	TotalOutBytes int32   `json:"total_out_bytes"`
+	TotalInBytes  float32 `json:"total_in_bytes"`
+	TotalOutBytes float32 `json:"total_out_bytes"`
 }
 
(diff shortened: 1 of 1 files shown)


scottford-lw:
@afiunelw I am testing this one out and still running into `unmarshal` errors. Check this one out in `customerdemo`...

```
» lacework events show 34133 -p customerdemo                                                                                                                                                                        scottford@scott-fords-mbp
Usage:
  lacework event show <event_id> [flags]

Flags:
  -h, --help   help for show

Global Flags:
      --noninteractive      disable interactive progress bars (i.e. 'spinners')

ERROR unable to get event details: json: cannot unmarshal string into Go struct field eventIpAddressEntity.data.entity_map.IpAddress.threat_tags of type []string
```

afiunelw:
@scottford-lw thank you for pointing that out, I have verified more event types and fixed a few other potential problems we could find. There is no other way to know if we will error with more parser errors but to use the tool and catch them as they show up, I would like to take this back to the platform team and make sure we provide consistent API responses.

PR #140, opened 2020-06-09
ci: upload artifacts to release

afiune:
tenor-141726988

### DIFF
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -358,7 +358,7 @@ create_release() {
   local _content_type
   local _artifact
   local _upload_url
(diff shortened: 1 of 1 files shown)


scripts/release.sh
```diff
@@ -358,7 +358,7 @@ create_release() {
   local _content_type
   local _artifact
   local _upload_url
-  _upload_url=$(jq .upload_url $_release)
+  _upload_url=$(jq .upload_url $_release | sed 's/"//g' | cut -d{ -f1)
```
afiune, line 361: This sed and cut is removing the `{?name,label}` at the end of the upload url

Issue #146, opened 2020-06-10
Add User-Agent headers to projects


afiune:
As a Lacework Engineer,
I would like to track the usage of the [Go API abstraction](https://github.com/lacework/go-sdk/tree/master/api)
As well as the usage coming from the [Lacework CLI](https://github.com/lacework/go-sdk/tree/master/cli),
So that I can understand how users interact with our tool via the Lacework API.

## Proposal Solution

We need to add a default `User-Agent` header to the Go-API and make it configurable
so that other components from this repository, as well as external components, can
customize this field (e.i. our terraform provider). This will allow us to track the usage
of different components and make sense of how users interact with our tools and APIs.

**JIRA: ALLY-135**

PR #152, opened 2020-06-17
feat: set User-Agent header (backend metrics)

afiune:
This change also includes the ability to override the User-Agent header
as well as adding any other global header to the Go api client, to
create a new client with a global header use the Option `WithHeader()`
like the following example:
```go
client, err := api.NewClient("test",
	api.WithHeader("User-Agent", "test-agent"),
	api.WithHeader("Custom-ABC", "value-abc"),
	api.WithHeader("Custom-XYZ", "value-xyz"),
)
```

By default, the Go API Client will set the `User-Agent` header to `Go Client/{VERSION}`
while other components will override this value, a list of components with their values
are:
- **Default:** Go API Client (`User-Agent: Go Client/{VERSION}`)
- Terraform Provider (`User-Agent: Terraform/{VERSION}`)
- Lacework CLI (`User-Agent: Command-Line/{VERSION}`)

Resolves: https://github.com/lacework/go-sdk/issues/146


scottford-lw:
@aaronlamlw I have reviewed but your review is critical as the owner of the api

Issue #112, opened 2020-05-25
Configure the Lacework CLI in non-interactive mode


ghost:
As a Lacework CLI user that uses the tool in CI/CD Pipelines,
I want to have a non-interactive way to configure the Lacework CLI,
So I don't have to manually create the config file `~/.lacework.toml` inside my pipeline. 

## Proposal Solution
When a user runs:
```
$ lacework configure -a foo -k bar -s bubu --noninteractive
```
The result should be the creation of the `~/.lacework.toml` without asking the user
to validate the inputs. (non-interactive mode) 

The command should load these settings from all sources, as usual, environment
variables, parameters (including `--json_file`), and the actual config file.

PR #158, opened 2020-07-06
feat(cli): configure in non-interactive mode

afiune:
>As a Lacework CLI user that uses the tool in CI/CD Pipelines,
>I want to have a non-interactive way to configure the Lacework CLI,
>So I don't have to manually create the config file `~/.lacework.toml` inside my pipeline.

When a user runs:
```
$ lacework configure -a foo -k bar -s bubu --noninteractive
```
The result is the creation of the `~/.lacework.toml` without asking the user
to validate the inputs. (non-interactive mode)

The command loads these settings from all sources, as usual, environment
variables, parameters (including `--json_file`), and the actual config file.

Closes https://github.com/lacework/go-sdk/issues/112

### DIFF
--- a/cli/cmd/configure.go
+++ b/cli/cmd/configure.go
@@ -54,6 +54,19 @@ type credsDetails struct {
 	ApiSecret string `toml:"api_secret" json:"api_secret" survey:"api_secret"`
 }
 
+func (c *credsDetails) Verify() error {
+	if c.Account == "" {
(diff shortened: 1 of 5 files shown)


afiune:
You tested that single case where it does NOT work! Amazing 🎉 thank you, fixing it now. 

tenor-153515441

afiune:
tenor-79052486

PR #177, opened 2020-07-28
feat(cli): --subaccount flag to manage sub-accounts

afiune:
This change will allow users to manage sub-accounts with a new flag named `--subaccount`.

To list integrations from the primary account
```
$ lacework integrations list
```

To list integrations from a sub-account
```
$ lacework integrations list --subaccount sub-account-name
```

Additionally, it is possible to configure a persistent sub-account with the command:

    $ lacework configure --subaccount sub-account-name

To remove the configured sub-account use the `--remove_subaccount` flag:

    $ lacework configure --remove_subaccount

Finally, this new flag comes with its environment variable `LW_SUBACCOUNT` that can be
used to override settings coming from the Lacework configuration file.


afiune:
We need to change this PR to match the terminology from our documentation,
instead of `--tenant` this argument needs to be called `--subaccount`.

afiune:
Superseded by https://github.com/lacework/go-sdk/pull/431

Issue #86, opened 2020-04-29
Configure account should validate customer does not include lacework.net


scottford-lw:
We should ensure that the `lacework configure` subcommand validates the account does not include `.lacework.net` or handles it correctly if that is passed.

PR #187, opened 2020-08-06
feat(cli): better ux in account validation

afiune:
This change will add a better user-experience to the `lacework configure`
command so that if the user provides the complete URL as the account
name we will subtract the account and notify the end-user:

```
$ lacework configure
Passing '.lacework.net' domain not required. Using 'tech-ally'
▸ Account: tech-ally
▸ Access Key ID: TECHALLY_1234567890
▸ Secret Access Key: (*****************************91f3)

You are all set!
```

Closes https://github.com/lacework/go-sdk/issues/86

### DIFF
--- a/cli/cmd/configure.go
+++ b/cli/cmd/configure.go
@@ -24,6 +24,8 @@ import (
 	"fmt"
 	"io/ioutil"
 	"path"
+	"regexp"
+	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/BurntSushi/toml"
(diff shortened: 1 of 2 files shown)


scottford-lw:
@afiune The update is working for me, but I am still not sure about the user experience...

This line...

`Account subtracted from URL 'tech-ally.lacework.net' (should only be the subdomain of URL)` 

...is not totally clear to me. I think the goal is to let the user know they don't need to pass `.lacework.net` so they don't do that again with other accounts they configure. 

What about this?

`Passing '.lacework.net' domain is not required. Using <account> `

I think it is the language `Account subtracted from URL` that is not totally clear.

afiune:
Sounds good. I'll update the message as suggested. 💯

afiune:
@scottford-lw updated!

tenor-137781822

PR #213, opened 2020-09-17
ci(slack): notify pipeline failures

afiune:
We want to have notifications when our nightly builds fail!

tenor-443900


scottford-lw (approved):
👍

PR #224, opened 2020-10-06
docs: automatically generate cli docs

afiune:
We are adding automatic documentation for the Lacework CLI. Now, on every release
we will have the folder `cli/docs/` deployed with detailed information about every command.


scottford-lw (approved):
👍 ⭐

Issue #209, opened 2020-09-14
Programatic access to configured profile data


alanwthatcher-lw:
It would be great to have programatic access to the configuration settings for profiles, similar to what the AWS cli provides. This would make it easy to export the necessary environment variables for workflows that prefer it. Thinking something like:

`lacework config get api_key --profile foo`

Which would allow you to do something like this in a script or shell function to export them:

`export LW_API_KEY="$(lacework configure get api_key --profile foo)"`

Or maybe perhaps a more direct way to export the values to env variables:

`lacework config env-export --profile foo`

PR #225, opened 2020-10-07
feat(cli): programatic access to profile data

afiune:
We are adding two new subcommands to the `lacework configure` command.

The `lacework configure list` command will list all profiles configured
into the configuration file `~/.lacework.toml`

The `lacework configure show <config_key>` command prints the current
computed configuration data from the specified configuration key. The
order of precedence to compute the configuration is flags, environment
variables, and the configuration file ~/.lacework.toml.

The available configuration keys are:
* `profile`
* `account`
* `api_secret`
* `api_key`

To show the configuration from a different profile, use the flag `--profile`
```
$ lacework configure show account --profile my-profile
```

Closes https://github.com/lacework/go-sdk/issues/209

### DIFF
--- a/cli/cmd/cli_state.go
+++ b/cli/cmd/cli_state.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"time"
 
+	"github.com/BurntSushi/toml"
 	"github.com/briandowns/spinner"
 	"github.com/fatih/color"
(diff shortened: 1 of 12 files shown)


scottford-lw (approved):
👍  Nice work!!!!

PR #229, opened 2020-10-12
docs(cli): disable timestamp for automatic docs

afiune:
On every release we are updating every single documentation, look at the
latest release https://github.com/lacework/go-sdk/pull/228, this causes noise to the reviewer.

This PR is disabling the timestamp for all documents living in `cli/docs/`.


scottford-lw (approved):
Glad you removed that!  👍

Issue #203, opened 2020-09-02
Upgrade to Golang 1.15


afiune:

afiune:
Closed by https://github.com/lacework/go-sdk/pull/265

PR #232, opened 2020-10-16
Upgrade to Golang 1.15 

dev117uday:
From issue #203 

Upgrading Go version to latest versions

If maintainers find this PR invalid/spam, Or i made a mistake, please take some take a few seconds out of precious time and correct me. I am a newbee and your remark be will helpful for me in learning.


afiune:
@dev117uday Thank you for your contribution! You did a good job, though, when upgrading the Golang version you don't have to upgrade the files inside the `vendor/` directory, these are Go modules that we vendor with the command `go mod vendor` and they shouldn't be modified manually. Please, revert the changes made inside the `vendor/` directory and I will proceed to test this change locally.

Thanks again! 🙌🏽💯🎉

dev117uday:
Oops !! I made a mistake. i will open a new PR with correct changes

PR #233, opened 2020-10-26
fix(ci) Update CI test node

scottford-lw:
Updating persistent Circle CI test node for host vulnerability testing. 

Signed-off-by: Scott Ford <scott.ford@lacework.net>

### DIFF
--- a/integration/host_vulnerability_test.go
+++ b/integration/host_vulnerability_test.go
@@ -29,10 +29,12 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-// persistent host "national-parks-mongodb"
+// persistent host "ip-10-0-1-170.us-west-2.compute.internal"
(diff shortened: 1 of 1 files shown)


scottford-lw:
@afiune the integration test is failing because the hostname is not matching the test. On the node the `hostname` and `hostname -f` both return `circle-ci-test-node`. In the UI the node shows up at `circle-ci-test-node`, but `lacework vul host show-assessment 30` returns a hostname of `ip-10-0-1-170.us-west-2.compute.internal`. 

Obviously we can update the test to match the response, but I am not clear what is causing the mis-match. Let me know if you have thoughts.

afiune:
@scottford-lw great catch! This is definitely a bug. I'll go check on the backend and in the meantime, I will open an issue.

afiune:
Bug ticket created: **RAIN-13990**

Let's omit to check for the hostname for now.

afiune (approved):
tenor-262859852

PR #246, opened 2020-11-16
ci: remove slack alert for win systems

afiune:
Since it is not supported by CircleCI :facepalm:


afiune:
tenor-112434238

PR #253, opened 2020-11-30
feat(compliance): new aws list-accounts command

afiune:
This new command will list all AWS accounts configured in the user's
Lacework account.

If there are no AWS accounts configured, the following message will be
displayed:
```
$ lacework compliance aws list
There are no AWS accounts configured in your account.

Get started by integrating your AWS accounts to analyze configuration compliance using the command:

    $ lacework integration create

Or, if you prefer to do it via the WebUI, log in to your account at:

    https://my-account.lacework.net

Then navigate to Settings > Integrations > Cloud Accounts.
```


smford22:
@afiune The concept of being able to list all of the accounts from the CLI/API is a great as you have explained. The concern here for me is that this feature doesn't seem to be specific to compliance. It almost seems to be more related to the `integration` command, rather than `compliance`. 

When the new API comes out, will this still be the best approach, or will you need to refactor again, or possibly redesign from scratch. 

I don't question at all the feature, I am just concerned that we may be rushing to get this out with what we have today over what we can design when the new API comes out.

afiune:
@scottford-io @scottford-lw @smford22 Thank you for your feedback but unfortunately, I don't understand your concern. This feature is 100% specific to compliance, the command shows the list of AWS accounts that a user has integrated into their Lacework account.

The reason why this is feasible is that to do AWS Compliance, we require to have one AWS Config (`AWS_CFG`) integration per AWS account. So this integration is 1-to-1. As many `AWS_CFG` integrations a user has, as many AWS accounts are being monitored.

I think the confusion here is the fact that for other cloud providers this correlation is different, for example, to do Azure Compliance we can have a single Azure Config (`AZURE_CFG`) integration that is monitoring hundreds of Subscriptions. So this integration is 1-to-many. 

Even for AWS CloudTrail monitoring (`AWS_CT_SQS`), a user can have one integration that might have hundreds of AWS accounts being monitored. This is also an integration that is 1-to-many, but for this one we can't know how many accounts are being monitored, mainly because there are no APIs to look at CloudTrail data. (Nor Acticity Log and Audit Log)

TL; DR; This is the right thing to do and will continue to be so after the new APIs get released.

scottford-lw:
@afiune thanks for walking me through this. I understand the use case better and agree that it moves the need and should be implemented. 🙌

scottford-lw (approved):
👍  🙌

Issue #249, opened 2020-11-19
ALLY-250: Fix install.sh script for systems without 'info' command


afiune:
Feedback from a new customer:
>The lacework (CLI)  installer uses the 'info' command which may not be available in all flavors of Linux.
>It should revert to using echo / or something like that if that doesn't exist.

PR #254, opened 2020-12-07
fix(install.sh): avoid logging with 'info' cmd

afiune:
Closes https://github.com/lacework/go-sdk/issues/249
JIRA: https://lacework.atlassian.net/browse/ALLY-250


scottford-lw (approved):
👍

PR #259, opened 2020-12-08
fix(api): avoid updating AgentTokenResponse.Props

afiune:
This change is making the `api.AgentTokenResponse` to have the field
`Props` to be a pointer of `AgentTokenProps`, this way we are not
sending an empty description and updating it.


scottford-lw (approved):
👍 Looks good @afiune

PR #274, opened 2020-12-22
feat(cli): enable agent install command

afiune:
We are ready to enable this new command to help onboarding new Lacework agents!

Users can now install agents with:

```
$ lacework agent install
```

When this command is executed without any additional flag, an interactive prompt will be
launched to help gather the necessary authentication information to access the remote host.

To authenticate to the remote host with a username and password.
```
$ lacework agent install <host> --ssh_username <your-user> --ssh_password <secret>
```

To authenticate to the remote host with an identity file instead.

```
$ lacework agent install <user@host> -i /path/to/your/key
```

To provide an agent access token of your choice, use the command 'lacework agent token list',
select a token and pass it to the '--token' flag.

```
$ lacework agent install <user@host> -i /path/to/your/key --token <token>
```

### Agent Install Example

lacework-cli-agent-install


lwrunner/runner.go
```diff
@@ -24,27 +24,85 @@ import (
...
+	run.Auth = []ssh.AuthMethod{ssh.Password(secret)}
+}
+
+func (run *Runner) Address() string {
+	return fmt.Sprintf("%s:%d", run.Hostname, run.Port)
+}
+
+// Exec executes a command on the configured remote host
+func (run *Runner) Exec(cmd string) (stdout bytes.Buffer, stderr bytes.Buffer, err error) {
+	// @afiune make the port configurable
```
afiune, line 87: delete

lwrunner/runner_test.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/lacework/go-sdk/lwrunner"
+
+	"github.com/stretchr/testify/assert"
+)
+
+func TestLwRunnerNew(t *testing.T) {
+	subject := lwrunner.New("ubuntu", "my-test-host")
+	assert.Equal(t, subject.Port, 22)
+	assert.Equal(t, subject.User, "ubuntu")
+}
```
afiune, line 33: add a few more tests

lwrunner/runner.go
```diff
@@ -24,27 +24,85 @@ import (
...
+	hostKeyCallback := ssh.InsecureIgnoreHostKey()
+
+	// try to use the known_hosts file as a host key callback
+	home, err := homedir.Dir()
+	if err == nil {
+		hostKeyFromKnownHosts, err := knownhosts.New(path.Join(home, ".ssh", "known_hosts"))
+		if err == nil {
+			hostKeyCallback = hostKeyFromKnownHosts
+		}
+	}
```
dmurray-lacework, line 51: Would this work in here instead?

```suggestion
	if home, err := homedir.Dir(); err == nil {
		hostKeyCallback, err = knownhosts.New(path.Join(home, ".ssh", "known_hosts"))
	}
```

dmurray-lacework (approved):
Looks Good Salim!

scottford-lw:
@afiune @dmurray-lacework Given that we provide an interactive menu, is there a reason why we require a host as well? Should we consider `lacework agent install` and the first interactive prompt is to ask for a host?

Right now the behavior if you run `lacework agent install` the command fails requiring a host.

```
lacework agent install                                                                                                                                                        scottford@scott-fords-mbp
Usage:
  lacework agent install <host> [flags]

Flags:
  -h, --help                   help for install
  -i, --identity_file string   identity (private key) for public key authentication (default "~/.ssh/id_rsa")
      --ssh_password string    password for authentication
      --ssh_username string    user to log in as on the remote host
      --token string           agent access token

Global Flags:

ERROR accepts 1 arg(s), received 0
```

scottford-lw:
@afiune When in interactive mode, unnamed tokens produce strange UX. 

The first prompt in interactive mode is `Choose an agent access token:` which is followed by a newline and a `>`  In `customerdemo` the list of tokens has a large number of tokens that are unnamed which requires the user to keep pressing arrows until a token is listed with a name. 

cli-interactive

my suggestion is that we either list unnamed tokens as `unnamed` and/or we list the token as well

```
lacework agent install ec2-user@54.144.199.123                                                                                                                                scottford@scott-fords-mbp
? Choose an agent access token:   [Use arrows to move, type to filter]

> (unnamed)    e2da76691058e9ab953e3ecfe74ee39fc8e63af4bizzzz                        Enabled
  (unnamed)    ea4aae15159a49efc12313312ff858c9ea26be5832barrr                        Enabled
  (unnamed)    ef778ff89b441288b44d2b8b2cf3f10e48301f8764fooo                        Enabled
```

lwrunner/runner.go
afiune: As talked over Slack, this misses the second error check. I will separate the logic into its own function to avoid the nested `if` statements.

afiune:
@scottford-lw this is absolutely great feedback! Thank you so much for testing this feature. I will
update the problem with the empty tokens. 💯

scottford-lw:
@afiune have you considered the following scenarios...

what should the behavior be when a node already has an agent installed? 

Is there a way to force install a newer agent, or a new token? 

Right now we just pass the output directly from the node. Here is an example of a attempting to install the agent on a node that already has an agent installed...

```
~/src/go-sdk(afiune/agent-install*) » lacework agent install ec2-user@54.144.199.123 -p tech-ally                                                                                                                   scottford@scott-fords-mbp
? Choose an agent access token:  lacework-cli-3
+ curl -sSL https://s3-us-west-2.amazonaws.com/www.lacework.net/download/3.3.5_2020-11-16_master_ac0e65055f11f4f59bab6ea4dfa61dcafaa9a3f1/lacework-3.3.5-1.x86_64.rpm
+ sh -c 'sleep 3; yum  -y install /tmp/2RbEkH.rpm'
Error: Nothing to do
Usage:
  lacework agent install <host> [flags]

Flags:
  -h, --help                   help for install
  -i, --identity_file string   identity (private key) for public key authentication (default "~/.ssh/id_rsa")
      --ssh_password string    password for authentication
      --ssh_username string    user to log in as on the remote host
      --token string           agent access token

Global Flags:

ERROR unable to install agent: Process exited with status 1
```

afiune:
@scottford-lw about the "why we have the host as an argument?", I think that we want to copy/mimic
the way the `ssh` command works as much as possible so that users can understand and feel naturally
using this command.

afiune:
@scottford-lw I think that we should try to add the functionality to upgrade agents later on but for now, we definitely need to check if the agent is already installed and if so, inform the user instead of trying to install and fail. Even though there are a few operating systems that we support reinstalling by only running again the install script, but since this is not a consistent behavior, we will avoid confusion by disallowing it.

A simple check like this would be enough:
```
  $ ps -fea | grep 'lacework/datacollector'
root     16606     1  0 15:02 ?        00:00:00 /var/lib/lacework/datacollector
root     16647 16606  0 15:02 ?        00:00:00 /var/lib/lacework/datacollector -r=collector
```

scottford-lw:
@afiune With the latest changes, when a user does not supply a token, the interactive prompt opens with a list of tokens to choose from, but seem to only list named tokens and excludes unnamed tokens. Is this the expected behavior? Can we consider listing unnamed tokens as '(unnamed)'? 

https://github.com/lacework/go-sdk/pull/274#issuecomment-749665547

afiune:
@scottford-lw Let us add that functionality later on, if the user wants to use an "unnamed" token they can pass it via the command line arguments, the command `lacework agent token list` should work as expected. 

Having said that, unnamed tokens should be disallowed, at least the CLI and API don't allow tokens to be created without a name (token alias).

scottford-lw (approved):
👍

PR #283, opened 2021-01-06
feat(cli): gen-pkg-manifest detect running kernel

afiune:
The default behavior of most Linux distros is to keep the last `N` kernel
packages installed for users that need to fall back in case the new
kernel does not boot.  However, the presence of the package does not mean
that kernel is active.  We must continue to allow the standard kernel
package preservation behavior without providing false-positives of
vulnerabilities that are not active.

This feature detects the active kernel and removes any other
installed-and-inactive kernel from the generated package manifest.

Additionally, we are adding instrumentation to trace the CLI feature
`gen_pkg_manifest` and understand how often is it used and find out
if there are problems with it.

### gen_pkg_manifest feature event

Screen Shot 2021-01-06 at 1 00 29 PM

With feature data:
```json
{
  "active_kernel": "4.14.209-160.339.amzn2.x86_64",
  "kernel_suppressed_158": "kernel-0:4.14.203-156.332.amzn2",
  "os": "amzn",
  "os_ver": "2",
  "pkg_manager": "rpm",
  "total_manifest_pkgs": 441
}
```

**JIRA: ALLY-264**

### DIFF
--- a/cli/cmd/package_manifest.go
+++ b/cli/cmd/package_manifest.go
@@ -33,7 +33,7 @@ import (
 	"github.com/lacework/go-sdk/api"
 )
 
-var SupportedPackageManagers = []string{"dpkg-query", "rpm"} // @afiune can we support ym and apk?
+var SupportedPackageManagers = []string{"dpkg-query", "rpm"} // @afiune can we support yum and apk?
 
 type OS struct {
 	Name    string
@@ -105,7 +105,7 @@ func (c *cliState) GeneratePackageManifest() (*api.PackageManifest, error) {
 		managerQuery = []byte(strings.Join(mq, "\n"))
 	default:
 		return manifest, errors.New(
-			"this is most likely a mistake on us, please report it to support.lacework.com.",
+			"this is most likely a mistake on us, please report it to support@lacework.com.",
 		)
 	}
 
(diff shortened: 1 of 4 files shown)


cli/cmd/package_manifest.go
```diff
@@ -236,3 +334,16 @@ func (c *cliState) checkPackageManagerWithNativeCommand(manager string) bool {
...
+		return pkgVer
+	}
+
+	pkgVerSplit := strings.Split(pkgVer, ":")
+	if len(pkgVerSplit) == 2 {
+		return pkgVerSplit[1]
+	}
+
+	return pkgVer
+}
```
dmurray-lacework, line 349: Can reduce this a little?
```suggestion

func removeEpochFromPkgVersion(pkgVer string) string {
	if strings.Contains(pkgVer, ":") {
	  pkgVerSplit := strings.Split(pkgVer, ":")
	  if len(pkgVerSplit) == 2 {
		  return pkgVerSplit[1]
	  }
	}

	return pkgVer
}
```

cli/cmd/package_manifest_test.go
```diff
@@ -0,0 +1,93 @@
...
+				Pkg: "kernel", PkgVer: "0:4.14.209-160.339.amzn2", // with EPOCH
+			},
+			api.OsPkgInfo{
+				Os: "amzn", OsVer: "2",
+				Pkg: "kernel", PkgVer: "4.14.203-156.332.amzn2", // without EPOCH
+			},
+		},
+	}
+	subject := cli.removeInactivePackagesFromManifest(manifest, "rpm")
+	assert.Empty(t, subject)
```
dmurray-lacework, line 49: After calling removeInactivePackagesFromManifest should api.PackageManifest contain only the active kernel? If so, can this be checked in a unit test.

dmurray-lacework (approved):
Hey, looks good. Have a couple of comments.

cli/cmd/package_manifest_test.go
afiune: Yes, and It is being tested already.

cli/cmd/package_manifest.go
afiune: Sounds good! I'll change it 💯

cli/cmd/package_manifest_test.go
dmurray-lacework: Awesome!

afiune:
@dmurray-lacework wait error from your suggestion, I will have to rebase:
```
#!/bin/bash -eo pipefail
make fmt-check
/bin/sh: 1: test: /go/src/github.com/lacework/go-sdk/cli/cmd/package_manifest.go: unexpected operator
make: *** [Makefile:52: fmt-check] Error 2

Exited with code exit status 2
CircleCI received exit code 2
```

Issue #237, opened 2020-10-30
support package-manifest with over 1k packages


scottford-lw:
The `api/v1/external/vulnerabilities/scan` api has a limit of 1k per package manifest. It would be great if `generate-pkg-manifest` command could automatically break up the manifest into multiple files.

```
NOTE: Calls to this operation are rate limited to 10 calls per hour, per access key. If this rate limit is exceeded, an exception is thrown.
Also, note that this operation is limited to 1k of packages per payload. If you require a payload larger than 1k, you must make multiple requests.
For more information about creating an API access key and token to run this operation and using this operation with organization resources,
```

afiune:
I think that it would be better if we make the `scan-pkg-manifest` command to support a single manifest file with over 1k packages. I feel that making the `generate-pkg-manifest` create multiple files will cause a bit more difficulties mainly because then the user would have to parse every file and submit multiple scans. Thoughts @scottford-lw ?

scottford-lw:
@afiune I agree that would be a better user experience. It would be great to get some feedback from a customer on this as well. Can you ask in customer slack with the customer that inspired this one?

PR #285, opened 2021-01-07
feat(cli): support manifest bigger than 1k packages

afiune:
## User Story

As a user of the Lacework CLI,
I would like to be able to submit scans of package manifests bigger than 1,000 packages,
So I don't have to implement a splitting mechanism and run multiple CLI commands for a single manifest.

## Implementation Details

The CLI will now check if the package manifest has more than the maximum
number of packages, if so, it will split the package manifest into
multiple chunks and trigger multiple API requests.

**NOTE:** We disallow more than 10 parallel requests (workers), which
are more than 10,000 packages on a single manifest/system.

Closes https://github.com/lacework/go-sdk/issues/237

**JIRA: ALLY-275**

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -147,6 +147,11 @@ func (c *Client) URL() string {
 	return c.baseURL.String()
 }
 
+// ValidAuth verifies that the client has valid authentication
+func (c *Client) ValidAuth() bool {
+	return c.auth.token != ""
+}
+
 // newID generates a new client id, this id is useful for logging purposes
 // when there are more than one client running on the same machine
 func newID() string {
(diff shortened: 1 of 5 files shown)


dmurray-lacework (approved):

afiune:
### Added new feature `split_pkg_manifest`

Screen Shot 2021-01-07 at 12 24 17 PM


This new feature event will show the following `feature.data`:
```
{
  "total_packages": 1159,
  "worker0_total_vulns": 14,
  "worker1_total_vulns": 3,
  "workers": 2
}
```
Which will match with what the user will see on the output of the command:

Screen Shot 2021-01-07 at 12 22 24 PM

scottford-lw (approved):
👍  Great work on this @afiune 🙌

PR #297, opened 2021-01-21
feat: Add telemetry to detect Homebrew installations

dmurray-lacework:
Add telemetry to detect when customers are running the Lacework CLI that has been installed via Homebrew.

### DIFF
--- a/cli/cmd/honeyvent.go
+++ b/cli/cmd/honeyvent.go
@@ -45,10 +45,6 @@ const (
 	// disable telemetry sent to Honeycomb
 	DisableTelemetry = "LW_TELEMETRY_DISABLE"
 
-	// HomebrewInstall is an environment variable that denotes the
-	// install method was via homebrew package manager
-	HomebrewInstall = "LW_HOMEBREW_INSTALL"
-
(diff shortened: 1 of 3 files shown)


cli/cmd/honeyvent.go
```diff
@@ -45,6 +45,10 @@ const (
 	// disable telemetry sent to Honeycomb
 	DisableTelemetry = "LW_TELEMETRY_DISABLE"
 
+	// HomebrewInstall is an environment variable that denotes the
+	// install method was via homebrew package manager
+	HomebrewInstall = "LW_HOMEBREW_INSTALL"
```
afiune, line 50: This is great!! Though, two things:
* Could we move this environment variable out of the `honeyvent.go` file and instead put it inside `cli_unix.go`
* After moving this variable, could we use it here https://github.com/lacework/go-sdk/blob/master/cli/cmd/cli_unix.go#L39

afiune (changes requested):
Really great work! Just one modification

afiune (approved):
tenor-261285180

PR #299, opened 2021-01-22
ci: trigger homebrew update script

dmurray-lacework:
Update the release workflow to trigger homebrew formula job

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -138,6 +138,10 @@ jobs:
           fail_only: true
           mentions: << pipeline.parameters.slack-mentions >>
           only_for_branches: <<pipeline.parameters.only_for_branches>>
+  trigger-homebrew-update:
+    executor: alpine
+    steps:
+      - run: curl -u "$CIRCLE_TOKEN" https://circleci.com/api/v1.1/project/gh/lacework/homebrew-tap/tree/main -d "build_parameters[CIRCLE_JOB]=update"
 
 workflows:
   version: 2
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -138,6 +138,10 @@ jobs:
           fail_only: true
           mentions: << pipeline.parameters.slack-mentions >>
           only_for_branches: <<pipeline.parameters.only_for_branches>>
+  trigger-homebrew-update:
+    executor: alpine
+    steps:
+      - run: curl -u "$CIRCLE_TOKEN" https://circleci.com/api/v1.1/project/gh/lacework/homebrew-tap/tree/main -d "build_parameters[CIRCLE_JOB]=update"
```
afiune, line 144: Really quick, why we don't use API v2? 🤔 https://circleci.com/docs/api/v2/#operation/triggerPipeline

dmurray-lacework: I don't see the option to trigger an individual job on the v2 api. As far as I know only the 1.1 api allows specifying a job. In this case we only want to run update.

afiune: Understood. I wonder if we should implement this trigger using workflows and parameters instead 🤔 

- We can revisit this later, let's move forward with this implementation.

afiune (approved):
tenor-130489947

.circleci/config.yml
dmurray-lacework: Sounds Good!

PR #303, opened 2021-01-26
feat(cli): New Service Now alert channel

dmurray-lacework:
Service Now alert channel

### DIFF
--- a/api/integration_alert_channels_service_now.go
+++ b/api/integration_alert_channels_service_now.go
@@ -1,0 +1,102 @@
+//
+// Author:: Darren Murray(<darren.murray@lacework.net>)
+// Copyright:: Copyright 2020, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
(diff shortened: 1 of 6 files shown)


api/integration_alert_channels_service_now.go
```diff
@@ -0,0 +1,102 @@
...
+	Message string                   `json:"message"`
+}
+
+type ServiceNowAlertChannel struct {
+	commonIntegrationData
+	Data ServiceNowChannelData `json:"DATA"`
+}
+
+type ServiceNowChannelData struct {
+	InstanceUrl   string `json:"INSTANCE_URL" mapstructure:"INSTANCE_URL"`
```
afiune, line 98: style: could we set URL capitalized? Just like we do with ID. 
```suggestion
	InstanceURL   string `json:"INSTANCE_URL" mapstructure:"INSTANCE_URL"`
```

afiune (approved):
One style change.

tenor-81739428

api/integration_alert_channels_service_now.go
dmurray-lacework: Will do!

PR #307, opened 2021-01-28
fix: Json mapping for Snow Username

dmurray-lacework:


scottford-lw (approved):

afiune (commented):
@scottford-lw @dmurray-lacework We need to find a way to avoid missing things like these. Maybe we should start building the CLI and create an integration for real before approval. Thoughts?

PR #321, opened 2021-02-16
fix(cli): match API client timeout with NGINX

afiune:
Currently, the Lacework CLI has a timeout of 60 seconds and users
are experiencing the following errors:
```
context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

We are increasing the timeout to match what we have on NGINX which
is 120 seconds, we add 5 more seconds to let NGINX close the connection
instead of the Lacework CLI.

Closes RAIN-15291

### DIFF
--- a/cli/cmd/cli_state.go
+++ b/cli/cmd/cli_state.go
@@ -183,6 +183,7 @@ func (c *cliState) NewClient() error {
 	client, err := api.NewClient(c.Account,
 		api.WithLogLevel(c.LogLevel),
 		api.WithApiKeys(c.KeyID, c.Secret),
+		api.WithTimeout(time.Second*125),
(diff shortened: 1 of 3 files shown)


cli/cmd/cli_state.go
```diff
@@ -183,6 +183,7 @@ func (c *cliState) NewClient() error {
 	client, err := api.NewClient(c.Account,
 		api.WithLogLevel(c.LogLevel),
 		api.WithApiKeys(c.KeyID, c.Secret),
+		api.WithTimeout(time.Second*125),
```
afiune, line 186: We purposely set the timeout to be 5 seconds more than the one configured in NGINX so that we let it close the connection instead of the CLI. That will make our monitoring system display failures/error-codes.

dmurray-lacework (approved):

PR #332, opened 2021-03-07
fix: Generate a new token upon a request with an expired token

dmurray-lacework:
When the Go client tries to submit an API request and the token stored in memory is expired, attempt to generate a new token.

https://lacework.atlassian.net/browse/ALLY-363

### DIFF
--- a/api/auth.go
+++ b/api/auth.go
@@ -20,6 +20,7 @@ package api
 
 import (
 	"fmt"
+	"time"
 
 	"go.uber.org/zap"
 )
@@ -33,6 +34,7 @@ type authConfig struct {
 	secret     string
 	token      string
 	expiration int
+	expiresAt  time.Time
 }
 
 // WithApiKeys sets the key_id and secret used to generate API access tokens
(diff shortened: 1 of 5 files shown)


api/auth.go
```diff
@@ -70,6 +72,7 @@ func WithToken(token string) Option {
 	return clientFunc(func(c *Client) error {
 		c.log.Debug("setting up auth", zap.String("token", token))
 		c.auth.token = token
+		c.auth.expiresAt = time.Now().Add(defaultTimeout)
```
afiune, line 75: Im not sure this is the right default token expiry time.... there should be a different variable...

afiune: Found it! We should use `DefaultTokenExpiryTime ` instead.

afiune (approved):
tenor-102072303

api/auth.go
```diff
@@ -70,6 +72,7 @@ func WithToken(token string) Option {
 	return clientFunc(func(c *Client) error {
 		c.log.Debug("setting up auth", zap.String("token", token))
 		c.auth.token = token
+		c.auth.expiresAt = time.Now().Add(DefaultTokenExpiryTime * time.Second)
```
afiune, line 75: Now that I think about this, there is no way we can know when will the token expire. 🤔

PR #334, opened 2021-03-09
ci: delete 'master' branch

afiune:
It is about time! 

tenor-41359850


afiune:
tenor-154518367

PR #343, opened 2021-03-18
test: Change target for container vuln scanning tests

dmurray-lacework:
Replace scan target container in vuln scanning integration tests
Add job to build and push test resources

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -162,6 +162,10 @@ jobs:
     executor: alpine
     steps:
       - run: curl -u "$CIRCLE_TOKEN" https://circleci.com/api/v1.1/project/gh/lacework/homebrew-tap/tree/main -d "build_parameters[CIRCLE_JOB]=update"
+prepare-test-resources:
+    executor: alpine
+    steps:
+      - run: make test-resources
 
 workflows:
   version: 2
@@ -224,6 +228,7 @@ workflows:
             branches:
               only: main
     jobs:
+      - prepare-test-resources
       - unit-test
       - build-cli
       - integration-test-linux:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -29,6 +29,10 @@ executors:
     resource_class: arm.medium
 
 jobs:
+  prepare-test-resources:
+    executor: go-executor
+    steps:
+      - run: make test-resources
   unit-test:
     executor: go-executor
     steps:
@@ -162,10 +166,6 @@ jobs:
     executor: alpine
     steps:
       - run: curl -u "$CIRCLE_TOKEN" https://circleci.com/api/v1.1/project/gh/lacework/homebrew-tap/tree/main -d "build_parameters[CIRCLE_JOB]=update"
-prepare-test-resources:
-    executor: alpine
-    steps:
-      - run: make test-resources
 
 workflows:
   version: 2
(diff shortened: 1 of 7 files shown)


integration/test_resources/vuln_scan/dirty.Dockerfile
```diff
@@ -0,0 +1 @@
+FROM node:latest
```
dmurray-lacework, line 1: Pin this version instead of using latest

scripts/prepare_test_resources.sh
```diff
@@ -0,0 +1,42 @@
...
+
+build_clean() {
+  echo "building clean container"
+  docker build --no-cache -f "integration/test_resources/clean.Dockerfile" -t techallylw/test-cli-clean .
+  docker push techallylw/test-cli-clean
+}
+
+build_dirty() {
+  echo "building dirty container"
+  docker build --no-cache -f "integration/test_resources/vuln_scan/dirty.Dockerfile" -t techallylw/test-cli-dirty .
```
dmurray-lacework, line 38: No need for --no-cache on this one

.circleci/config.yml
```diff
@@ -224,6 +228,7 @@ workflows:
             branches:
               only: main
     jobs:
+      - prepare-test-resources
```
dmurray-lacework, line 231: Do we only want to run this on nightly?

integration/test_resources/vuln_scan/dirty.Dockerfile
afiune: Wait, the latest Node container is dirty?!!! 😮

integration/container_vulnerability_test.go
```diff
@@ -31,10 +31,8 @@ import (
 )
 
 const (
-	registry   = "index.docker.io"
-	repository = "lacework/lacework-cli"
-	tag1       = "ubuntu-1804"
-	tag2       = "debian-10"
+	registry        = "index.docker.io"
+	cleanRepository = "techallylw/test-cli-clean"
```
afiune, line 35: How so you are using the clean docker image? I see it does have one vulnerability: 

https://4451-242235794-gh.circle-artifacts.com/0/circleci-artifacts/techallylw-test-cli-clean-sha256%3A7bf7cef98cceeb038437e7af713153ea28e93a00875d11f284b6f7c6dcb26ba9.html

Two questions:
* Shouldn't we need to assert that the clean image has NO vulnerabilities? (but it does 🤔)
* Shouldn't we be using the dirty image here instead?

integration/test_resources/vuln_scan/dirty.Dockerfile
dmurray-lacework: Yup, quick and easy way to get a vulnerable container. Just add NPM.

integration/container_vulnerability_test.go
dmurray-lacework: Only if the tests expect vulns we should use dirty container. It was vuln free initially. Having something guaranteed to be vuln free will be difficult. Best we can do is hope that patches happen quickly for new CVE's. Perhaps there's a base image with a focus on quickly patching new cve's?

dmurray-lacework: But you're right for 1 of the tests that rely on the assessment output from this scan expect vulns. So we should switch to to the dirty container.

afiune (approved):
tenor-264505136

PR #344, opened 2021-03-19
fix: Parse seconds in expiresAt response

dmurray-lacework:
Parse seconds in expiresAt response


dmurray-lacework:
The time parsing isn't the issue. The token response does not contain seconds :(

PR #345, opened 2021-03-21
ci: automatic minor version bump

afiune:
JIRA: ALLY-370


dmurray-lacework (approved):
👍

PR #348, opened 2021-03-22
ci: fix prepare-test-resources job

dmurray-lacework:
fix prepare-test-resources job


afiune (approved):
tenor-2546179

PR #349, opened 2021-03-22
feat(api): implement account information endpoint

afiune:
As a Lacework Engineer that needs to migrate tools and integrations from APIv1 to APIv2,
 I need to have a mechanism to identify both authentication methods,
So that I can start using both API versions and migrate endpoints gradually (not all at once).

JIRA ALLY-369

### DIFF
--- a/api/account.go
+++ b/api/account.go
@@ -24,7 +24,7 @@ type AccountService struct {
 	client *Client
 }
 
-func (svc *AccountService) OrganizationInfo() (
+func (svc *AccountService) GetOrganizationInfo() (
 	response accountOrganizationInfoResponse,
 	err error,
 ) {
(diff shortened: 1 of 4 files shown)


scottford-lw (approved):
👍

api/account.go
```diff
@@ -0,0 +1,42 @@
...
+
+package api
+
+// AccountService is a service that interacts with Account related
+// endpoints from the Lacework Server
+type AccountService struct {
+	client *Client
+}
+
+func (svc *AccountService) OrganizationInfo() (
```
afiune, line 27: Every function inside the Go-Client has to have an action, this was missing:
```suggestion
func (svc *AccountService) GetOrganizationInfo() (
```

Issue #211, opened 2020-09-16
Different exit codes for different results


scarolan:
User asked if we could do something like this:

0 script ran fine, no vulnerabilities
1 script had errors
2 vulnerability scan failed
etc.

he said this would make it even easier to integrate within ci/cd envs

scottford-lw:
@scarolan @afiune I think this is a great idea and something we should dig into. I am pretty sure we all know the standard Nagios style output code for monitoring checks...

```
0 - Service is OK.
1 - Service has a WARNING.
2 - Service is in a CRITICAL status.
3 - Service status is UNKNOWN.
```

When it comes to exit codes here I think we are going to need more coverage, or flexibility to configure exit codes. A quick example of some of the status might include what @scarolan documented but what about all of these statuses:

Critical Vulns no fix
Critical vulns with fixes

Thresholds:
High and above with fixes
High and above no fixes

i would like to talk this through with customers but I think we need to consider the possibility that each pipeline might need to have flexibility define what to do with results. For instance, if we are talking about the build pipeline for a service that is external facing service then it will likely have tighter policy than something not as critical. I don't have answers here, but I think we need to talk to our customers

afiune:
Closed by https://github.com/lacework/go-sdk/pull/352

PR #352, opened 2021-03-24
feat: add new flags to fail host and container vuln scans

dmurray-lacework:
## New Flags

- `--fail_on_severity <severity>`
- `--fail_on_fixable` 

## Description
Add 2 new flags to vulnerability host commands `scan-pkg-manifest` `show-assessment`
and vulnerability container commands `scan` `scan-status` `show-assessment`.
When set the CLI will exit with an error code `9` if the specified severity threshold is met.

## Example Usages:
`lacework vuln container scan <registry> <repository> <tag> --poll --fail_on_severity critical`
`lacework vuln ctr show <digest> --fail_on_severity high --fail_on_fixable`
`lacework vuln host scan-pkg-manifest <manifest> --fail_on_severity high --fail_on_fixable`


cli/cmd/vulnerability.go
```diff
@@ -371,6 +399,34 @@ func pollScanStatus(requestID string) error {
 	}
 }
 
+func vulnFailureThreshold(assessment api.HostVulnHostAssessment) {
+	severityRating, _ := eventSeverityToProperTypes(assessment.HighestSeverity())
```
dmurray-lacework, line 403: Move to vulnerabilities_host.go as` func (assessment *HostVulnHostAssessment) vulnFailureThreshold()`

cli/cmd/vulnerability.go
```diff
@@ -371,6 +399,34 @@ func pollScanStatus(requestID string) error {
...
+		os.Exit(1)
+	}
+
+	if !vulCmdState.FailOnFixable && severityRating <= threshold {
+		os.Exit(1)
+	}
+}
+
+func vulnContainerFailureThreshold(assessment *api.VulnContainerAssessment) {
+	severityRating, _ := eventSeverityToProperTypes(assessment.HighestSeverity())
```
dmurray-lacework, line 417: Move to vulnerabilities_container.go as` func (assessment **VulnContainerAssessment) vulnFailureThreshold()`

cli/cmd/vuln_container.go
```diff
@@ -61,6 +66,9 @@ Arguments:
 		Long:    "Check the status of an on-demand container vulnerability assessment.",
 		Args:    cobra.ExactArgs(1),
 		RunE: func(_ *cobra.Command, args []string) error {
+			if err := validateFailureFlags(); err != nil {
```
dmurray-lacework, line 69: Need to look into handling validation better. Not a fan of pasting this into all the commands the have this flag enabled.

cli/cmd/vulnerability.go
```diff
@@ -364,13 +394,67 @@ func pollScanStatus(requestID string) error {
...
+
+		if vulCmdState.FailOnSeverity != "" {
+			vulnContainerFailureThreshold(assessment)
+		}
+
 		return nil
 	}
 }
 
+func vulnFailureThresholdc(assessment api.HostVulnHostAssessment) {
```
dmurray-lacework, line 410: This needs fixed. The vulnFailureThreshold function is done twice for HostVulnHostAssessment and HostVulnScanPkgManifestResponse. See if this can be reduced to 1 function. If not rename the functions appropriately

api/vulnerabilities_host.go
```diff
@@ -195,6 +195,82 @@ func (assessment *HostVulnHostAssessment) VulnerabilityCounts() HostVulnCounts {
...
+		return "medium"
+	}
+	if vulnCounts.Low != 0 {
+		return "low"
+	}
+	return "unknown"
+}
+
+// HighestFixableSeverity returns the highest fixable severity level vulnerability in a HostVulnScanPkgManifestResponse
+func (scanPkgManifest *HostVulnScanPkgManifestResponse) HighestFixableSeverity() string {
```
dmurray-lacework, line 218: Lot of duplication here, see if this can be reduced

cli/cmd/vulnerability.go
```diff
@@ -270,6 +279,27 @@ func setDetailsFlag(cmds ...*flag.FlagSet) {
...
+					strings.Join(api.ValidEventSeverities, ", ")),
+			)
+		}
+	}
+}
+
+func setFailOnFixableFlag(cmds ...*flag.FlagSet) {
+	for _, cmd := range cmds {
+		if cmd != nil {
+			cmd.BoolVar(&vulCmdState.FailOnFixable, "fail_on_fixable", false,
```
dmurray-lacework, line 296: Is `fail_on_fixable` the right name for this? It's always used in conjunction with `fail_on_severity` so is the 'fail_on' redundant.

cli/cmd/vulnerability.go
```diff
@@ -270,6 +279,27 @@ func setDetailsFlag(cmds ...*flag.FlagSet) {
...
+				"fail if the assessed container has fixable vulnerabilities (use in conjunction with --fail_on_severity)",
```
afiune, line 297: Since this was very confusing internally, let's remove it until we find a better way to document that it is possible to use both flags together.
```suggestion
				"fail if the assessed container has fixable vulnerabilities",
```

cli/cmd/vuln_host.go
```diff
@@ -360,6 +387,18 @@ func init() {
 		vulHostShowAssessmentCmd.Flags(),
 	)
 
+	setFailOnSeverityFlag(
+		vulHostListCvesCmd.Flags(),
```
afiune, line 391: why are we adding these flags to the sub-command `list-cves`? 🤔

cli/cmd/vuln_host.go
```diff
@@ -360,6 +387,18 @@ func init() {
...
+		vulHostShowAssessmentCmd.Flags(),
+		vulHostScanPkgManifestCmd.Flags(),
+	)
+
+	setFailOnFixableFlag(
+		vulHostListCvesCmd.Flags(),
```
afiune, line 397: same 🤔

cli/cmd/vuln_container.go
```diff
@@ -340,6 +362,16 @@ func checkOnDemandContainerVulnerabilityStatus(reqID string) error {
 	if vulCmdState.Html {
 		return generateVulnAssessmentHTML(results)
```
afiune, line 363: With the current implementation you have, if a user passes the `--html` flag, we won't evaluate the fail_on flags. We will need to move the evaluation of those flags somewhere else that allow the user to use all the flags available.

cli/cmd/vulnerability.go
```diff
@@ -364,13 +394,51 @@ func pollScanStatus(requestID string) error {
...
+	fixableSeverityRating, _ := eventSeverityToProperTypes(assessment.fixableSeverityRating)
+	threshold, _ := eventSeverityToProperTypes(vulCmdState.FailOnSeverity)
+
+	if vulCmdState.FailOnSeverity == "" && vulCmdState.FailOnFixable && assessment.fixableVulCount > 0 {
+		cli.OutputHuman("\nFixable vulnerabilities found. Exiting with error code %d\n", 9)
+		os.Exit(9)
+	}
+
+	if vulCmdState.FailOnFixable && severityRating <= fixableSeverityRating {
+		cli.OutputHuman("\nFixable vulnerabilities found with threshold \"%s\". Exiting with error code %d\n", vulCmdState.FailOnSeverity, 9)
```
afiune, line 432:
```suggestion
		cli.OutputHuman("\nFixable vulnerabilities found with severity threshold \"%s\". Exiting with error code %d\n", vulCmdState.FailOnSeverity, 9)
```

cli/cmd/vulnerability.go
```diff
@@ -364,13 +394,51 @@ func pollScanStatus(requestID string) error {
...
+		os.Exit(9)
+	}
+
+	if !vulCmdState.FailOnFixable && severityRating <= threshold {
+		cli.OutputHuman("\nVulnerabilities found with threshold \"%s\". Exiting with error code %d\n", vulCmdState.FailOnSeverity, 9)
+		os.Exit(9)
```
afiune, line 438: I am not 100% sure that we want to exit spontaneously in different places of the CLI codebase. I fear we are going to lose track of the places/functions that could exit a command executing and we won't provide end-to-end features and functionalities. For example, the daily version check or telemetry could be lost here. 🤔

Let me think for a min how can we work something out so that we avoid doing these kinds of exit.

afiune: Additionally, there is already a function you could use to exit with an exit code `exitwithCode()` https://github.com/lacework/go-sdk/blob/main/cli/cmd/root.go#L277

cli/cmd/vulnerability.go
```diff
@@ -364,13 +394,51 @@ func pollScanStatus(requestID string) error {
...
+	fixableVulCount       int32
+}
+
+func (assessment *vulAssessment) validate() {
+	severityRating, _ := eventSeverityToProperTypes(assessment.severityRating)
+	fixableSeverityRating, _ := eventSeverityToProperTypes(assessment.fixableSeverityRating)
+	threshold, _ := eventSeverityToProperTypes(vulCmdState.FailOnSeverity)
+
+	if vulCmdState.FailOnSeverity == "" && vulCmdState.FailOnFixable && assessment.fixableVulCount > 0 {
+		cli.OutputHuman("\nFixable vulnerabilities found. Exiting with error code %d\n", 9)
```
afiune, line 427: Choosing to exit with `9` is fine. Though, we had already an idea/feature-request that suggests we should exit with different codes depending on the severity of the vulnerabilities found: https://github.com/lacework/go-sdk/issues/211

Maybe a quick example/idea would be to:
* exit with code `9` if we found `critical` vulnerabilities
* exit with code `8` if we found `high` vulnerabilities
* exit with code `7` if we found `medium` vulnerabilities
* exit with code `6` if we found `low` vulnerabilities

afiune (changes requested):
Looking pretty pretty gut! Let us just sync-up on how and where to exit the execution of the commands as well as defining the exit codes.

cli/cmd/vulnerability.go
dmurray-lacework: Will do!

cli/cmd/vuln_host.go
dmurray-lacework: No, originally planned to add that here but decided against it. Will remove that

cli/cmd/vuln_host.go
dmurray-lacework: Same as above, it was an idea to add it to list cve's aswell.

cli/cmd/vulnerability.go
dmurray-lacework: Mmm, i'm not sure on this. With `fail_on_severity `flag the user is providing the severity level already. So they should assume if they input `critical` the failure was due to a critical failing being found. Happy to discuss this more though.

cli/cmd/vulnerability.go
dmurray-lacework: I'll change these to exiwithCode in the mean time

cli/cmd/vulnerability.go
dmurray-lacework: Another point for discussion is to distinguish the `fail_on_severity` and `fail_on_fixable` with different error codes. Perhaps instead on different codes per vuln severity type.

cli/cmd/vuln_container.go
dmurray-lacework: mmm, I think setting the err here then returning it after would work. Question is do we want to continue with the vulnFailureValidate if the html output fails or not. I would say no, in which case the code would be.

`	cli.OutputHuman(buildVulnerabilityReportTable(results))
	if vulCmdState.Html {
		err = generateVulnAssessmentHTML(results)
	}

	if err !=nil {
		return err
	}

	if vulFailureFlagsEnabled() {
		vulnAssess := vulAssessment{
			severityRating:        results.HighestSeverity(),
			fixableSeverityRating: results.HighestFixableSeverity(),
			fixableVulCount:       results.FixableVulnerabilities,
		}
		vulnAssess.validate()
	}`

afiune: That is better 👆🏽

cli/cmd/vulnerability.go
afiune: The team agreed that this level of error code is too granular for our first iteration of this feature. We will stick to a single error code that is not `1`.

api/vulnerabilities_host.go
```diff
@@ -195,6 +195,40 @@ func (assessment *HostVulnHostAssessment) VulnerabilityCounts() HostVulnCounts {
...
+	}
+	if vulnCounts.Low != 0 {
+		return "low"
+	}
+	return "unknown"
+}
+
+// HighestFixableSeverity returns the highest fixable severity level vulnerability
+func HighestFixableSeverity(vulnCounts HostVulnCounts) string {
+	if vulnCounts.Critical != 0 {
```
afiune, line 217: All of these are wrong.
```suggestion
	if vulnCounts.CritFixable != 0 {
```

afiune (approved):
tenor-31259773

afiune (approved):
tenor-148158778

PR #354, opened 2021-03-26
Project Luna: "lacework query" CLI Command

hazedav:
### Lacework Query Commands
```
❯ lacework query
❯ lacework query sources
❯ lacework query describe <dataSource>
❯ lacework query list                      
❯ lacework query show <query_id>
❯ lacework query validate [query_id]
❯ lacework query run [queryID] --validate_only
❯ lacework query run [queryID]
❯ lacework query run -f <file>
❯ lacework query run -u <url>
❯ lacework query create
❯ lacework query create -f <file>
❯ lacework query create -u <url>
❯ lacework query update
❯ lacework query update -f <file>
❯ lacework query update -u <url>
❯ lacework query delete <query_id>
```

### Open Items
1. ~~Automated unit testing~~
2. ~~Reasonable defaults for start/end~~ (https://lacework.atlassian.net/browse/ALLY-427)
3. ~~Put back hidden flags~~
4. ~~Error handling for start/end~~
5. ~~Automated integration testing~~

### Open Questions
Q. Do we care for tabular output of query run?
A. Not at this time

Q. Do we want `--since` for human consumable time ranges ?
A. Not at this time (https://lacework.atlassian.net/browse/ALLY-427)

Q. Human readable output of create/update/delete.
A. Fine as is for now.


cli/cmd/lql_sources.go
```diff
@@ -0,0 +1,78 @@
...
+const (
+	lqlSourcesDebugMsg    string = "retrieving LQL data sources"
+	lqlSourcesNotFoundMsg string = "There were no data sources found.\n"
+	lqlSourcesUnableMsg   string = "unable to retrieve LQL data sources"
+)
+
+var (
+	// lqlSourcesCmd represents the lql data sources command
+	lqlSourcesCmd = &cobra.Command{
+		Use:   "sources",
```
afiune, line 35: I vote to name this command `list-sources` and have an alias for `sources`
```suggestion
		Use:     "list-sources",
		Aliases: []string{"sources"},
```

cli/cmd/lql_list.go
```diff
@@ -0,0 +1,80 @@
...
+const (
+	lqlListDebugMsg    string = "retrieving LQL queries"
+	lqlListNotFoundMsg string = "There were no queries found."
+	lqlListUnableMsg   string = "unable to retrieve LQL queries"
+)
+
+var (
+	// lqlListCmd represents the lql list command
+	lqlListCmd = &cobra.Command{
+		Use:   "list",
```
afiune, line 37: This `list` command looks good. I wouldn't change it to `list-queries`.

cli/cmd/lql_update.go
```diff
@@ -0,0 +1,72 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/lacework/go-sdk/api"
+)
+
+const (
+	lqlUpdateDebugMsg   string = "updating LQL query"
+	lqlUpdateSuccessMsg string = "LQL query (%v) updated successfully.\n"
```
rmoles, line 32: General Go question. Any reason to use `(%v)` here rather than being explicit with the type? eg. `(%s)`

hazedav: `%s` is probably better.  My use of `%v` is likely a result of the tutorials i was watching where they wanted to easily interpolate arbitrary data types...

api/lql.go
```diff
@@ -18,23 +18,115 @@
...
+
+		if err == nil {
+			q.StartTimeRange = t.StartTimeRange
+			q.EndTimeRange = t.EndTimeRange
+			q.QueryText = t.QueryText
+			// if QueryBlob is LQL
+		} else if matched, _ := regexp.MatchString(reLQL, q.QueryBlob); matched {
+			q.QueryText = q.QueryBlob
+		}
+	}
```
afiune, line 56: nit: we like to keep the nested code as little as possible, this could be change to

```suggestion
	// only if QueryText is empty
	if q.QueryText != "" {
		return
	}
	// if QueryBlob is JSON
	var t LQLQuery
	if err := json.Unmarshal([]byte(q.QueryBlob), &t); err == nil {
		q.StartTimeRange = t.StartTimeRange
		q.EndTimeRange = t.EndTimeRange
		q.QueryText = t.QueryText
		return
	}

	// if QueryBlob is LQL
	if matched, _ := regexp.MatchString(reLQL, q.QueryBlob); matched {
		q.QueryText = q.QueryBlob
	}
```

api/lql.go
```diff
@@ -18,23 +18,115 @@
...
 }
 
-type lqlQueryRequest struct {
-	QueryText string `json:"QUERY_TEXT"`
+func (svc *LQLService) CreateQuery(query string) (
+	response LQLQueryResponse,
+	err error,
+) {
+	lqlQuery := LQLQuery{QueryBlob: query}
+	lqlQuery.translate()
```
afiune, line 76: What happens if the `LQLQuery` struct is malformed? Meaning that if someone doesn't put any Query, should we throw an error inside the `translate()` function? 🤔

api/lql.go
```diff
@@ -18,23 +18,115 @@
...
+	response LQLQueryResponse,
+	err error,
+) {
+	var uri string
+
+	if queryID != "" {
+		uri = apiLQL + "?LQL_ID=" + url.QueryEscape(queryID)
+	} else {
+		uri = apiLQL
+	}
```
afiune, line 104: So here the `queryID` can be empty? Why and what will the response look like on an empty ID?

hazedav: If `queryID` is empty then we issue our `GET` without `LQL_ID` meaning we will get everything.  This is illustrated in `GetQueries()` above:

```
func (svc *LQLService) GetQueries() (
	response LQLQueryResponse,
	err error,
) {
	return svc.GetQueryByID("")
}
```

api/lql.go
```diff
@@ -18,23 +18,115 @@
...
+func (svc *LQLService) RunQuery(query, start, end string) (
 	response map[string]interface{},
 	err error,
 ) {
-	err = svc.client.RequestEncoderDecoder("POST",
+	lqlQuery := LQLQuery{
+		StartTimeRange: start,
+		EndTimeRange:   end,
+		QueryBlob:      query,
+	}
```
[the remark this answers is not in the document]
hazedav: We certainly can and there's pros and cons to both approaches.  The downside to doing validation in the SDK is that you have a validation disconnect with the underlying API and will need to remain aligned if/when the underlying api changes.  The upside is as you suggested...gated server requests...

api/lql.go
hazedav: > What happens if the LQLQuery struct is malformed?

In this case however the underlying api deals with an empty `QUERY_TEXT` value, but we can add error checking in the `translate()` function if so desired...

api/lql.go
afiune, line 104: If we support this empty ID case, we could recude the code like:
```suggestion
	var uri = apiLQL
	if queryID != "" {
		uri = apiLQL + "?LQL_ID=" + url.QueryEscape(queryID)
	} 
```

api/lql.go
```diff
@@ -18,23 +18,115 @@
...
+	} else {
+		uri = apiLQL
+	}
+
+	err = svc.client.RequestDecoder(
+		"GET",
+		uri,
+		nil,
+		&response,
+	)
```
afiune, line 111: nit: we break down the code when the line exceeds `100` characters. (more or less)
```suggestion
	err = svc.client.RequestDecoder("GET", uri, nil, &response)
```

cli/cmd/lql_update.go
hazedav: Fixed.

api/lql.go
hazedav: Fixed.

api/lql_delete.go
```diff
@@ -0,0 +1,53 @@
...
+	response LQLDeleteResponse,
+	err error,
+) {
+	var uri string
+
+	if queryID != "" {
+		uri = apiLQL + "?LQL_ID=" + url.QueryEscape(queryID)
+	} else {
+		uri = apiLQL
+	}
```
afiune, line 44: Same code improvement as above 👆🏽, we try to reduce the use of if.. else.. statements
```suggestion
	var uri = apiLQL
	if queryID != "" {
		uri = apiLQL + "?LQL_ID=" + url.QueryEscape(queryID)
	}
```

api/lql_delete.go
```diff
@@ -0,0 +1,53 @@
...
+
+	err = svc.client.RequestDecoder(
+		"DELETE",
+		uri,
+		nil,
+		&response,
+	)
```
afiune, line 51: nit:
```suggestion
	err = svc.client.RequestDecoder("DELETE", uri, nil, &response)
```

api/lql_describe.go
```diff
@@ -0,0 +1,70 @@
...
+	Props    map[string]interface{} `json:props`
+	Required bool                   `json:required`
+	Type     string                 `json:type`
+}
+
+type LQLDescribeSchema struct {
+	Name  string                 `json:name`
+	Props map[string]interface{} `json:props`
+	Type  string                 `json:type`
+}
```
afiune, line 54: All of the structs above are missing the `json:"name"` double quoting on the name 👆🏽

api/lql_describe.go
```diff
@@ -0,0 +1,70 @@
...
+
+func (svc *LQLService) Describe(dataSource string) (
+	response LQLDescribeResponse,
+	err error,
+) {
+	uri := "%s/%s"
+	uri = fmt.Sprintf(uri, apiLQLDescribe, url.QueryEscape(dataSource))
```
afiune, line 61: No need to do this in two lines
```suggestion
	uri := fmt.Sprintf("%s/%s", apiLQLDescribe, url.QueryEscape(dataSource))
```

api/lql_describe.go
```diff
@@ -0,0 +1,70 @@
...
+
+	err = svc.client.RequestDecoder(
+		"GET",
+		uri,
+		nil,
+		&response,
+	)
```
afiune, line 68:
```suggestion
	err = svc.client.RequestDecoder("GET", uri, nil, &response)
```

cli/cmd/lql.go
```diff
@@ -19,71 +19,217 @@
...
 
 var (
-	lqlFile string
+	lqlEnd      string
+	lqlEnv      bool
+	lqlFile     string
+	lqlRepo     bool
+	lqlStart    string
+	lqlURL      string
+	lqlValidate bool
```
afiune, line 42: We try to keep these flag states inside a single struct per command, just like this https://github.com/lacework/go-sdk/blob/main/cli/cmd/vulnerability.go#L35

```suggestion
	lqlCmdState = struct {
		End      string
		Env      bool
		File     string
		Repo     bool
		Start    string
		URL      string
		Validate bool
	}{}
```

cli/cmd/lql.go
```diff
@@ -19,71 +19,217 @@
...
+			return
+		}
+		defer response.Body.Close()
+
+		if response.StatusCode != 200 {
+			err = errors.Wrap(errors.New(response.Status), msg)
+			return
+		}
+
+		body, err = io.ReadAll(response.Body)
```
dmurray-lacework, line 206: ReadAll is from the ioutil package

```suggestion
		body, err = ioutil.ReadAll(response.Body)
```

cli/cmd/lql.go
```diff
@@ -19,71 +19,217 @@
 package cmd
 
 import (
+	"fmt"
+	"io"
```
dmurray-lacework, line 23: io package won't be used when line 206 is changed

```suggestion
```

api/lql.go
hazedav: Fixed.

api/lql.go
hazedav: Fixed.

api/lql_delete.go
hazedav: Fixed.

api/lql_delete.go
hazedav: Fixed.

api/lql_describe.go
hazedav: Fixed.

api/lql_describe.go
hazedav: Fixed.

api/lql_describe.go
hazedav: Fixed.

cli/cmd/lql.go
```diff
@@ -19,71 +19,217 @@
...
+		Use:     "query",
+		Short:   "Run and manage LQL queries",
+		Long: `Run and manage LQL queries.
+
+NOTE: This feature is not yet available!`,
+	}
+
+	// lqlRunCmd represents the lql run command
+	lqlRunCmd = &cobra.Command{
+		Use:   "run <query|queryID>",
```
afiune, line 56: We use `<param>` for a required parameter and `[param]` for an optional parameter.

We also use `query_id` underscores.

Example:

- https://github.com/lacework/go-sdk/blob/main/cli/cmd/vuln_container.go#L40
- https://github.com/lacework/go-sdk/blob/main/cli/cmd/vuln_container.go#L63

This should look like:
```suggestion
		Use:   "run [query|query_id]",
```

cli/cmd/lql.go
```diff
@@ -19,71 +19,217 @@
...
+			err = errors.Wrap(err, msg)
+			return
+		}
+		defer response.Body.Close()
+
+		if response.StatusCode != 200 {
+			err = errors.Wrap(errors.New(response.Status), msg)
+			return
+		}
+
```
dmurray-lacework, line 205: Quite a lot happening here. Anyway to extract into separate functions, logic might be easier to follow.

dmurray-lacework, line 205: Quite a lot happening here. Anyway to extract into separate functions, logic might be easier to follow.

cli/cmd/lql.go
```diff
@@ -19,71 +19,217 @@
...
 
-	// lqlCmd represents the lql command
+	// lqlCmd represents the lql parent command
 	lqlCmd = &cobra.Command{
-		Use:    "lql <query>",
-		Hidden: true,
-		Short:  "run an LQL query",
+		Aliases: []string{"lql"},
+		Use:     "query",
+		Short:   "Run and manage LQL queries",
```
afiune, line 48: short descriptions start with lowercase
```suggestion
		Short:   "run and manage LQL queries",
```

cli/cmd/lql.go
```diff
@@ -19,71 +19,217 @@
...
+	lqlValidate bool
 
-	// lqlCmd represents the lql command
+	// lqlCmd represents the lql parent command
 	lqlCmd = &cobra.Command{
-		Use:    "lql <query>",
-		Hidden: true,
-		Short:  "run an LQL query",
+		Aliases: []string{"lql"},
+		Use:     "query",
```
afiune, line 47: We must add that hidden flag before merge, please 😏

cli/cmd/lql.go
```diff
@@ -19,71 +19,217 @@
...
 
-	// file flag to specify a query from disk
-	lqlCmd.Flags().StringVarP(&lqlFile,
-		"file", "f", "",
-		"path to an LQL query to run",
+	// add sub-commands to the lql command
+	lqlCmd.AddCommand(lqlRunCmd)
+
+	// for commands that take query "text" as input
+	for _, cmd := range lqlQueryCommands {
```
afiune, line 107: Interesting approach, I am ok adding a common flag to multiple commands in this way, though we do follow this other way:

https://github.com/lacework/go-sdk/blob/main/cli/cmd/vuln_container.go#L259-L293

Totally up to you to update it or not. 😉

cli/cmd/lql.go
hazedav: Are you considering this optional because we will open the editor?

api/lql.go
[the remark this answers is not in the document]
afiune: I could NOT agree more with you, I wouldn't want to add validations in the SDK that could change in the future, though, this validation is just to check that the provider START time is before the END time. I don't think we will ever change that in the API. I hope. 😅

cli/cmd/lql_create.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/lacework/go-sdk/api"
+)
+
+const (
+	lqlCreateDebugMsg    string = "creating LQL query"
+	lqlCreateNotFoundMsg string = "Query created succesfully but not returned.\n"
```
dmurray-lacework, line 32: nitpick: typo
```suggestion
	lqlCreateNotFoundMsg string = "Query created successfully but not returned.\n"
```

api/api.go
```diff
@@ -62,7 +62,11 @@ const (
 	apiAccountOrganizationInfo = "external/account/organizationInfo"
 
 	// Alpha
-	apiLQLQuery = "external/lql/query"
+	apiLQL            = "external/lql"
+	apiLQLCompile     = "external/lql/compile"
+	ApiLQLDataSources = "external/lql/dataSources"
```
afiune, line 67: Why this changed to be exported? 
```suggestion
	apiLQLDataSources = "external/lql/dataSources"
```

api/lql_data_sources.go
```diff
@@ -0,0 +1,33 @@
...
+	Data    []string `json:"data"`
+	Ok      bool     `json:"ok"`
+	Message string   `json:"message"`
+}
+
+func (svc *LQLService) DataSources() (
+	response LQLDataSourcesResponse,
+	err error,
+) {
+	err = svc.client.RequestDecoder("GET", ApiLQLDataSources, nil, &response)
```
afiune, line 31:
```suggestion
	err = svc.client.RequestDecoder("GET", apiLQLDataSources, nil, &response)
```

cli/cmd/lql_create.go
```diff
@@ -0,0 +1,77 @@
...
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/lacework/go-sdk/api"
+)
+
+const (
```
dmurray-lacework, line 30: nitpick: Some of these output messages start with an upper case letter and some with lower case. Would be nice to have them consistent.

api/lql_delete.go
```diff
@@ -0,0 +1,46 @@
...
+	"net/url"
+)
+
+type LQLDeleteResponse struct {
+	Ok      bool             `json:"ok"`
+	Message LQLDeleteMessage `json:"message"`
+}
+
+type LQLDeleteMessage struct {
+	ID string `json:"lqlDeleted"`
```
afiune, line 31: ugh, this struct is confusing but I understand that this is how the API response looks like... 🤷🏽‍♂️

cli/cmd/lql.go
hazedav: Absolutely...I have this as number 3 in my open item list.

cli/cmd/lql.go
hazedav: Seems legit.  Let me give a a spin.

api/api.go
hazedav: So that I can access this via `api.ApiLQLDataSources` in my test file:

```
func TestDataSourcesMethod(t *testing.T) {
	fakeServer := lacework.MockServer()
	fakeServer.MockAPI(
		api.ApiLQLDataSources,
```

api/lql_data_sources.go
hazedav: So that I can access this via `api.ApiLQLDataSources` in my test file:

```
func TestDataSourcesMethod(t *testing.T) {
	fakeServer := lacework.MockServer()
	fakeServer.MockAPI(
		api.ApiLQLDataSources,
```

cli/cmd/lql_create.go
hazedav: ?  They should all start with `lql`...

api/lql_delete.go
hazedav: Pretty wild 🐅

cli/cmd/lql.go
hazedav: Good eye.  No idea why this actually works...

cli/cmd/lql.go
hazedav: We can.

cli/cmd/lql.go
hazedav: I think optional makes sense here.  I'll update everything as needed.

api/lql.go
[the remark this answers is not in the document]
hazedav: I'll wire up some checks!

cli/cmd/lql.go
hazedav: Fixed.

cli/cmd/lql.go
hazedav: Fixed.

cli/cmd/lql.go
hazedav: Fixed.

cli/cmd/lql.go
hazedav: Fixed.

cli/cmd/lql.go
hazedav: Fixed.

cli/cmd/lql.go
hazedav: Fixed.

cli/cmd/lql_create.go
hazedav: Fixed.

cli/cmd/lql.go
```diff
@@ -19,71 +19,215 @@
...
-		if err != nil {
-			return err
-		}
+		err = survey.AskOne(prompt, &query)
 	}
 
-	cli.Log.Debugw("running LQL query", "query", query)
-	response, err := cli.LwApi.LQL.Query(query)
+	return
+}
```
afiune, line 225: The logic in this function has to be reduced. I will review it once the other comments are resolved, specially the one that tries to remove the `--env` flag.

cli/cmd/lql.go
```diff
@@ -94,5 +238,28 @@ func runLQLQuery(_ *cobra.Command, args []string) error {
...
+
+	if lqlCmdState.ValidateOnly {
+		// validate_only should compile
+		return CompileQueryAndOutput(query)
+	} else {
+		// !validate_only should should run
+		response, err = cli.LwApi.LQL.RunQuery(query, lqlCmdState.Start, lqlCmdState.End)
+	}
+
+	return output(response, err, msg)
```
afiune, line 264: Here is an example of a more readable code, maybe you can try to replicate these modifications to other functions? If not, I can continue adding suggestions 😉

```suggestion
	query, err := inputQuery(cmd, args)
	if err != nil {
		return errors.Wrap(err, "unable to run LQL query")
	}

	cli.Log.Debugw("running LQL query", "query", query)

	if lqlCmdState.ValidateOnly {
		return CompileQueryAndOutput(query)
	}
	
	response, err := cli.LwApi.LQL.RunQuery(query, lqlCmdState.Start, lqlCmdState.End)
	if err != nil {
		return errors.Wrap(err, "unable to run LQL query")
	}

	if data, ok := response["data"]; ok {
		return cli.OutputJSON(data)
	}
	
	if err := cli.OutputJSON(response); err != nil {
		return errors.Wrap(err, "unable to format json response")
	}
	return nil
```

cli/cmd/lql_create.go
afiune: Chiming in here. A bigger question is: Do we need do define constants for debug statements? 🤔 My gut tells me not but I am curious to know why you defined them as constants.

cli/cmd/lql_create.go
```diff
@@ -0,0 +1,79 @@
...
+	if cli.JSONOutput() {
+		return cli.OutputJSON(create.Data)
+	}
+	if len(create.Data) == 0 {
+		cli.OutputHuman(lqlCreateNotFoundMsg)
+	} else {
+		cli.OutputHuman(
+			fmt.Sprintf(lqlCreateSuccessMsg, create.Data[0].ID))
+	}
+	return nil
```
afiune, line 78: Here is one more suggestion, I'll stop now and let you review the rest of the functions:
```suggestion
	query, err := inputQuery(cmd, args)
	if err != nil {
		return errors.Wrap(err, "unable to create LQL query")
	}

	cli.Log.Debugw("creating LQL query", "query", query)
	create, err = cli.LwApi.LQL.CreateQuery(query)
	if err != nil {
		return errors.Wrap(err, "unable to create LQL query")
	}

	if cli.JSONOutput() {
		return cli.OutputJSON(create.Data)
	}
	if len(create.Data) == 0 {
		// This should not be an error?
		return cli.OutputHuman("Query created successfully but not returned")
	}
	
	return cli.OutputHuman(fmt.Sprintf(
		"LQL query '%s' created successfully.\n",
		create.Data[0].ID,
	))
```

cli/cmd/lql_delete.go
```diff
@@ -0,0 +1,75 @@
...
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+const (
+	lqlDeleteBadInputMsg string = "Please specify a valid query ID."
+	lqlDeleteDebugMsg    string = "deleting LQL query"
+	lqlDeleteSuccessMsg  string = "LQL query (%s) deleted successfully.\n"
+	lqlDeleteUnableMsg   string = "unable to delete LQL query"
```
afiune, line 32: Same here, my opinion would be not do define constants for messages, but happy to hear the why behind this. 😉

afiune (commented):
tenor-194167493

We are getting there @hazedav !!! Left a number of comments. Will come back once you have resolved them all. 🙌🏽

cli/cmd/lql_create.go
hazedav: These are strings that either print, or log.  They are constant and IMO easier to modify when declared in this matter.  Furthermore most of these sub commands follow a similar pattern for which it was easy to just declare sub command specific messages that drop into the code...

cli/cmd/lql.go
```diff
@@ -19,71 +19,217 @@
...
+		err = survey.AskOne(prompt, &query)
 	}
 
-	cli.Log.Debugw("running LQL query", "query", query)
-	response, err := cli.LwApi.LQL.Query(query)
+	return
+}
+
+// standardized cli/error output
+func output(response map[string]interface{}, err error, msg string) error {
```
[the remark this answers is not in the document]
hazedav: This function was used by everything at one time up until things were broken out and then we layered on both json and human readable formats.  I can just roll this back into the `runQuery()`. 

With respect to human-readable output, what do you propose here?  Asking because `EVENT` is a massive AWS Json blob....

cli/cmd/lql.go
```diff
@@ -19,71 +19,217 @@
...
+		)
+	}
+
+	// run specific flags
+	// env flag to specify a query from disk
+	lqlRunCmd.Flags().BoolVarP(
+		&lqlEnv,
+		"env", "e", false,
+		"run an LQL query by ID (using active profile)",
+	)
```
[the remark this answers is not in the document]
hazedav: Would you want `lacework lql run <query_id> -r` to "run a query by active repo"?

The reason I ask is because for `--file` you specify a file path, and for `--url` you specify a url.  But for `--repo` you're specifying a `query_id`... (since the active repo is a configure thing)...

cli/cmd/lql.go
hazedav: Fixed.

cli/cmd/lql.go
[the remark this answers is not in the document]
hazedav: Fixed.

cli/cmd/lql.go
```diff
@@ -94,5 +240,28 @@ func runLQLQuery(_ *cobra.Command, args []string) error {
...
+
+	cli.Log.Debugw("running LQL query", "query", query)
+
+	if lqlValidate {
+		// validate_only should compile
+		return CompileQueryAndOutput(query)
+	} else {
+		// !validate_only should should run
+		response, err = cli.LwApi.LQL.RunQuery(query, lqlStart, lqlEnd)
+	}
```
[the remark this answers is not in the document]
hazedav: Fixed.

cli/cmd/lql.go
hazedav: Fixed.

cli/cmd/lql_delete.go
hazedav: Partially discussed above.  Maybe this is more common in python where you would just set up a class of constant variables like this for output/logging messages.

1. This allows these to easily be referred to in test code if needed.
2. Allows for modification w/o actually touching function code.
3. Makes for easy templating of like functions

cli/cmd/lql_create.go
hazedav: There seems to be an open question as to whether or not a success from the API w/ unexpected data should be an error or not?
```
	if len(create.Data) == 0 {
		// This should not be an error?
		return cli.OutputHuman("Query created successfully but not returned")
```

Personally I think this output is transparent as to what is happening.  The only thing we're doing "special" is outputting the ID of the query that was updated.  If this isn't important we can simplify the output to `LQL query created successfully`...

cli/cmd/lql.go
hazedav: Fixed.

api/lql.go
hazedav: Fixed.

cli/cmd/lql_create.go
hazedav: I cleaned this up quite a bit...

cli/cmd/lql.go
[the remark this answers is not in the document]
hazedav: For now we decided not to worry about tabular output for query run...

cli/cmd/lql_delete.go
hazedav: @afiune and I discussed the merits of this approach and we're sorta on the same page that there are pro's and con's to doing it like this vs. not.  He was reasonable enough to allow this approach to be pursued for now and potentially changed in the future.

cli/cmd/lql_create.go
hazedav: @afiune and I discussed the merits of this approach and we're sorta on the same page that there are pro's and con's to doing it like this vs. not. He was reasonable enough to allow this approach to be pursued for now and potentially changed in the future.

api/lql.go
```diff
@@ -18,24 +18,177 @@
...
+	// if QueryBlob is LQL
+	if matched, _ := regexp.MatchString(reLQL, q.QueryBlob); matched {
+		q.QueryText = q.QueryBlob
+		return
+	}
+
+	return errors.New(LQLQueryTranslateError)
+}
+
+func (q *LQLQuery) Validate(allowEmptyTimes bool) error {
```
afiune: Nit!! 

For readability in Go we prefer to create multiple functions to be explicit about their use,
for example, this `Validate()` function could have been two functions:

- `Validate()` or `ValidateEnforceTimes()`
- `ValidateAllowEmptyTimes()`

This will avoid us devs having to go look at the function and understand what
the boolean argument does.

cli/cmd/lql_delete.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+func deleteQuery(_ *cobra.Command, args []string) error {
+	var queryID string
+
+	if len(args) != 0 && args[0] != "" {
+		queryID = args[0]
+	} else {
+		return errors.Wrap(
+			errors.New(api.LQLDeleteBadInputMsg),
```
afiune, line 59: This is a big no-no for me. Using a defined message from the API in the CLI, as I mentioned before,
this kind of error, log, user-facing messages defined as constant could end up biting us in the but.

I'm ok if you define messages within the same Go package, but reusing messages between packages...

Let's not do this. Can we please make all constants private, please? (just define them with the first
letter as lowercase, not uppercase)

api/lql_delete.go
```diff
@@ -0,0 +1,52 @@
...
+package api
+
+import (
+	"net/url"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	LQLDeleteBadInputMsg string = "Please specify a valid query ID."
```
afiune, line 28: Is this an API bad input message or a CLI message? 🤔

api/lql_delete.go
```diff
@@ -0,0 +1,52 @@
...
+
+func (svc *LQLService) DeleteQuery(queryID string) (
+	response LQLDeleteResponse,
+	err error,
+) {
+	if queryID == "" {
+		err = errors.New(LQLDeleteBadInputMsg)
+		return
+	}
+	var uri string = ApiLQL + "?LQL_ID=" + url.QueryEscape(queryID)
```
afiune, line 48: No need to define this variable as a `string`,  the compiler will detect it.

Also, for string manipulation it is best to use the `fmt` package:
```suggestion
	uri := fmt.Sprintf("%s?LQL_ID=%s", ApiLQL, url.QueryEscape(queryID))
```

cli/cmd/lql_show.go
```diff
@@ -0,0 +1,68 @@
...
+package cmd
+
+import (
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+const (
+	lqlShowDebugMsg    string = "retrieving LQL query"
+	lqlShowNotFoundMsg string = "There were no queries found."
```
afiune, line 28: This constant is never used. Another reason why this model is a bit confusing.

FYI, the linter will complain until you either use it or delete it. 😅
```
cli/cmd/lql_show.go:28:2: `lqlShowNotFoundMsg` is unused (deadcode)                                                                             
        lqlShowNotFoundMsg string = "There were no queries found."                                                                              
        ^                                                                                                                                       
```

cli/cmd/errors.go
```diff
@@ -110,3 +111,10 @@ func (e *vulnerabilityPolicyError) validate() bool {
 	e.ExitCode = 0
 	return true
 }
+
+func yikes(msg string) error {
```
afiune, line 115: I LOOOOVE THIS!!

tenor-156234403

cli/cmd/lql.go
afiune: Still missing 😂 (kind reminder)

afiune (commented):
Done with the API, will move to the CLI changes.

cli/cmd/lql_show.go
hazedav: I got the linter going locally now, so we'll address this.  I thought unused things result in compiler errors.  Perhaps that's a var vs. const difference?

api/lql_delete.go
hazedav: Fixing.

api/lql_delete.go
hazedav: API

api/lql.go
```diff
@@ -18,24 +18,177 @@
...
+func (q *LQLQuery) Translate() (returnError error) {
+	// if QueryText is populated; return
+	if q.QueryText != "" {
+		return
+	}
+
+	// if QueryBlob is JSON
+	var t LQLQuery
+
+	if err := json.Unmarshal([]byte(q.QueryBlob), &t); err == nil {
```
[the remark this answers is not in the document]
hazedav: No.  From an end user standpoint the fact that we tried to unmarshal something that wasn't JSON is irrelevant here (since we're dealing with a Blob). We would want a consistent err msg experience from this fun.

api/lql.go
```diff
@@ -18,24 +18,177 @@
...
+	ID             string `json:"LQL_ID,omitempty"`
+	StartTimeRange string `json:"START_TIME_RANGE,omitempty"`
+	EndTimeRange   string `json:"END_TIME_RANGE,omitempty"`
+	QueryText      string `json:"QUERY_TEXT"`
+	// QueryBlob is a special string that supports type conversion
+	// back and forth from LQL to JSON
+	QueryBlob string `json:"-"`
+}
+
+func (q *LQLQuery) Translate() (returnError error) {
```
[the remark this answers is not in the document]
hazedav: I will:
1. Use fewer named returns
2. In the event that a named returned simplifies things I will still be more explicit about what we are returning as opposed to using blank `return`

cli/cmd/lql_delete.go
hazedav: Sure...

api/lql.go
hazedav: This passes through to multiple functions, so instead of 2 functions you'll have 4 functions that do basically the same thing...

api/lql.go
```diff
@@ -18,24 +18,177 @@
...
+		return
+	}
+
+	// if QueryBlob is LQL
+	if matched, _ := regexp.MatchString(reLQL, q.QueryBlob); matched {
+		q.QueryText = q.QueryBlob
+		return
+	}
+
+	return errors.New(LQLQueryTranslateError)
```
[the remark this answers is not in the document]
hazedav: We don't want to wrap anything here...

cli/cmd/lql_delete.go
hazedav: I will simply test for `NotNil` instead of error msg

api/lql_delete.go
hazedav: Fixed.

cli/cmd/lql_delete.go
hazedav: Fixed.

api/lql.go
```diff
@@ -18,24 +18,177 @@
...
+			q.EndTimeRange = t.EndTimeRange
+		}
+		q.QueryText = t.QueryText
+		return
+	}
+
+	// if QueryBlob is LQL
+	if matched, _ := regexp.MatchString(reLQL, q.QueryBlob); matched {
+		q.QueryText = q.QueryBlob
+		return
```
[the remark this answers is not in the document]
hazedav: Fixed.

api/lql.go
```diff
@@ -18,24 +18,177 @@
...
+
+	if err := json.Unmarshal([]byte(q.QueryBlob), &t); err == nil {
+		if q.StartTimeRange == "" {
+			q.StartTimeRange = t.StartTimeRange
+		}
+		if q.EndTimeRange == "" {
+			q.EndTimeRange = t.EndTimeRange
+		}
+		q.QueryText = t.QueryText
+		return
```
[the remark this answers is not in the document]
hazedav: Fixed.

api/lql.go
```diff
@@ -18,24 +18,177 @@
...
+	QueryText      string `json:"QUERY_TEXT"`
+	// QueryBlob is a special string that supports type conversion
+	// back and forth from LQL to JSON
+	QueryBlob string `json:"-"`
+}
+
+func (q *LQLQuery) Translate() (returnError error) {
+	// if QueryText is populated; return
+	if q.QueryText != "" {
+		return
```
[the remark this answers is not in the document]
hazedav: Fixed.

api/lql.go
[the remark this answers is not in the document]
hazedav: Fixed.

cli/cmd/vuln_html.go
```diff
@@ -186,7 +186,7 @@ func calcHtmlBarChartWidth(severity string, htmlData vulnImageAssessmentHtml) st
 	case "info":
 		return fmt.Sprintf("%f%%", htmlData.InfoVulnPercent())
 	default:
-		return fmt.Sprintf("0")
+		return "0"
```
hazedav, line 189: @afiune this lint failed for me locally, so I went and fixed here if that's cool...

afiune: Awesome! I wonder why we didn't catch this before.... 🤔

# Thank you
tenor-240583896

api/lql.go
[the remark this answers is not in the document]
afiune: I disagree. If we are trying to give a really amazing experience creating and running queries from the CLI, we must provide the reason why we couldn't translate a query. This is a major problem we have had where we assume that by saying `"unable to translate query blob"` the user will know what to do next. They won't. They will be confused. If we can tell them why we couldn't translate the query, we have to tell them, not hide it, this will allow them to troubleshoot the problem on their own.

I just ran a quick exercise where I am trying to run a simple query:
Screen Shot 2021-04-02 at 6 02 27 PM

I type my query:
Screen Shot 2021-04-02 at 6 00 40 PM

And, run it!
Screen Shot 2021-04-02 at 6 13 55 PM

What did I do wrong? How can I troubleshoot this problem? There is nothing....

Instead, if we bubble-up the error message to the end-user, we will get messages like:
```
ERROR unable to run LQL query: unable to translate query blob: invalid character '"' after object key:value pair
```

```
ERROR unable to run LQL query: unable to translate query blob: unexpected end of JSON input
```

```
ERROR unable to run LQL query: unable to translate query blob: invalid character ']' looking for beginning of value
```

This 👆🏽 in my opinion is more helpful. I am happy to jump into a zoom with you to talk about it further. 😈

integration/lql_test.go
```diff
@@ -0,0 +1,169 @@
...
+	// run
+	out, stderr, exitcode := LaceworkCLIWithTOMLConfig(
+		"query", "run", "-u", lqlQueryURL, "--start", lqlQueryStart, "--end", lqlQueryEnd)
+	assert.Contains(t, out.String(), `"INSERT_ID"`)
+	assert.Empty(t, stderr.String(), "STDERR should be empty")
+	assert.Equal(t, 0, exitcode, "EXITCODE is not the expected one")
+}
+
+func TestQueryRunValidateOnly(t *testing.T) {
+	if env := os.Getenv("CI_BETA"); env == "" {
```
afiune, line 160: you should be able to do this instead:
```suggestion
	if os.Getenv("CI_BETA") == "" {
```

api/lql.go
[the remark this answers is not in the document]
hazedav: I couldn't agree more that we can provide better messaging.  However we must keep in mind that there are multiple query formats in play.  So if you mess up a text-based query, it doesn't matter that this doesn't parse in JSON.  If you mess up a JSON-based query it doesn't matter that it doesn't match the text-based regex.  So when there's completely arbitrary text I'm not certain how specific we can be about what you specifically messed up...

integration/lql_test.go
hazedav: Man was going whole hog on the inline assignment stuff, but obviously when it only returns a single value and you don't need an assignment this is obv more clear.

hazedav: actually, `os.Getenv` does not return a boolean, which is why I was doing this...

hazedav: Fixed.

cli/cmd/vuln_html.go
hazedav: @afiune; Perhaps this is a difference in `golangci-lint` versions.  brew gave me `golangci-lint/1.39.0`...

api/lql.go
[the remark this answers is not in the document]
hazedav: @afiune , I think we are in agreement here that we want to improve this.  I went ahead and filed https://lacework.atlassian.net/browse/ALLY-432 to make that happen.  Let me know if you feel that this is blocking for the PR.

api/lql.go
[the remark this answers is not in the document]
hazedav: This will be improved via https://lacework.atlassian.net/browse/ALLY-432

cli/cmd/lql.go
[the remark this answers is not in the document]
hazedav: https://lacework.atlassian.net/browse/ALLY-423

cli/cmd/lql.go
hazedav: Fixed.

hazedav:
This is ready for final review!

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+	// query
+	if err := q.TranslateQuery(); err != nil {
+		return err
+	}
+	// start
+	if start, err := q.TranslateTime(q.StartTimeRange); err != nil {
+		return err
+	} else {
+		q.StartTimeRange = start
+	}
```
afiune, line 72: This else is not needed since you are returning on line `69`
```suggestion
	}
	q.StartTimeRange = start
```

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+	// end
+	if end, err := q.TranslateTime(q.EndTimeRange); err != nil {
+		return nil
+	} else {
+		q.EndTimeRange = end
+	}
```
afiune, line 78: same here
```suggestion
	}
	q.EndTimeRange = end
```

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+
+func (q LQLQuery) TranslateTime(inTime string) (outTime string, err error) {
+	// empty
+	if inTime == "" {
+		return
+	}
+	// parse time as RFC3339
+	if t, err := time.Parse(time.RFC3339, inTime); err == nil {
+		outTime = t.UTC().Format(time.RFC3339)
+		return outTime, err
```
afiune, line 116: when using named returns do not pass fields to the `return` statement
```suggestion
		return
```

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+	}
+	// parse time as millis
+	if t, err := strconv.ParseInt(inTime, 10, 64); err == nil {
+		outTime = time.Unix(0, t*int64(time.Millisecond)).UTC().Format(time.RFC3339)
+		return outTime, err
```
afiune, line 121: same here:
```suggestion
		return
```

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+	err = errors.New("unable to parse time (" + inTime + ")")
+	return
+}
+
+func (q LQLQuery) ValidateRange(allowEmptyTimes bool) (err error) {
+	// validate start
+	var start time.Time
+	if q.StartTimeRange != "" {
+		if start, err = time.Parse(time.RFC3339, q.StartTimeRange); err != nil {
+			return err
```
afiune, line 132: same here:
```suggestion
			return
```

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+		}
+	} else if allowEmptyTimes {
+		start = time.Unix(0, 0)
+	} else {
+		return errors.New("start time must not be empty")
```
afiune, line 137:
```suggestion
		err = errors.New("start time must not be empty")
		return
```

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+	}
+	// validate end
+	var end time.Time
+	if q.EndTimeRange != "" {
+		if end, err = time.Parse(time.RFC3339, q.EndTimeRange); err != nil {
+			return err
```
afiune, line 143:
```suggestion
			return
```

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+		}
+	} else if allowEmptyTimes {
+		end = time.Now()
+	} else {
+		return errors.New("end time must not be empty")
```
afiune, line 148:
```suggestion
		err = errors.New("end time must not be empty")
		return
```

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+	}
+	// validate range
+	if start.After(end) {
+		return errors.New("date range should have a start time before the end time")
```
afiune, line 152: apologies but if you are using "named returns" this is an antipattern, either use them or don't. 😅
```suggestion
		err = errors.New("date range should have a start time before the end time")
		return
```

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+	}
+	return nil
```
afiune, line 154:
```suggestion
	return
```

api/lql.go
```diff
@@ -18,24 +18,201 @@
...
+	}
+
+	err = svc.client.RequestEncoderDecoder("POST", ApiLQL, lqlQuery, &response)
+	return
+}
+
+func (svc *LQLService) GetQueries() (
+	response LQLQueryResponse,
+	err error,
+) {
```
afiune, line 185: here you are not using "named returns", clean the function definition
```suggestion
func (svc *LQLService) GetQueries() (LQLQueryResponse, error) {
```
This is more readable 👆🏽 - right?

api/lql_delete.go
```diff
@@ -0,0 +1,52 @@
...
+package api
+
+import (
+	"net/url"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	lqlDeleteBadInputMsg string = "Please specify a valid query ID."
```
afiune, line 28: These kinds of messages look like are meant to be for the Lacework CLI and not the API client.
All messages you have defined for API client start with a lower case and without a period in the end,
this is one that doesn't follow the pattern but it does follow the pattern we follow for CLI messages.

```suggestion
	lqlDeleteBadInputMsg string = "query ID must be provided"
```

api/lql_delete.go
```diff
@@ -0,0 +1,52 @@
...
+	response LQLDeleteResponse,
+	err error,
+) {
+	if queryID == "" {
+		err = errors.New(lqlDeleteBadInputMsg)
+		return
+	}
+	uri := ApiLQL + "?LQL_ID=" + url.QueryEscape(queryID)
+
+	err = svc.client.RequestDecoder("DELETE", uri, nil, &response)
```
afiune, line 50: nit:
```suggestion

	err = svc.client.RequestDecoder("DELETE",
		fmt.Sprintf("%s?LQL_ID=%s", ApiLQL, url.QueryEscape(queryID)),
		nil, &response,
	)
```

afiune (changes requested):
Just finished with the API changes..... I will come back to review the CLI changes...

tenor-52498932

api/lql_test.go
```diff
@@ -0,0 +1,632 @@
...
+			if err == nil {
+				assert.Equal(t, lqlValidateRangeTest.Return, err)
+			} else {
+				assert.Equal(t, lqlValidateRangeTest.Return, err.Error())
+			}
+		})
+	}
+}
+
+func TestValidate(t *testing.T) {
```
afiune, line 337: This test function name is too general, there could be a number of `Validate()` functions that we can test, this one, in particular, is testing the validation inside the LQL query, let's be specific to avoid collisions in the future.
```suggestion
func TestLQLQueryValidate(t *testing.T) {
```
We should also change these functions to reference the `LQL` service,
otherwise we won't know what are we testing:

- func TestTranslateTime(t *testing.T) {
- func TestValidateRange(t *testing.T) {
- func TestCreateMethod(t *testing.T) {
- func TestCreateBadInput(t *testing.T) {
- func TestCreateOK(t *testing.T) {
- func TestCreateError(t *testing.T) {
- func TestGetQueriesMethod(t *testing.T) {
- func TestGetQueryByIDOK(t *testing.T) {
- func TestGetQueryByIDNotFound(t *testing.T) {
- func TestRunQueryMethod(t *testing.T) {
- func TestRunQueryBadInput(t *testing.T) {
- func TestRunQueryOK(t *testing.T) {
- func TestRunQueryError(t *testing.T) {

A suggestion would be to start the test function to `TestLQL...`

api/lql_test.go
```diff
@@ -0,0 +1,632 @@
...
+	Input    *api.LQLQuery
+	Return   interface{}
+	Expected *api.LQLQuery
+}
+
+func TestTranslateTime(t *testing.T) {
+	for _, lqlTranslateTimeTest := range lqlTranslateTimeTests {
+		t.Run(lqlTranslateTimeTest.Name, func(t *testing.T) {
+			outTime, err := api.LQLQuery{}.TranslateTime(lqlTranslateTimeTest.Input)
+			if err == nil {
```
afiune, line 312: Inside tests, we normally do `err == nil` or `err != nil`, instead we use:
```suggestion
			if assert.Nil(t, err) {
```

api/lql.go
hazedav: I thought named returns weren't developer friendly because you have to track what you're returning.  This seems like the best of both worlds since it makes it more obvious.  At this point the named returns are just a means of initializing the return variables...

api/lql_test.go
hazedav: I don't want to `assert.Nil` here, I simply want different assert logic when `err == nil`...

api/lql.go
hazedav: `else` is needed because `start` is assigned via `if`.

api/lql.go
hazedav: `else` is needed because end is assigned via `if`

api/lql.go
hazedav: Fixed.

api/lql_delete.go
hazedav: Fixed.

api/lql_delete.go
hazedav: Fixed.

api/lql_test.go
hazedav: Fixed.

api/lql.go
hazedav: If this is something that we don't want to do, I can address this via the entire PR we don't need to comment each one.

hazedav: Fixed.

api/lql.go
hazedav: Fixed.

api/lql.go
hazedav: Fixed.

api/lql.go
hazedav: Fixed.

api/lql.go
hazedav: Fixed.

api/lql.go
hazedav: Fixed.

api/lql.go
hazedav: Fixed.

api/lql.go
hazedav: Fixed.

cli/cmd/lql_delete.go
```diff
@@ -0,0 +1,69 @@
...
+			lqlDeleteUnableMsg,
+		)
+	}
+
+	cli.Log.Debugw("deleting LQL query", "queryID", queryID)
+	delete, err := cli.LwApi.LQL.DeleteQuery(queryID)
+
+	if err != nil {
+		return errors.Wrap(err, lqlDeleteUnableMsg)
+	}
```
afiune, line 62: Once you have used `ExactArgs(1)` above, you can simplify this code like this:

```suggestion
	cli.Log.Debugw("deleting LQL query", "queryID", args[0])
	delete, err := cli.LwApi.LQL.DeleteQuery(args[0])
	if err != nil {
		return errors.Wrap(err, "unable to delete LQL query")
	}
```

You already codified the check of an empty query ID in the API LQL `DeleteQuery()` function.

Much much cleaner, right? 🎉

cli/cmd/lql_show.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	// lqlShowCmd represents the lql show command
+	lqlShowCmd = &cobra.Command{
+		Use:   "show <query_id>",
+		Short: "show an LQL query",
+		Long:  `Show an LQL query.`,
+		Args:  cobra.MaximumNArgs(1),
```
afiune, line 32: Lets change all other commands that have a required argument to follow this pattern:
```suggestion
		Args:  cobra.ExactArgs(1),
```

cli/cmd/lql_show.go
```diff
@@ -0,0 +1,69 @@
...
+		return errors.Wrap(
+			errors.New("Please specify a valid query ID."),
+			lqlShowUnableMsg,
+		)
+	}
+
+	cli.Log.Debugw("retrieving LQL query", "queryID", queryID)
+
+	queryResponse, err := cli.LwApi.LQL.GetQueryByID(queryID)
+
```
afiune, line 57: Likewise here, lets follow this pattern for any other command that has to check for required arguments:
```suggestion
	lqlShowUnableMsg := "unable to show LQL query"
	cli.Log.Debugw("retrieving LQL query", "queryID", args[0])
	queryResponse, err := cli.LwApi.LQL.GetQueryByID(args[0])
```

cli/cmd/lql_show.go
```diff
@@ -0,0 +1,69 @@
...
+	if err != nil {
+		return errors.Wrap(err, lqlShowUnableMsg)
+	}
+	if cli.JSONOutput() {
+		return cli.OutputJSON(queryResponse.Data)
+	}
+	if len(queryResponse.Data) == 0 {
+		return yikes(lqlShowUnableMsg)
```
afiune, line 65: Loooove this! ♥️

cli/cmd/lql_update.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	// lqlUpdateCmd represents the lql update command
+	lqlUpdateCmd = &cobra.Command{
+		Use:   "update",
+		Short: "update an LQL query",
+		Long:  `Update an LQL query.`,
+		Args:  cobra.NoArgs,
```
afiune, line 34: ?? This doesn't seem right, don't we need at least an ID of the query to update?

Am I missing something 🤔

integration/lql_delete_test.go
```diff
@@ -0,0 +1,64 @@
...
+	assert.Equal(t, 1, exitcode, "EXITCODE is not the expected one")
+}
+
+func TestQueryDelete(t *testing.T) {
+	if os.Getenv("CI_BETA") == "" {
+		t.Skip("skipping test in production mode")
+	}
+
+	// setup
+	LaceworkCLIWithTOMLConfig("query", "create", "-u", lqlQueryURL)
```
afiune, line 55: Are you coming back to assert this command execution?

integration/lql_show_test.go
```diff
@@ -0,0 +1,55 @@
...
+	assert.Empty(t, out.String(), "STDOUT should be empty")
+	assert.Contains(t, err.String(), "ERROR unable to show LQL query: Please specify a valid query ID.")
+	assert.Equal(t, 1, exitcode, "EXITCODE is not the expected one")
+}
+
+func TestQueryShow(t *testing.T) {
+	if os.Getenv("CI_BETA") == "" {
+		t.Skip("skipping test in production mode")
+	}
+	// tested by virtue of TestQueryCreateFile
```
afiune, line 53: could we add a ticket where you will revisit this?

afiune (approved):
Overall the codes look pretty great! There are definitely areas of improvement but we will revisit those as we add, remove, refactor or improve functionality.

Thank you so much for this work, @hazedav !!!

tenor-171381768

api/lql.go
```diff
@@ -18,24 +18,203 @@
...
+	// query
+	if err := q.TranslateQuery(); err != nil {
+		return err
+	}
+	// start
+	if start, err := q.TranslateTime(q.StartTimeRange); err != nil {
+		return err
+	} else {
+		q.StartTimeRange = start
+	}
```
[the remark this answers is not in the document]
hazedav: Makes sense.  I can move the assignment out of the `if`...

api/lql_test.go
```diff
@@ -0,0 +1,636 @@
...
+	"net/http"
+	"testing"
+
+	"github.com/lacework/go-sdk/api"
+	"github.com/lacework/go-sdk/internal/lacework"
+	"github.com/stretchr/testify/assert"
+)
+
+var (
+	lqlQueryID    string = "my_lql"
```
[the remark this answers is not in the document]
hazedav: I will check any variable based assignments and let Go assume type...

cli/cmd/lql.go
```diff
@@ -19,78 +19,241 @@
...
+				&lqlCmdState.File,
+				"file", "f", "",
+				"path to an LQL query to run",
+			)
+			/* repo flag to specify a query from repo
+			cmd.BoolVarP(
+				&lqlCmdState.Repo,
+				"repo", "r", false,
+				"id of an LQL query to run via active repo",
+			)*/
```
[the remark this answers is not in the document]
hazedav: I was leaving it as a placeholder for ALLY-423, but it's not absolutely necessary...

cli/cmd/lql.go
```diff
@@ -19,78 +19,241 @@
...
+		return
 	}
+	query = string(body)
+	return
+}
 
+func inputQueryFromEditor(action string) (query string, err error) {
+	prompt := &survey.Editor{
+		Message:  fmt.Sprintf("Type a query to %s", action),
+		FileName: "query*.sh",
```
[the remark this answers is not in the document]
hazedav: I think `.lql` would be best.  Good eye.

cli/cmd/lql_create.go
```diff
@@ -0,0 +1,69 @@
...
+
+func init() {
+	// add sub-commands to the lql command
+	lqlCmd.AddCommand(lqlCreateCmd)
+
+	setQueryFlags(lqlCreateCmd.Flags())
+}
+
+func createQuery(cmd *cobra.Command, args []string) error {
+	lqlCreateUnableMsg := "unable to create LQL query"
```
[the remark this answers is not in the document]
hazedav: This is used in a wrapped manner, so there should be no confusion as it will be accompanied by a more specific error.  It's a variable simply to save on duplicate strings, but we can easily explode it.

cli/cmd/lql_delete.go
```diff
@@ -0,0 +1,69 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	// lqlDeleteCmd represents the lql delete command
+	lqlDeleteCmd = &cobra.Command{
+		Use:   "delete <query_id>",
+		Short: "delete an LQL query",
+		Long:  `Delete an LQL query.`,
+		Args:  cobra.MaximumNArgs(1),
```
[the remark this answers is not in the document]
hazedav: Agreed. `ExactArgs(1)` is what we want here...

cli/cmd/lql_update.go
hazedav: The update API is slightly wonky for the following reasons:

1. The API will take the LQL_ID from the QUERY_TEXT
2. You can specify an LQL_ID in the JSON POST body
3. You can specify an LQL_ID argument in the query param

- So if you're using just LQL syntax, or a JSON with just QUERY_TEXT then LQL ID will be inferred.
- If you're using JSON you can specify LQL_ID in the JSON
- If you're using JSON and don't specify LQL_ID in the JSON perhaps you want to specify it as a CLI argument?

I think we should have a separate ticket to sort this all out.

integration/lql_delete_test.go
hazedav: I don't think we need to assert here, especially if the query already exists.

integration/lql_list_test.go
```diff
@@ -0,0 +1,58 @@
...
+	assert.Equal(t, 0, exitcode, "EXITCODE is not the expected one")
+}
+
+func TestQueryList(t *testing.T) {
+	if os.Getenv("CI_BETA") == "" {
+		t.Skip("skipping test in production mode")
+	}
+
+	// setup
+	LaceworkCLIWithTOMLConfig("query", "create", "-u", lqlQueryURL)
```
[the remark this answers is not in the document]
hazedav: I don't think we need to assert here, especially if the query already exists.

integration/lql_show_test.go
hazedav: This isn't something that needs revisiting.  I'm just calling out that this test case is satisfied by virtue of the `TestQueryCreateFile` which performs a `show`...

api/lql.go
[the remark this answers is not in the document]
hazedav: Fixed.

api/lql_test.go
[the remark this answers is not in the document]
hazedav: Fixed.

api/lql_test.go
```diff
@@ -0,0 +1,636 @@
...
+)
+
+type LQLTranslateTimeTest struct {
+	Name       string
+	Input      string
+	ReturnTime string
+	ReturnErr  interface{}
+}
+
+var lqlTranslateTimeTests []LQLTranslateTimeTest = []LQLTranslateTimeTest{
```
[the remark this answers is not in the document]
hazedav: Fixed.

cli/cmd/lql.go
```diff
@@ -19,78 +19,241 @@
...
 
-	if data, ok := response["data"]; ok {
-		err := cli.OutputJSON(data)
-		return err
+	query = string(fileData)
+	return
+}
+
+func inputQueryFromURL(url string) (query string, err error) {
+	msg := "unable to open URL"
```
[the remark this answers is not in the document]
hazedav: Fixed.

cli/cmd/lql.go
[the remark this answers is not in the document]
hazedav: Fixed.

cli/cmd/lql_create.go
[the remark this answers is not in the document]
hazedav: Fixed.

cli/cmd/lql_delete.go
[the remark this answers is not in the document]
hazedav: Fixed.

cli/cmd/lql_delete.go
hazedav: Fixed.

cli/cmd/lql_show.go
hazedav: Fixed.

cli/cmd/lql_describe.go
```diff
@@ -0,0 +1,105 @@
...
+	"github.com/lacework/go-sdk/api"
+)
+
+var (
+	// lqlDescribeCmd represents the lql describe command
+	lqlDescribeCmd = &cobra.Command{
+		Use:   "describe <data_source>",
+		Short: "describe an LQL data source",
+		Long:  `Describe an LQL data source.`,
+		Args:  cobra.MaximumNArgs(1),
```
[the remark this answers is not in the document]
hazedav: Fixed.

cli/cmd/lql_show.go
hazedav: Fixed.

cli/cmd/lql_update.go
hazedav: ALLY-442

PR #355, opened 2021-03-29
chore: Validation script for git commit message conventions

dmurray-lacework:
Add githook for commit message validation.
Add script and make cmd for updating the githook

The githook checks if commit message contains one of `feat:\|fix:\|style:\|refactor:\|test:\|docs:\|chore:\|build:\|ci:\|perf:\|metric:\|misc:`

In order for the githook to be used when `git commit` is called, it must be moved to .git/hooks/ directory.
`make git-env` checks if the hook exists or is not latest version before adding it to .git/hooks/ 

Upon success output is 
```
❯ gc -m "chore: Validation script for git commit message conventions"
--> go-sdk: Commit message is valid
[dmurray-lacework/enforce-commit-message 491a091] chore: Validation script for git commit message conventions
...
```

Upon failure output is 
```
❯ gc -m "cho: Add make cmd for git hooks update" 
--> go-sdk: Invalid commit message
--> go-sdk: Message must contain one of feat:|fix:|style:|refactor:|test:|docs:|chore:|build:|ci:|perf:|metric:|misc:
```

### DIFF
--- a/scripts/git_env.sh
+++ b/scripts/git_env.sh
@@ -1,3 +1,4 @@
+
 #!/bin/bash
 #
 # Name::        git_env.sh
@@ -5,7 +6,7 @@
 # Author::      Darren Murray (<dmurray-lacework@lacework.net>)
 #
 
-readonly commit_hook=scripts/githooks/prepare-commit-msg
+readonly commit_hook=scripts/githooks/commit-msg
 readonly hooks_path=scripts/githooks/
 readonly project_name=go-sdk
 
--- a/scripts/git_env.sh
+++ b/scripts/git_env.sh
@@ -1,40 +1,22 @@
 #!/bin/bash
 #
 # Name::        git_env.sh
-# Description:: Use this script to configure local git env,
-#               including commit message validation
+# Description:: Use this script to configure local git env
 # Author::      Darren Murray (<dmurray-lacework@lacework.net>)
 #
 
-readonly commit_hook=.git/hooks/prepare-commit-msg
-readonly new_commit_hook=scripts/githooks/prepare-commit-msg
+readonly commit_hook=scripts/githooks/prepare-commit-msg
+readonly hooks_path=scripts/githooks/
 readonly project_name=go-sdk
 
-create_prepare_commit_msg(){
-  if [ ! -f $commit_hook ]; then
-    log "Adding git commit hooks"
-    cp scripts/githooks/prepare-commit-msg $commit_hook
-    chmod +x $commit_hook
-  else 
-    log "Git commit hook already exists"
-    update
-  fi
-}
-
-update() {
-  currentVersion=$(cat $commit_hook | grep 'readonly version=v'| sed 's/.*readonly version=\(.*\)/\1/')
-  newVersion=$(cat $new_commit_hook | grep 'readonly version=v'| sed 's/.*readonly version=\(.*\)/\1/')
-    if [ $newVersion != $currentVersion ]; then
-        log "Updating git commit hooks version $currentVersion -> $newVersion"
-        cp scripts/githooks/prepare-commit-msg $commit_hook
-        chmod +x $commit_hook
-    else 
-        log "No changes made to version"
-  fi
+prepare_git_env(){
+  chmod +x $commit_hook
+  log "Setting git-hooks path to $hooks_path"
+  git config core.hooksPath $hooks_path
 }
 
 log() {
   echo "--> ${project_name}: $1"
 }
 
-create_prepare_commit_msg
+prepare_git_env
--- a/scripts/git_env.sh
+++ b/scripts/git_env.sh
@@ -1,0 +1,40 @@
+#!/bin/bash
+#
+# Name::        git_env.sh
+# Description:: Use this script to configure local git env,
+#               including commit message validation
+# Author::      Darren Murray (<dmurray-lacework@lacework.net>)
+#
+
+readonly commit_hook=.git/hooks/prepare-commit-msg
+readonly new_commit_hook=scripts/githooks/prepare-commit-msg
+readonly project_name=go-sdk
+
+create_prepare_commit_msg(){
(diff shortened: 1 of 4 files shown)


scripts/githooks/prepare-commit-msg
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+readonly version=v0.0.1
+readonly project_name=go-sdk
+readonly commit_tags=feat:\|fix:\|style:\|refactor:\|test:\|docs:\|chore:\|build:\|ci:\|perf:\|metric:\|misc:
```
dmurray-lacework, line 4: I've added 'misc' to the list of valid commands. Giving us a way to circumvent the validation if we need to for some reason. Undecided on whether to keep it in or not.

afiune (commented):
Before starting to review this, have you looked at https://pre-commit.com/ ?

scripts/git_env.sh
```diff
@@ -0,0 +1,40 @@
...
+#
+
+readonly commit_hook=.git/hooks/prepare-commit-msg
+readonly new_commit_hook=scripts/githooks/prepare-commit-msg
+readonly project_name=go-sdk
+
+create_prepare_commit_msg(){
+  if [ ! -f $commit_hook ]; then
+    log "Adding git commit hooks"
+    cp scripts/githooks/prepare-commit-msg $commit_hook
```
afiune, line 16: Why not doing a link instead?

scripts/git_env.sh
```diff
@@ -0,0 +1,40 @@
...
+    chmod +x $commit_hook
+  else 
+    log "Git commit hook already exists"
+    update
+  fi
+}
+
+update() {
```
afiune, line 24: With a link we wouldn't have to worry about updating. 😏

afiune (commented):
If we don't go for the pre-commit route, I would at least like this to be links so we don't have to worry about the update of the hooks.

scripts/git_env.sh
dmurray-lacework: Cool, good suggestion Salim. I'll do that instead to save the updating trouble.

dmurray-lacework:
> Before starting to review this, have you looked at https://pre-commit.com/ ?

I've only now had a look at pre-commit, I think I'd rather use a native git hook than throw a 3rd party python dependency into the mix. Would using pre-commit mean contributors to the go-sdk have to have python & pip? 
Also use case at the minute is fairly simple I'm not sure we need all the other functionality.

scripts/git_env.sh
```diff
@@ -0,0 +1,22 @@
...
+#
+
+readonly commit_hook=scripts/githooks/prepare-commit-msg
+readonly hooks_path=scripts/githooks/
+readonly project_name=go-sdk
+
+prepare_git_env(){
+  chmod +x $commit_hook
+  log "Setting git-hooks path to $hooks_path"
+  git config core.hooksPath $hooks_path
```
dmurray-lacework, line 15: Changed the script to set the hooks path instead of using an update function. I also thought this might be better than symlinking each individual githook. Now to add a new hook we just need to drop it in scripts/githooks
Let me know your thoughts on this @afiune

afiune: I'm not 100% on board to change the Git configuration of our contributors but on the other hand, this looks way cleaner now. let's just give this a try, worse case if we see problems we can always try the symlinks. 😉

afiune (changes requested):
I tested this PR and everything look ok, we can try it but, we need to allow
git commits with a title similar to `type(scope): subject` or the one you have
already working without scope, that is `type: subject`. We need both.

Look at this [confluence page](https://lacework.atlassian.net/wiki/spaces/ALLY/pages/853934952/Git+Commits+PRs#Mastering-Git-Commit-Messages) for examples.

```
➜  go-sdk git:(dmurray-lacework/enforce-commit-message) make git-env
scripts/git_env.sh
--> go-sdk: Setting git-hooks path to scripts/githooks/
➜  go-sdk git:(dmurray-lacework/enforce-commit-message) touch foo
➜  go-sdk git:(dmurray-lacework/enforce-commit-message) ✗ ga foo
➜  go-sdk git:(dmurray-lacework/enforce-commit-message) ✗ gc -m "test"
--> go-sdk: Invalid commit message
--> go-sdk: Message must contain one of feat:|fix:|style:|refactor:|test:|docs:|chore:|build:|ci:|perf:|metric:|misc:
➜  go-sdk git:(dmurray-lacework/enforce-commit-message) ✗ gc -m "test(cli): test"
--> go-sdk: Invalid commit message
--> go-sdk: Message must contain one of feat:|fix:|style:|refactor:|test:|docs:|chore:|build:|ci:|perf:|metric:|misc:
➜  go-sdk git:(dmurray-lacework/enforce-commit-message) ✗ gc -m "test: test"
--> go-sdk: Commit message is valid
[dmurray-lacework/enforce-commit-message d9f0562] test: test
 1 file changed, 0 insertions(+), 0 deletions(-)
 create mode 100644 foo
➜  go-sdk git:(dmurray-lacework/enforce-commit-message)
```

From that test 👆🏽this command should have worked:
```
git commit -m "test(cli): test"
```

afiune:
Another thing, I just tested using the command `git commit` without any message and it doesn't allow me to proceed. Normally it will open my EDITOR but it is not doing that anymore:
```
$ git commit
--> go-sdk: Invalid commit message
--> go-sdk: Message must contain one of feat:|fix:|style:|refactor:|test:|docs:|chore:|build:|ci:|perf:|metric:|misc:
```

Could we fix that?

afiune:
> Would using pre-commit mean contributors to the go-sdk have to have python & pip?

Nope, there is a pre-commit implementation in Go: https://pre-commit.com/#golang

> Also use case at the minute is fairly simple I'm not sure we need all the other functionality.

Aggreed! Let's move forward with your script.

afiune (approved):
# Let's try it!
tenor-51239171

PR #360, opened 2021-04-01
fix(api): update time format in API response

afiune:
This bug-fix was originated from a change in the response of our API in
the server side.


scottford-lw (approved):
👍🎉

PR #375, opened 2021-04-06
feat: Filter vuln host list cves and show assessment by severity

dmurray-lacework:
Filter vuln host commands by severity:
- list cves 
- show assessment 

Usage :
`lacework vuln host list-cves --severity high`
`lacework vuln host list-cves --active --severity high --fixable `
`lacework vuln host show-assessment 101 --severity high`
`lacework vuln host show-assessment 101 --packages --severity high`

Output:
```
      CVE ID        SEVERITY   SCORE       PACKAGE         CURRENT VERSION         FIX VERSION        OS VERSION    HOSTS   PKG STATUS   VULN STATUS
------------------+----------+-------+-----------------+----------------------+--------------------+--------------+-------+------------+--------------
...

 10 of 100 cve(s) showing
```

### DIFF
--- a/cli/cmd/event.go
+++ b/cli/cmd/event.go
@@ -843,6 +843,8 @@ func severityToProperTypes(severity string) (int, string) {
 		return 4, "Low"
 	case "5", "info":
 		return 5, "Info"
+	case "6", "negligible":
+		return 6, "Negligible"
 	default:
 		return 0, "Unknown"
 	}
--- a/cli/cmd/event.go
+++ b/cli/cmd/event.go
@@ -843,8 +843,6 @@ func severityToProperTypes(severity string) (int, string) {
 		return 4, "Low"
 	case "5", "info":
 		return 5, "Info"
-	case "6", "negligible":
-		return 6, "Negligible"
 	default:
 		return 0, "Unknown"
 	}
--- a/cli/cmd/vuln_host.go
+++ b/cli/cmd/vuln_host.go
@@ -583,14 +583,14 @@ func hostVulnCVEsTable(cves []api.HostVulnCVE) ([][]string, string) {
 	}
 
 	if filteredCves > 0 {
-		showing :=  totalCves - filteredCves
+		showing := totalCves - filteredCves
 		return out, fmt.Sprintf("\n %v of %v cve(s) showing \n", showing, totalCves)
 	}
 
 	return out, ""
 }
 
-func hostVulnCVEsTableForSeverity(cves []api.HostVulnCVE, severity string) ([][]string, int ,int) {
+func hostVulnCVEsTableForSeverity(cves []api.HostVulnCVE, severity string) ([][]string, int, int) {
 	var (
 		filtered = 0
 		total    = 0
@@ -686,7 +686,7 @@ func hostVulnHostDetailsToTable(assessment api.HostVulnHostAssessment) string {
 	)
 
 	mainBldr.WriteString("\n")
-	if vulCmdState.Details || vulCmdState.Fixable || vulCmdState.Packages || vulCmdState.Active {
+	if vulCmdState.Details || vulCmdState.Fixable || vulCmdState.Packages || vulCmdState.Active || vulCmdState.Severity != "" {
 		if vulCmdState.Packages {
 			mainBldr.WriteString(
 				renderSimpleTable(
@@ -695,7 +695,7 @@ func hostVulnHostDetailsToTable(assessment api.HostVulnHostAssessment) string {
 				),
 			)
 		} else {
-			rows := hostVulnCVEsTableForHostView(assessment.CVEs)
+			rows, filtered := hostVulnCVEsTableForHostView(assessment.CVEs)
 
 			// if the user wants to show only vulnerabilities of active packages
 			// and we don't have any, show a friendly message
@@ -716,11 +716,15 @@ func hostVulnHostDetailsToTable(assessment api.HostVulnHostAssessment) string {
 					rows,
 				))
 			}
+
+			if filtered != "" {
+				mainBldr.WriteString(filtered)
+			}
 		}
 		mainBldr.WriteString("\n")
 	}
 
-	if !vulCmdState.Details && !vulCmdState.Active && !vulCmdState.Fixable && !vulCmdState.Packages {
+	if !vulCmdState.Details && !vulCmdState.Active && !vulCmdState.Fixable && !vulCmdState.Packages && vulCmdState.Severity != "" {
 		mainBldr.WriteString(
 			"Try adding '--details' to increase details shown about the vulnerability assessment.\n",
 		)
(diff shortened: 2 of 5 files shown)


cli/cmd/event.go
```diff
@@ -843,6 +843,8 @@ func severityToProperTypes(severity string) (int, string) {
 		return 4, "Low"
 	case "5", "info":
 		return 5, "Info"
+	case "6", "negligible":
```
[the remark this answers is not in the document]
afiune: Actually, we need to replace all the code that references **Negligible** to be now **Info**. Somehow the APIs changed and now we are displaying Info like the container vulnerability scape.....

cli/cmd/vuln_host.go
```diff
@@ -556,30 +569,53 @@ func hostVulnPackagesTable(cves []api.HostVulnCVE, withHosts bool) [][]string {
...
+	}
+
+	return out, ""
 }
 
-func hostVulnCVEsTableForSeverity(cves []api.HostVulnCVE, severity string) [][]string {
+func hostVulnCVEsTableForSeverity(cves []api.HostVulnCVE, severity string) ([][]string, int ,int) {
 	out := [][]string{}
+	var filtered = 0
+	var total = 0
```
afiune, line 595: nit:
```suggestion
	var (
		filtered = 0
		total    = 0
		out      = [][]string{}
	)
```

cli/cmd/vuln_host.go
```diff
@@ -601,7 +637,7 @@ func hostVulnCVEsTableForSeverity(cves []api.HostVulnCVE, severity string) [][]s
 		return stringToInt(out[i][7]) > stringToInt(out[j][7])
 	})
 
-	return out
+	return out, total-filtered, total
```
afiune, line 640: This return is confusing. If we are returning the total vulnerabilities minus the filtered ones, what exactly are we returning? The unfiltered or filtered? 

I ask this since you call this function like this:
```
output, filtered, total := hostVulnCVEsTableForSeverity(cves, sev)
```

cli/cmd/vuln_host_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+func TestListCvesFilterSeverity(t *testing.T) {
+	vulCmdState.Severity = "critical"
+	mockCves := []api.HostVulnCVE{mockCveOne}
+	result, output := hostVulnCVEsTable(mockCves)
+
+	assert.Equal(t, len(result), 1)
+	assert.Equal(t, output, "\n 1 of 2 cve(s) showing \n")
+
+	clearVulnFilters()
```
afiune, line 35: After setting `vulCmdState.Severity` to critical, defer this clear function.

cli/cmd/vuln_host_test.go
```diff
@@ -0,0 +1,97 @@
...
+	VulnerabilityStatus: "",
+	Version:             "",
+	HostCount:           "1",
+	PackageStatus:       "",
+	CveLink:             "",
+	CvssScore:           "",
+	CvssV2Score:         "",
+	CvssV3Score:         "",
+	FixAvailable:        "1",
+	FixedVersion:        "",
```
afiune, line 95: any reason why we set all these fields to an empty string? Isn't that the default value for strings?

cli/cmd/vuln_host_test.go
```diff
@@ -0,0 +1,97 @@
...
+
+package cmd
+
+import (
+	"github.com/lacework/go-sdk/api"
+	"github.com/stretchr/testify/assert"
+	"testing"
+)
+
+func TestListCvesFilterSeverity(t *testing.T) {
```
afiune, line 27: could we add more tests? maybe some integration tests as well?

afiune (changes requested):

cli/cmd/vuln_host_test.go
dmurray-lacework: 👍

cli/cmd/vuln_host_test.go
dmurray-lacework: 👍  Will do!

cli/cmd/vuln_host_test.go
dmurray-lacework: I let the IDE autocomplete all fields and didn't remove them, will remove empty fields.

cli/cmd/vuln_host.go
dmurray-lacework: Yeh, I had this thought too, it's not returning the number of filtered items. I'll rethink this.

afiune (changes requested):
When developing these new flags we need to make sure we give proper messages to our users if not, they will be confused.

For example, if I run these command with the severity flag set to critical:
```
$ lacework vuln host list-cves --severity critical
There are no vulnerabilities in your environment.
```
That message is misleading 👆🏽 because there are indeed vulnerabilities, but there are NO vulnerabilities with a critical severity, the message should be instead:
 
```
There are no critical vulnerabilities running in your environment.
```

This is the pattern we followed for the other flags `--active` and `--fixable`,
here an example:
```
$ lacework vuln host list-cves --active --fixable
There are no fixable vulnerabilities of packages actively running in your environment.
```

afiune:
This makes me think that the other `--severity` flags we added to the container vulnerability space are not following this same standard ... 🤔 It would be nice to double check.

cli/cmd/vuln_host.go
```diff
@@ -493,93 +510,141 @@ func hostVulnSummaryFromHostDetail(hostVulnSummary *api.HostVulnCveSummary) (str
...
 	}
 
 	// order by severity
 	sort.Slice(out, func(i, j int) bool {
 		return severityOrder(out[i][1]) < severityOrder(out[j][1])
 	})
 
-	return out
+	if len(filteredPackages) > 0 {
+		filteredOutput := fmt.Sprintf("%v of %v package(s) showing \n", len(out), len(aggregatedPackages)+len(filteredPackages))
```
afiune, line 591: nit: Everytime I read this footer I feel it looks way better if we type it like:
```
showing X of Y package(s)
```
But maybe this is just me..... 😓 
tenor-83302290

dmurray-lacework: I quite like this way, with the most important details first. But open to be persuaded otherwise.

`showing 1 of 100`
vs
`1 of 100 showing`

dmurray-lacework:
> This makes me think that the other `--severity` flags we added to the container vulnerability space are not following this same standard ... 🤔 It would be nice to double check.

@afiune Slightly differenct behaviour in vuln ctr code path as we check the result of the vuln assessment when displaying the 0 vulns output instead checking the length of the output rows. I've putting in a fix for the current vuln host scenario. I will address the inconsistency in the 2 code paths in our refactor ticket.

cli/cmd/vuln_host.go
afiune: all good, this is just my bad English 😂

tenor-168867785

cli/cmd/vuln_host.go
```diff
@@ -493,93 +510,141 @@ func hostVulnSummaryFromHostDetail(hostVulnSummary *api.HostVulnCveSummary) (str
...
+		_, sev := severityToProperTypes(severity)
+		output, filtered, total := hostVulnCVEsTableForSeverity(cves, sev)
+		filteredCves += filtered
+		totalCves += total
+		out = append(out, output...)
+	}
+
+	if filteredCves > 0 {
+		showing := totalCves - filteredCves
+		return out, fmt.Sprintf("\n %v of %v cve(s) showing \n", showing, totalCves)
```
afiune, line 613: I think this should not have a space.... 🤔 maybe...
```suggestion
		return out, fmt.Sprintf("\n%v of %v cve(s) showing \n", showing, totalCves)
```

cli/cmd/vuln_host.go
```diff
@@ -729,7 +811,12 @@ func hostVulnCVEsTableForHostView(cves []api.HostVulnCVE) [][]string {
 		return severityOrder(out[i][1]) < severityOrder(out[j][1])
 	})
 
-	return out
+	if len(out) < total {
+		showing := total - len(out)
+		return out, fmt.Sprintf("\n %v of %v cve(s) showing \n", showing, total)
```
afiune, line 816: same here, I think
```suggestion
		return out, fmt.Sprintf("\n%d of %vd cve(s) showing \n", showing, total)
```

cli/cmd/vuln_host.go
afiune, line 591:
```suggestion
		filteredOutput := fmt.Sprintf("%d of %d package(s) showing \n", len(out), len(aggregatedPackages)+len(filteredPackages))
```

afiune (approved):
tenor-94268517

PR #377, opened 2021-04-07
feat(cli): split list of Google project id and alias

afiune:
We are implementing a workaround into the Lacework CLI to split the ID
and alias from the list of Google projects when a user runs the command:
```
lacework compliance google list-projects <org_id>
```

Ultimately we will do this in the Lacework backend, but for now, we are
going to adopt this workaround until we start using APIv2.

Example of the new JSON and human-readable outputs:
```
➜ lacework comp gcp list 123456789012 -p mini
       PROJECT ID        PROJECT ALIAS
-----------------------+----------------
  abc-demo-project-123   demo-project
➜ lacework comp gcp list 123456789012 -p mini --json
{
  "organization": {
    "alias": "",
    "id": "123456789012"
  },
  "projects": [
    {
      "alias": "demo-project",
      "id": "abc-demo-project-123"
    }
  ]
}
```

Closes ALLY-431


cli/cmd/compliance_gcp.go
```diff
@@ -242,3 +249,57 @@ func complianceGcpReportDetailsTable(report *api.ComplianceGcpReport) [][]string
 		[]string{"Report Time", report.ReportTime.UTC().Format(time.RFC3339)},
 	}
 }
+
+// ALLY-431 Workaround to split the Project ID and Project Alias
+// ultimately, we need to fix this in the API response
+func fixGcpProjectsApiResponse(gcpInfo api.CompGcpProjects) cliComplianceGcpInfo {
```
dmurray-lacework, line 255: Nitpick: Can we change the function name to be more descriptive. Not a fan of using the word "fix" here.

dmurray-lacework (approved):

cli/cmd/compliance_gcp.go
afiune: Any suggestion?

dmurray-lacework: Nothing great springs to mind. Think I'm ok with fix, I can't think of a better alternative. The argument I was considering was whether we're "fixing" something or not. The api response is what is intended for it's original purpose, even if it's a strange behaviour.

northwestcoder: I would call it what it does: "**split**GcpProjectsApiResponse"

afiune: Done! Thank you for your review and suggestion.

PR #378, opened 2021-04-08
chore: run goimports-reviser to order imports

afiune:
This is the first time I am using `goimports-reviser` to order our imports.
As a best practice, we must have a clean and understandable order of imports
and this tool might help us to do so.

https://github.com/incu6us/goimports-reviser

Commands I ran:
```
$ go get github.com/incu6us/goimports-reviser
$ find $(go list -f {{.Dir}} ./...) -name \*.go -type f -exec goimports-reviser -file-path {} -project-name github.com/lacework/go-sdk \;
```
## TODO
Discuss if this will be something we would like to adopt.

### DIFF
--- a/api/events_test.go
+++ b/api/events_test.go
@@ -24,9 +24,10 @@ import (
 	"testing"
 	"time"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/lacework/go-sdk/api"
 	"github.com/lacework/go-sdk/internal/lacework"
-	"github.com/stretchr/testify/assert"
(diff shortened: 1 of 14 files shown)


dmurray-lacework:
@afiune If I'm reading correctly, it also runs a gofmt perhaps we should replace our current `make fmt `command?

dmurray-lacework (approved):
Approved this PR, but I'd like to see this replace our make fmt command. Provided that it does the same job with the addition of ordering imports

PR #387, opened 2021-04-14
feat: Relative/Natural Time Package For Query Command

hazedav:
This pull request introduces two main time concepts into the Lacework CLI.

TODO:
- [x] Verify integration testing
- [x] Add package testing
- [x] Verify doc string usage

### Concept 1: Relative Times
Relative times allow a user to represent time values dynamically by using specifiers to communicate how to offset from the current time.  For instance, a relative time of `-24h` would produce a date/time that was 24 hours less the current time.  Relative times can also snap to a particular time.  For instance, a relative time of `@d` would represent the start of the current day.

A relative time has 3 components:
1. A signed (+/-) integer.
2. A relative time unit.
3. A relative time snap.

We support the following relative time units (disambiguating month and minute by design):
- y - year
- mon - month
- w - week
- d - day
- h - hour
- m - minute
- s - second

A few more things to consider:
- To represent the current time, one can either specify `now` or `+0s`.
- When specifying an integer and relative time unit, snaps are optional. 
- When specifying a snap, the integer and relative time unit are optional.  For instance, `@d` is actually interpreted as `+0s@d`.

Usage:
`lacework query run --start @m --end now`

### Concept 2: Natural Time Ranges
Natural time ranges allow a user to represent time range values using natural language.  For instance, a natural time range of `yesterday` would represent a relative start time of `-1d@d` and a relative end time of `@d`.

A natural time has 3 components:
1. An adjective.
2. A positive number (only when using the `last` adjective).
3. The full text representation of a relative time unit (i.e. year/years).

We support the following adjectives (disambiguating previous and last by design):
- this/current
- previous
- last

A few more things to consider:
- `last` implies "in the last".  So `last week` reads as "in the last week" and represents a start time of `-1w` and an end time of `now`.
- `previous` will always snap.  So `previous week` represents a start time of `-1w@w` and an end time of `@w`.
- `yesterday` is a valid natural time and is equivalent to `previous day`.
- `today` is a valid natural time and is equivalent to `this day` or `current day`.

Usage:
`lacework query run --range "this month"`


dmurray-lacework (approved):

lwtime/nattime.go
```diff
@@ -0,0 +1,196 @@
...
+		ntr.End = "now"
+	default:
+		// This would indicate a code mismatch between NatTime.Parse() and NatTimeRange.Parse()
+		return errors.New(
+			fmt.Sprintf("invalid adjective for natural time (%s)", s))
+	}
+	return nil
+}
+
+func (ntr NatTimeRange) Range() (start time.Time, end time.Time, err error) {
```
afiune, line 161: What is this function for? Why is not being tested?

lwtime/reltime_test.go
```diff
@@ -0,0 +1,415 @@
...
+			if err != nil {
+				assert.FailNow(t, "unable to parse reference time")
+			}
+			rt := lwtime.RelTime{}
+			err = rt.Parse(rt_t.Input)
+			// if we're expecting an error don't worry about this
+			if rt_t.Error == nil {
+				assert.Nil(t, err)
+			}
+			outTime, err := rt.Time_(refTime)
```
afiune, line 405: I don't think this is what you really want to do here. I feel that you are trying to define a "Constructor" so that you can transform any given time to a "Relative Time", something like this:
```suggestion
			relTime, err := Relative(refTime)
```

lwtime/reltime.go
```diff
@@ -0,0 +1,211 @@
...
+	if !rt.Snap.IsValid() {
+		return errors.New(fmt.Sprintf("invalid snap for relative time specifier (%s)", s))
+	}
+	return nil
+}
+
+func (rt RelTime) Time() (outTime time.Time, err error) {
+	outTime, err = rt.Time_(time.Now())
+	return
+}
```
afiune, line 174: unnecessary naked return
```suggestion
func (rt RelTime) Time() (time.Time, error) {
	return rt.Time_(time.Now())
}
```

afiune: but what is this function for?

lwtime/reltime.go
```diff
@@ -0,0 +1,211 @@
...
+
+type RelTime struct {
+	Num  string
+	INum int
+	Unit RelTimeUnit
+	Snap RelTimeUnit
+}
+
+func (rt *RelTime) Parse(s string) error {
```
afiune, line 126: Why would you want to have a Relative time and after that, Parse another time?

I feel that you need a constructor here again. Let's sync up over Zoom 📹. ⭐

lwtime/nattime.go
hazedav: Testing is being done extensively/exclusively on `Range_()`.  This func has no logic other than to call `Range_()` with the current time.  I can add a simple test for this...

lwtime/reltime.go
hazedav: Similar to `Range_` this function (`Time()`) calls the `Time_()` function with the current time.

lwtime/reltime.go
hazedav: Fixed.

lwtime/reltime.go
hazedav: Fixed.

lwtime/reltime_test.go
hazedav: Fixed.

cli/cmd/lql.go
```diff
@@ -269,6 +279,18 @@ func runQuery(cmd *cobra.Command, args []string) error {
...
+	if lqlCmdState.Range != "" {
+		cli.Log.Debugw("retrieving natural time range")
+
+		var start, end time.Time
+		start, end, err = lwtime.ParseNatural(lqlCmdState.Range)
+		if err != nil {
+			return errors.Wrap(err, msg)
+		}
+		lqlCmdState.Start = start.UTC().Format(time.RFC3339)
+		lqlCmdState.End = end.UTC().Format(time.RFC3339)
```
afiune, line 291: why do we need to do this conversion and formatting here? 🤔 Why not the start and end times get back from the new `lwtime` package ready to be used?

lwtime/nattime.go
```diff
@@ -0,0 +1,198 @@
...
+	nt := natural{}
+	s = strings.ToLower(s)
+
+	// Today
+	if naturalAdjective(s) == Today {
+		nt.adjective = This
+		nt.num = "1"
+		nt.iNum = 1
+		nt.unit = relativeUnit("d")
+		return nt, nil
```
afiune, line 94: nit: instead of reusing the `nt` above, why not just return the object this way?
```suggestion
		return natural{
			adjective: This,
			num: "1", iNum: 1,
			unit: relativeUnit("d"),
		}, nil
```

lwtime/nattime.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+	// Unit
+	if ok := nt.loadRelativeUnit(nt_parts[3]); !ok {
+		// this would indicate a code mismatch between relative units and regex
+		return nt, errors.New(fmt.Sprintf("invalid unit for natural time (%s)", s))
+	}
+	return nt, nil
+}
+
+func (nt natural) range_(t time.Time) (start time.Time, end time.Time, err error) {
```
afiune, line 136: can we call this function `range` only? the underscore `_` is used as a black identifier to denote that "things are not being used"
```suggestion
func (nt natural) range(t time.Time) (start time.Time, end time.Time, err error) {
```

lwtime/nattime.go
```diff
@@ -0,0 +1,198 @@
...
+	return parseNaturalFromTime(n, time.Now())
+}
+
+func parseNaturalFromTime(n string, fromTime time.Time) (time.Time, time.Time, error) {
+	var start, end time.Time
+
+	natural, err := newNatural(n)
+	if err != nil {
+		return start, end, err
+	}
```
afiune, line 195: nit: no need to instantiate those start and end variables
```suggestion
	natural, err := newNatural(n)
	if err != nil {
		return time.Time{}, time.Time{}, err
	}
```

lwtime/reltime.go
```diff
@@ -0,0 +1,237 @@
...
+	return parseRelativeFromTime(s, time.Now())
+}
+
+func parseRelativeFromTime(s string, fromTime time.Time) (time.Time, error) {
+	var t time.Time
+
+	relative, err := newRelative(s)
+	if err != nil {
+		return t, err
+	}
```
afiune, line 234: same here
```suggestion
	relative, err := newRelative(s)
	if err != nil {
		return time.Time{}, err
	}
```

afiune (approved):
Outstanding work @hazedav !!

tenor-122014860

cli/cmd/lql.go
hazedav: This is very astute of you.  I believe `lwtime.ParseNatural` is doing the right thing here by returning go `time.Time` for start/end.  However, the API for `run` is expecting  a string (hence the conversion)...

lwtime/nattime.go
hazedav: I think `range` is a built-in so Go was complaining about this.  I can just find a different name for it since it's internal anyway...

lwtime/nattime.go
hazedav: Fixed.

lwtime/nattime.go
hazedav: Fixed.

lwtime/nattime.go
hazedav: Fixed.

lwtime/reltime.go
hazedav: Fixed.

PR #388, opened 2021-04-14
fix(cli): Fix filters with json output for vuln host command

dmurray-lacework:
Fix json output to should work in combination with any filtering flags

vuln host scan-pkg-manifest vuln_host.go:165 
vuln host list-cves vuln_host.go:207
vuln host show-assessment  vuln_host.go:342


cli/cmd/vuln_host.go
```diff
@@ -211,67 +202,14 @@ with fixes:
 			}
 
 			response, err := cli.LwApi.Vulnerabilities.Host.ListCves()
+
```
afiune, line 205: nit: better not have empty lines between the err and the `if` check.
```suggestion
```

cli/cmd/vuln_host.go
```diff
@@ -595,18 +532,14 @@ func hostVulnPackagesTable(cves []api.HostVulnCVE, withHosts bool) ([][]string,
 	return out, ""
 }
 
-func hostVulnCVEsTable(cves []api.HostVulnCVE) ([][]string, string) {
-	var out [][]string
+func hostVulnCVEs(cves []api.HostVulnCVE) ([]api.HostVulnCVE, string) {
```
afiune, line 535: if this function is filtering please add that to the name of the function just like you did for `filterHostVulnCVEs()`

cli/cmd/vuln_host.go
```diff
@@ -616,60 +549,73 @@ func hostVulnCVEsTable(cves []api.HostVulnCVE) ([][]string, string) {
...
+				pkg.FixedVersion,
+				pkg.Namespace,
+				pkg.HostCount,
+				pkg.PackageStatus,
+				pkg.Status,
+			})
+		}
+	}
+
+	//order by total number of host
```
afiune, line 572: nit: comments should always have space in between
```suggestion
	// order by the total number of host
```

cli/cmd/vuln_host.go
```diff
@@ -949,13 +875,23 @@ func hostScanPackagesVulnDetailsTable(vulns []api.HostScanPackageVulnDetails) []
 			fixedVersion = vuln.FixInfo.FixedVersion
 		}
 
+		vuln.FixInfo.FixedVersion = fixedVersion
```
afiune, line 878: ugh, I might need to understand why are we changing the API response here 🤔 , I don't think we should do that, could we sync up?

cli/cmd/vuln_host.go
```diff
@@ -967,47 +903,39 @@ func hostScanPackagesVulnDetailsTable(vulns []api.HostScanPackageVulnDetails) []
...
-		}
-
-		added := false
-		for i := range out {
-			if out[i][1] == vuln.Severity &&
-				out[i][2] == vuln.OsPkgInfo.Pkg &&
-				out[i][3] == vuln.OsPkgInfo.PkgVer {
+func filterHostScanPackagesVulnPackages(vulns []api.HostScanPackageVulnDetails) filteredPackageTable {
+	var filteredPackages []packageTable
+	var aggregatedPackages []packageTable
```
afiune, line 908: nit
```suggestion
	var (
		filteredPackages   []packageTable
		aggregatedPackages []packageTable
	)
```

cli/cmd/vuln_host.go
```diff
@@ -967,47 +903,39 @@ func hostScanPackagesVulnDetailsTable(vulns []api.HostScanPackageVulnDetails) []
...
 
-		fixes := "0"
-		if vuln.HasFix() {
-			fixes = "1"
-		}
+		aggregatedPackages = aggregatePackagesWithHosts(aggregatedPackages, pack, false, vuln.HasFix())
+	}
+
+	totalUnfiltered := len(filteredPackages) + len(aggregatedPackages)
+	return filteredPackageTable{packages: aggregatedPackages, totalPackages: len(aggregatedPackages), totalUnfiltered: totalUnfiltered}
```
afiune, line 927: nit: style and no need to instantiate `totalUnfiltered` variable.
```suggestion
	return filteredPackageTable{
		packages:        aggregatedPackages,
		totalPackages:   len(aggregatedPackages),
		totalUnfiltered: len(filteredPackages) + len(aggregatedPackages),
	}
```

cli/cmd/vuln_host.go
```diff
@@ -1018,3 +946,119 @@ func hostScanPackagesVulnPackagesView(vulns []api.HostScanPackageVulnDetails) []
...
+		)
+	}
+	return nil
+}
+
+// Build the cli output for vuln host scan-package-manifest
+func buildVulnHostScanPkgManifestReports(response api.HostVulnScanPkgManifestResponse) error {
+	if len(response.Vulns) == 0 {
+		// @afiune add a helpful message, possible things are:
+		cli.OutputHuman("There are no vulnerabilities found.\n")
```
afiune, line 1049: I wanted to change this for months, let us do it now that I remembered and we can:
```suggestion
		cli.OutputHuman(fmt.Sprintf("There are no vulnerabilities found! Time for %s\n", randomEmoji()))
```

afiune (changes requested):

cli/cmd/vuln_host.go
dmurray-lacework: Good catch Salim, I think you're right. In fact I think 872-877 is redundant and can just be removed.

afiune (approved):
tenor-110317699

afiune (approved):
tenor-70719112

PR #389, opened 2021-04-16
fix(cli): Hard coding polling mechanism to start after 64 seconds

scottford-lw:
This PR adds 64 seconds to the start time for the polling mechanism on container vulnerability scans. This fix is to mitigate issues in the field where long scan times cause API rate limits to be exceeded.

Signed-off-by: Scott Ford <scott.ford@lacework.net>


afiune (approved):
YOLO!

PR #391, opened 2021-04-19
test(policy): adding policy test resource

hazedav:
Adding policy resource for -u URL testing

ALLY-451


afiune (approved):
tenor-267548868

PR #392, opened 2021-04-20
feat(cli): lacework policy create

hazedav:
Introducing policy create command

ALLY-445

### DIFF
--- a/api/policy.go
+++ b/api/policy.go
@@ -36,7 +36,7 @@ type Policy struct {
 	PolicyID string `json:"policy_id"`
 }
 
-func (svc *PolicyService) CreatePolicy(policy string) (
+func (svc *PolicyService) Create(policy string) (
 	response PolicyCreateResponse,
 	err error,
 ) {
--- a/api/policy.go
+++ b/api/policy.go
@@ -1,0 +1,49 @@
+//
+// Author:: Salim Afiune Maya (<afiune@lacework.net>)
+// Copyright:: Copyright 2020, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package api
+
+import "encoding/json"
+
+// PolicyService is a service that interacts with the Custom Policies
+// endpoints from the Lacework Server
+type PolicyService struct {
+	client *Client
+}
+
+type PolicyCreateResponse struct {
+	Data    []Policy `json:"data"`
+	Ok      bool     `json:"ok"`
+	Message string   `json:"message"`
+}
+
+type Policy struct {
+	PolicyID string `json:"policy_id"`
+}
+
+func (svc *PolicyService) CreatePolicy(policy string) (
+	response PolicyCreateResponse,
+	err error,
+) {
+	var p map[string]interface{}
+	if err = json.Unmarshal([]byte(policy), &p); err != nil {
+		return
+	}
+	err = svc.client.RequestEncoderDecoder("POST", ApiPolicy, p, &response)
+	return
+}
--- a/cli/cmd/policy.go
+++ b/cli/cmd/policy.go
@@ -25,6 +25,7 @@ import (
 	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/lacework/go-sdk/api"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
@@ -108,7 +109,7 @@ func setPolicySourceFlags(cmds ...*cobra.Command) {
 }
 
 // for commands that take a policy as input
-func inputPolicy(cmd *cobra.Command, args []string) (string, error) {
+func inputPolicy(cmd *cobra.Command) (string, error) {
 	// if running via repo
 	if policyCmdState.Repo {
 		return inputPolicyFromRepo()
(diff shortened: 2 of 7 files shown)


cli/cmd/policy.go
```diff
@@ -0,0 +1,202 @@
...
+var (
+	policyCmdState = struct {
+		File string
+		Repo bool
+		URL  string
+	}{}
+
+	// policyCmd represents the policy parent command
+	policyCmd = &cobra.Command{
+		Hidden: false,
```
hazedav, line 41: This will be put to `Hidden: true` prior to merging the feature branch

api/policy.go
```diff
@@ -0,0 +1,49 @@
...
+	Data    []Policy `json:"data"`
+	Ok      bool     `json:"ok"`
+	Message string   `json:"message"`
+}
+
+type Policy struct {
+	PolicyID string `json:"policy_id"`
+}
+
+func (svc *PolicyService) CreatePolicy(policy string) (
```
afiune, line 39: this looks redundant to me, if you create a new service called Policy and then a function called CreatePolicy you would call it like `client.Policy. CreatePolicy()` shouldn't be better to call this function just `Create()`?

I am assuming that Policy is in fact a separate thing from LQL, if not, why not put it inside the LQL service? That way you could do something like `client.LQL.CreatePolicy()` which is explicit and not redundant.

What do you think?

cli/cmd/policy.go
```diff
@@ -0,0 +1,202 @@
...
+{
+    "policy_id": "lacework-example-1",
+    "title": "My Policy",
+    "enabled": false,
+    "lql_id": "MyQuery",
+    "severity": "high",
+    "description": "My Policy Description",
+    "remediation": "My Policy Remediation"
+}`,
+		Args: cobra.NoArgs,
```
afiune, line 66: uhm, you are not passing any arguments but down in the functions you seem to be using the variables.... 🤔

cli/cmd/policy.go
```diff
@@ -0,0 +1,202 @@
...
+		cmd.Flags().StringVarP(
+			&policyCmdState.URL,
+			"url", "u", "",
+			fmt.Sprintf("url to a policy to %s", action),
+		)
+	}
+}
+
+// for commands that take a policy as input
+func inputPolicy(cmd *cobra.Command, args []string) (string, error) {
```
afiune, line 111: if you really don't need the arguments, don't pass them:
```suggestion
func inputPolicy(cmd *cobra.Command) (string, error) {
```

cli/cmd/policy.go
```diff
@@ -0,0 +1,202 @@
...
+		Message:  fmt.Sprintf("Type a policy to %s", action),
+		FileName: "policy*.json",
+	}
+	err = survey.AskOne(prompt, &policy)
+
+	return
+}
+
+func createPolicy(cmd *cobra.Command, args []string) error {
+	policy, err := inputPolicy(cmd, args)
```
afiune, line 181: if you really don't need the arguments, don't use them:

```suggestion
func createPolicy(cmd *cobra.Command, _ []string) error {
	policy, err := inputPolicy(cmd)
```

cli/cmd/policy.go
```diff
@@ -0,0 +1,202 @@
...
+	fileData, err := ioutil.ReadFile(filePath)
+
+	if err != nil {
+		err = errors.Wrap(err, "unable to read file")
+		return
+	}
+
+	policy = string(fileData)
+	return
+}
```
afiune, line 144: Named returns here doesn't help here, it actually makes it more confusing.
```suggestion
func inputPolicyFromFile(filePath string) (string, error) {
	fileData, err := ioutil.ReadFile(filePath)
	if err != nil {
		return "", errors.Wrap(err, "unable to read file")
	}

	return string(fileData), nil
}
```

cli/cmd/policy.go
```diff
@@ -0,0 +1,202 @@
...
+func createPolicy(cmd *cobra.Command, args []string) error {
+	policy, err := inputPolicy(cmd, args)
+	if err != nil {
+		return errors.Wrap(err, "unable to create policy")
+	}
+
+	cli.Log.Debugw("creating policy", "policy", policy)
+
+	create, err := cli.LwApi.Policy.CreatePolicy(policy)
+
```
afiune, line 189: nit: try not to separate the `err := foo` assignation with the `if err != nil` check:
```suggestion
```

cli/cmd/policy.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return errors.Wrap(err, "unable to create policy")
+	}
+	if cli.JSONOutput() {
```
afiune, line 193: This is a good point to use a new line if you want, mainly since you are now doing a different process by checking for JSON output or Human-readable:
```suggestion

	if cli.JSONOutput() {
```

afiune (commented):
Sorry to mention this again but try not to default to use "Named Returns",
they could be confusing sometimes, especially if you are not checking for
one single error but multiple ones.

api/policy.go
hazedav: @afiune .  You make a great point here.  I think having distinct services provides the most flexibility.  In turn we can and should simplify the function names.  I think this pattern should also be retrofitted to the `LQLService` such that we have things like `Run` vs. `RunQuery`

cli/cmd/policy.go
hazedav: This is most likely copy/pasta issue from `inputQuery` which does use `args`.  Will fix.

cli/cmd/policy.go
hazedav: This is most likely copy/pasta issue from `inputQuery` which does use `args`.  Will fix.

cli/cmd/policy.go
hazedav: This is most likely copy/pasta issue from `inputQuery` which does use `args`.  Will fix.

api/policy.go
hazedav: Fixed.

cli/cmd/policy.go
hazedav: Fixed.

cli/cmd/policy.go
hazedav: Fixed.

cli/cmd/policy.go
hazedav: Fixed.

cli/cmd/policy.go
hazedav: Fixed.

cli/cmd/policy.go
hazedav: Fixed.

cli/cmd/policy.go
hazedav: Fixed.

PR #393, opened 2021-04-20
refactor: Remove severity Negligible and replace with Info

dmurray-lacework:
Remove severity Negligible and replace with Info


afiune (approved):
tenor-57755930

PR #394, opened 2021-04-22
feat(cli): lacework query crumbs

hazedav:
1.  Adding verbosity to the output when query text is invalid
2.  Made query flag output dynamic based on command

ALLY-432


afiune (approved):
tenor-194479193

PR #397, opened 2021-05-06
style: Add gotestsum to Makefile

rmoles:
Adding gotestsum would be a nice addition. It colorizes the test output and gives a nice summary at the end of failed tests

image
image

Signed-off-by: Ross Moles <ross.moles@lacework.net>


Makefile
```diff
@@ -24,8 +24,8 @@ export GOFLAGS GO_LDFLAGS CGO_ENABLED
 
 prepare: install-tools go-vendor
 
-test:
-	go test -v -cover -coverprofile=$(COVERAGEOUT) $(shell go list ./... | grep -v integration)
+test: prepare
```
rmoles, line 27: make install-tools could be the pre-req here or that could even be added as a build step. I just added `prepare` for now to get the build passing

afiune (approved):
tenor-139520001

PR #409, opened 2021-05-12
fix(api): graceful degradation for HTTP 204

hazedav:
If the Lacework API returns a 204, we get an EOF error when attempting to JSON decode.  This causes the UI to present an error even though the request was successful.

ALLY-479


afiune (commented):
This is nice! Could you add a quick test? Then we will be ready to merge. 🙌🏽  🎉

hazedav:
> This is nice! Could you add a quick test? Then we will be ready to merge. 🙌🏽 🎉

Absolutely.

dmurray-lacework (approved):

PR #411, opened 2021-05-13
feat: introducing first APIv2 endpoint UserProfile

afiune:
# User Story

As a Lacework engineer using the Go SDK,
I would like to have a path to implement endpoints from the new APIv2,
So I can start leveraging them and I can stop using APIv1 endpoints. 

# Description

This pull request is implementing one APIv2 endpoint, which includes a
defined architecture for implementing new APIv2 endpoints as well as a
new Authentication method.
 
This new implementation has to consider the fact that we will continue
to use APIv1 endpoints until we migrate them all to the new ones in APIv2.

# Acceptance Criteria

- [x] We will be able to authenticate to both, APIv2 and APIv1 at the same time.
- [x] We have implemented one APIv2 endpoint `/v2/UserProfile`
- [x] Clear documentation about how to use new APIv2 architecture

**JIRA: ALLY-375**

## EXTRA: Enable `lacework api` command to work for both API v1 & v2

Users will be able to run the following commands to access APIv1:
```
$ lacework api get /external/{endpoint}
$ lacework api get /v1/external/{endpoint}
$ lacework api get /api/v1/external/{endpoint}
```
And the following commands to access APIv2:
```
$ lacework api get /v2/{endpoint}
$ lacework api get /api/v2/{endpoint}
```
A few examples that work with this change:
```
$ lacework api get /api/v1/integrations
$ lacework api get /external/integrations
$ lacework api get /api/v2/schemas
$ lacework api get /api/v2/AlertRules
$ lacework api get /api/v2/ContractInfo

```

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -51,7 +51,8 @@ type Client struct {
 	Compliance      *ComplianceService
 	Integrations    *IntegrationsService
 	Vulnerabilities *VulnerabilitiesService
-	V2              *V2Endpoints
+
+	V2 *V2Endpoints
 }
 
 type Option interface {
@@ -102,7 +103,7 @@ func NewClient(account string, opts ...Option) (*Client, error) {
 	c.Compliance = &ComplianceService{c}
 	c.Integrations = &IntegrationsService{c}
 	c.Vulnerabilities = NewVulnerabilityService(c)
-	c.V2 = &V2Endpoints{c}
+	c.V2 = NewV2Endpoints(c)
 
 	// init logger, this could change if a user calls api.WithLogLevel()
 	c.initLogger()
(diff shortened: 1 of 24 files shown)


api/client.go
```diff
@@ -50,6 +51,7 @@ type Client struct {
 	Compliance      *ComplianceService
 	Integrations    *IntegrationsService
 	Vulnerabilities *VulnerabilitiesService
+	V2              *V2Endpoints
```
dmurray-lacework, line 54: Are all v2 endpoints going to live in the V2Endpoints service?

afiune: @dmurray-lacework do you see a different pattern? They are all based on Schemas and there are not sections or grouping. 🤔 

They are all `/v2/{endpoint}` vs V1 that are `/v1/external/suppressions/{endpoint}` and `/v1/external/vulnerability/{endpoint}`

Let us sync up today about this since it is a very important design pattern.

dmurray-lacework (approved):

PR #415, opened 2021-05-17
feat(cli): lacework query fail_on_count

hazedav:
As a Lacework user I would like the "lacework query" CLI command to return a non-zero exit code when the number of results returned by my query meets a certain criteria.

I believe the cleanest way to implement this functionality would be to implement a "--fail_on_count" flag which can take on values such as ">0", "=1", "<10", etc.  This allows us to evaluate the number of results returned by the query without having to alter the semantics of the query in any way.

ALLY-485


cli/cmd/lql.go
```diff
@@ -304,11 +327,59 @@ func runQuery(cmd *cobra.Command, args []string) error {
...
+	re := regexp.MustCompile(operationRE)
+
+	s = strings.TrimSpace(s)
+
+	var op_parts []string
+	if op_parts = re.FindStringSubmatch(s); s == "" || op_parts == nil {
+		return errors.New(
+			fmt.Sprintf("count operation (%s) is invalid", s))
 	}
+	co.num, _ = strconv.Atoi(op_parts[2])
```
afiune, line 364: always check errors 🙏🏽

cli/cmd/lql.go
```diff
@@ -304,11 +327,59 @@ func runQuery(cmd *cobra.Command, args []string) error {
...
+		return err
+	}
+	// fail_on_count post
+	if lqlCmdState.FailOnCount != "" {
+		isFail, err := co.isFail(len(response.Data))
+		if err != nil {
+			return err
+		}
+		if isFail {
+			os.Exit(9)
```
afiune, line 341: This is a no-no in regards to design, the CLI should never have random exits, there are other features that will get impacted by this so the recommendation is to create policy enforcement error like https://github.com/lacework/go-sdk/blob/main/cli/cmd/errors.go#L28-L63

afiune (changes requested):
It took a lot of Customer interviews and UX sessions to figure out that a fail flag would help, have you reached out to anyone to verify that this is indeed something we want to add to the query commands? If so, go for it! 🙌🏽

One major blocker though, the `os.Exit()`.

cli/cmd/lql.go
hazedav: I tend to agree with you.  But it seems unnecessary to check that `\d+` regex can be converted from string to int; no?

hazedav:
> It took a lot of Customer interviews and UX sessions to figure out that a fail flag would help, have you reached out to anyone to verify that this is indeed something we want to add to the query commands? If so, go for it! 🙌🏽

Would the previous interview and UX sessions not just apply here?   I personally haven't reached out to anyone, but we're welcome to stash this until we can validate that customers will adopt such a thing...

cli/cmd/lql.go
hazedav: Fair point regarding random exits.  I saw all this stuff you pointed to, but it seemed like a lot of complexity for what amounts to just wanting a different exit code...  Is there something minimalist here that needs to be satisfied...?

cli/cmd/lql.go
afiune: Super valid point!  I withdraw my comment 🤝

cli/cmd/lql.go
hazedav: @afiune , I got this PR back into shape based on the evolution of the query commands over the last 9 months.  However, we still need to close on the preferred exit workflow per this thread.

afiune: I forgot about this! Thank you for the reminder. Let me look at this 👀

afiune: https://github.com/lacework/go-sdk/pull/767

afiune (commented):
Should we close this PR?

hazedav:
> Should we close this PR?

I thought this was good to go we just wanted to refactor the "policy enforcement flows"...

> @hazedav lets refactor both "policy enforcement" workflows.

Also captured in this thread:

https://lacework.slack.com/archives/C01RSNZ0C5V/p1649787895250409?thread_ts=1648222632.313529&cid=C01RSNZ0C5V

hazedav:
Updated to use new`(FAIL-ON)` exit terminology.

integration/lql_test.go
```diff
@@ -170,3 +170,38 @@ func TestQueryRunURL(t *testing.T) {
 	assert.Empty(t, err.String(), "STDERR should be empty")
 	assert.Equal(t, 0, exitcode, "EXITCODE is not the expected one")
 }
+
+func TestQueryRunFailOnBadInput(t *testing.T) {
+	if os.Getenv("CI_BETA") == "" {
```
afiune, line 175: when are we enabling these tests?

cli/cmd/lql.go
```diff
@@ -162,6 +177,12 @@ func init() {
 		"validate_only", "", false,
 		"validate query only (do not run)",
 	)
+	// fail on count
+	queryRunCmd.Flags().StringVarP(
+		&queryCmdState.FailOnCount,
+		"fail_on_count", "", "",
+		"fail if the query matches the fail_on_count expression",
```
afiune, line 184: nit: normally we don't reference a flag name that is not a different one, I would suggest this but I am biased so, let's ask @dmurray-lacework for his opinion too
```suggestion
		"fail if the results from a query match the provided expression",
```

afiune (approved):
tenor-39258445

cli/cmd/lql.go
dmurray-lacework: I agree with the suggested change, if the flag name isn't apparent then the description should be worded differently

hazedav: Fixed.

cli/cmd/root.go
```diff
@@ -101,9 +101,13 @@ This will prompt you for your Lacework account and a set of API access keys.`,
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() (err error) {
 	defer func() {
-		var e *vulnerabilityPolicyError
-		if errors.As(err, &e) {
-			exitwithCode(e, e.ExitCode)
+		var vpe *vulnerabilityPolicyError
```
hazedav, line 104: cc @afiune .  we needed some of this magic.

PR #422, opened 2021-05-21
feat(cli): lacework policy command

hazedav:
Building out lacework policy command functionality.

Includes:
- `lacework policy list`
- `lacework policy show`
- `lacework policy create`
- `lacework policy update`
- `lacework policy delete`

ALLY-422


hazedav:
This is final review of work that had individual reviews per each of the subtasks under ALLY-422.  Thanks!

api/api.go
```diff
@@ -85,6 +85,9 @@ const (
 	apiLQLDescribe    = "v1/external/lql/describe"
 	apiLQLQuery       = "v1/external/lql/query"
 
+	// Custom Policies
+	apiPolicy = "external/lqlPolicies"
```
afiune, line 89:
```suggestion
	apiPolicy = "v1/external/lqlPolicies"
```

afiune (changes requested):
only one single blocker, the rest looks great!

api/api.go
hazedav: Fixed.

afiune (approved):
tenor-46500580

PR #423, opened 2021-05-26
feat(cli): v2 configure cmd undestands subaccounts

afiune:
This change also adds a global flag `--subaccount` to switch between
accounts. (ALLY-378)

**User Story**
As a Lacework CLI User, I would like to avoid introducing my account and
sub-account information, So that I don't do any mistakes and my CLI
configuration is faster.

**Description**
The Lacework CLI needs to understand the new JSON fields that users can
download from their accounts:
```json
{
  "keyId": "MINIALLY_138DB0DAA9084BD93E50F587A58DEA572E543A5C8C20006",
  "secret": "_f28a1c84014f34b948422395fb0691f3",
  "subAccount": "mini-ally",
  "account": "customerdemo.lacework.net"
}
```
The two new fields, `account`, and `subAccount`, will be stored inside
the configuration file stored at `~/.lacework.toml` and marked as
`version 2`
```toml
[default]
  account = "CORP"
  subaccount = "CORP-PROD"
  api_key = "CORP_A3C8910D0BDCDBD7AFD4355A1C5284104AAA2AE5953938C"
  api_secret = "_1234abca752e65f94651bcb6e6c798f1"
  version = 2
```

This change is backward compatible with the previous JSON format for
API keys as well as previous profiles inside the configuration file.

**Acceptance Criteria**
After completing this task, new Lacework CLI users that download the new
JSON format of an API key from the UI will be able to configure the CLI
without any human interaction (since all necessary fields are already
inside the JSON file).

Additionally, old profiles configured should continue to work.

JIRA: ALLY-374

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -123,14 +123,6 @@ func NewClient(account string, opts ...Option) (*Client, error) {
 	return c, nil
 }
 
-// RemoveSubaccount removes the subaccount from the API client
-func (c *Client) RemoveSubaccount() {
-	c.log.Debug("removing subaccount from client")
-	c.subaccount = ""
-	c.log.Debug("removing header", zap.String("field", "Account-Name"))
-	delete(c.headers, "Account-Name")
-}
-
 // WithSubaccount sets a subaccount into an API client
 func WithSubaccount(subaccount string) Option {
 	return clientFunc(func(c *Client) error {
--- a/api/client.go
+++ b/api/client.go
@@ -123,11 +123,23 @@ func NewClient(account string, opts ...Option) (*Client, error) {
 	return c, nil
 }
 
-// WithSubaccount sets a subaccount
+// RemoveSubaccount removes the subaccount from the API client
+func (c *Client) RemoveSubaccount() {
+	c.log.Debug("removing subaccount from client")
+	c.subaccount = ""
+	c.log.Debug("removing header", zap.String("field", "Account-Name"))
+	delete(c.headers, "Account-Name")
+}
+
+// WithSubaccount sets a subaccount into an API client
 func WithSubaccount(subaccount string) Option {
 	return clientFunc(func(c *Client) error {
-		c.log.Debug("setting up client", zap.Reflect("subaccount", subaccount))
-		c.subaccount = subaccount
+		if subaccount != "" {
+			c.log.Debug("setting up client", zap.String("subaccount", subaccount))
+			c.subaccount = subaccount
+			c.log.Debug("setting up header", zap.String("Account-Name", subaccount))
+			c.headers["Account-Name"] = subaccount
+		}
 		return nil
 	})
 }
--- a/api/user_profile.go
+++ b/api/user_profile.go
@@ -19,8 +19,9 @@
 package api
 
 import (
-	"regexp"
 	"strings"
+
+	"github.com/lacework/go-sdk/internal/domain"
 )
 
 // UserProfileService is the service that interacts with the UserProfile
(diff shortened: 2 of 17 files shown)


api/user_profile.go
```diff
@@ -30,18 +35,53 @@ func (svc *UserProfileService) Get() (response UserProfileResponse, err error) {
...
+	Username   string    `json:"username"`
+	OrgAccount bool      `json:"orgAccount"`
+	URL        string    `json:"url"`
+	OrgAdmin   bool      `json:"orgAdmin"`
+	OrgUser    bool      `json:"orgUser"`
+	Accounts   []Account `json:"accounts"`
+}
+
+func (p *UserProfile) OrgAccountName() string {
+	// TODO @afiune should we handle other datacenters?
```
afiune, line 51: I am trying to get access to an account deployed in our Frankfurt datacenter 

tenor-54643987

api/user_profile.go
```diff
@@ -30,18 +35,53 @@ func (svc *UserProfileService) Get() (response UserProfileResponse, err error) {
...
+	rx, err := regexp.Compile(`\.lacework\.net.*`)
+	if err == nil {
+		if split := rx.Split(p.URL, -1); len(split) != 0 {
+			return strings.ToLower(split[0])
+		}
+	}
+
+	return p.URL
+}
```
dmurray-lacework, line 60: Are there any tests for this regex? Ideally that tests US and EU.

api/client.go
```diff
@@ -123,11 +123,23 @@ func NewClient(account string, opts ...Option) (*Client, error) {
 	return c, nil
 }
 
-// WithSubaccount sets a subaccount
+// RemoveSubaccount removes the subaccount from the API client
+func (c *Client) RemoveSubaccount() {
```
dmurray-lacework, line 127: What situation would the subaccount need to be removed?

^ Not a comment on the usefulness of this function, genuinely not sure of the use case to remove the subaccount.

api/user_profile.go
```diff
@@ -30,18 +35,53 @@ func (svc *UserProfileService) Get() (response UserProfileResponse, err error) {
...
+	names := make([]string, 0)
+	orgAccountName := p.OrgAccountName()
+	for _, acc := range p.Accounts {
+		accName := strings.ToLower(acc.AccountName)
+		if accName == orgAccountName {
+			continue
+		}
+		if acc.Enabled() {
+			names = append(names, accName)
+		}
```
dmurray-lacework, line 72: Can this logic be condensed? Not sure which is more readable to be fair. Ignore if you don't like it (or my logic is incorrect)
```suggestion
		if accName != orgAccountName && acc.Enabled() {
			names = append(names, accName)
		}
```

dmurray-lacework (approved):

api/client.go
afiune: To access the primary account, instead of the sub-account

api/user_profile.go
afiune: I'm fine with either one of them. 🙌🏽

api/user_profile.go
afiune: tenor-168906724

api/user_profile.go
afiune: Now they are 🙌🏽

tenor-67817893

api/client.go
afiune: Just realized that this was part of the next change for the migration. I'll remove it from here. 👍🏽

PR #424, opened 2021-05-26
feat(cli): Add CSV rendering output for select commands

ipcrm:
Adds the ability to get CSV output from three commands using the `--csv` flag:

`lacework comp [csp] get-report`
`lacework vuln host list-cves`
`lacework vuln host list-hosts`

ALLY-465


afiune:
Make it so! 

tenor-90333659

cli/cmd/compliance.go
```diff
@@ -211,63 +214,70 @@ func complianceReportRecommendationsTable(recommendations []api.ComplianceRecomm
...
-			tableFunc(func(t *tablewriter.Table) {
-				t.SetBorder(false)
-				t.SetAutoWrapText(false)
-				t.SetColumnSeparator(" ")
-			}),
-		),
-	)
-
-	if compCmdState.Details || complianceFiltersEnabled() {
+	if compCmdState.Csv {
```
dmurray-lacework, line 217: The check for CSV should live along side the json flag check and not in the buildComplianceReportTable function.

It'd be great if there was symmetry across the alternative outputs.

`			if cli.JSONOutput() {
				return cli.OutputJSON(report)
			}

			if compCmdState.Csv {
				return cli.OutputCSV(report)
			}
`

cli/cmd/csv_render.go
```diff
@@ -0,0 +1,54 @@
...
+func csvCleanData(input []string) []string {
+	var data []string
+	for _, h := range input {
+		data = append(data, strings.Replace(h, "\n", "", -1))
+	}
+	return data
+}
+
+// Used to produce CSV output
+func renderAsCSV(headers []string, data [][]string) string {
```
dmurray-lacework, line 37: I think this would be better placed in outputs.go alongside the OutputJson function.

`func (c *cliState) OutputCSV(v interface{})
`

As the CSV writer only operates on []string we could use 	"github.com/fatih/structs"

something to the likes of this

```
	for _, header := range structs.Fields(v) {
		headers = append(headers, header.Name())
	}
	
	for _, value := range structs.Values(v) {
		values = append(values, value.(string))
	}
```

cli/cmd/compliance.go
```diff
@@ -211,63 +214,70 @@ func complianceReportRecommendationsTable(recommendations []api.ComplianceRecomm
...
+						tableFunc(func(t *tablewriter.Table) {
+							t.SetBorder(false)
+							t.SetColumnSeparator(" ")
+						}),
+					),
+				}},
 				tableFunc(func(t *tablewriter.Table) {
 					t.SetBorder(false)
-					t.SetRowLine(true)
+					t.SetAutoWrapText(false)
```
dmurray-lacework, line 247: Does this have any effect on the existing outputs?

cli/cmd/compliance_aws.go
```diff
@@ -84,6 +84,10 @@ Then navigate to Settings > Integrations > Cloud Accounts.
 		Use:     "get-report <account_id>",
 		Aliases: []string{"get"},
 		PreRunE: func(_ *cobra.Command, _ []string) error {
+			if compCmdState.Csv {
+				cli.NonInteractive()
+			}
```
dmurray-lacework, line 89: Does interactive affect the csv output?

cli/cmd/compliance.go
ipcrm: @dmurray-lacework The reason I did not do this is because the CSV output should be the same data we feed through the tables (which has had some further processing than the raw response).  That's not the case with JSON output where we drop the raw body.  I felt it was better to not have duplicate code to process the response body (which would be required in the suggestion above).  

Thoughts on this?

cli/cmd/compliance_aws.go
ipcrm: Yes, with the CSV output we do not want any of the spinners/upgrade-notes/etc to output, so this disables those.

ipcrm:
@dmurray-lacework As discussed, I'll do the following:

1) move the renderCSV stuff into outputs.go 
2) move the render out of the buildComplianceReportTable and do the ‘pre-processing’ in a block just below the JSON output.

dmurray-lacework:
make it so

dmurray-lacework:
make it so

cli/cmd/outputs.go
```diff
@@ -64,3 +65,32 @@ func (c *cliState) FormatJSONString(s string) (string, error) {
...
+	return data
+}
+
+// Used to produce CSV output
+func renderAsCSV(headers []string, data [][]string) string {
+	csvOut := &strings.Builder{}
+	csv := csv.NewWriter(csvOut)
+
+	if len(headers) > 0 {
+		csv.Write(csvCleanData(headers))
```
dmurray-lacework, line 84: The test build pipeline is failing here on linting check. Can you run a `make lint` and correct the issue its throwing.

cli/cmd/compliance_aws.go
afiune: I have a feeling that you are trying to pipe the CSV output to a file `lacework comp ... > file.csv` which in that case, the user should use the global flag `--noninteractive`.

Are we trying to download the file like the PDF flag?

cli/cmd/compliance_aws.go
```diff
@@ -236,6 +250,10 @@ func init() {
 		"download report in PDF format",
 	)
 
+	complianceAwsGetReportCmd.Flags().BoolVar(&compCmdState.Csv, "csv", false,
+		"download report in CSV format",
```
afiune, line 254: if we are not downloading a report to disk, let's just say we change the format/output

cli/cmd/compliance_azure.go
```diff
@@ -275,6 +289,11 @@ func init() {
 		"download report in PDF format",
 	)
 
+	// Output the report in CSV format
+	complianceAzureGetReportCmd.Flags().BoolVar(&compCmdState.Csv, "csv", false,
+		"download report in CSV format",
```
afiune, line 294: this one too

cli/cmd/vuln_host.go
```diff
@@ -242,25 +254,35 @@ To list the CVEs found in the hosts of your environment run:
...
-					cli.OutputHuman("There are no offline hosts.\n")
+			if vulCmdState.Csv {
+				cli.OutputHuman(
+					renderAsCSV(
+						[]string{"Machine ID", "Hostname", "External IP", "Internal IP",
+							"Os/Arch", "Provider", "Instance ID", "Vulnerabilities", "Status"},
+						rows,
+					),
+				)
+			} else {
```
afiune, line 265: nitpick: we prefer to avoid `if .. else .. ` statements when possible, this `if` could end with `return nil`

cli/cmd/vuln_host.go
```diff
@@ -1002,16 +1029,27 @@ func buildListCVEReports(cves []api.HostVulnCVE) error {
...
+
+		if vulCmdState.Csv {
+			cli.OutputHuman(
+				renderAsCSV(
+					[]string{"CVE Count", "Severity", "Package", "Current Version", "Fix Version", "Pkg Status", "Hosts"},
+					packages,
+				),
+			)
+			return nil
+		} else {
```
afiune, line 1041: same here, avoid nested things...

afiune (changes requested):
After our conversation, we agreed that you will add a new `csvOutput bool` field to the CLI state. Great work abstracting this feature on a smaller change!! We are so close to merging this!

tenor-157471280

afiune:
Make it so!!!! 👍🏽

cli/cmd/outputs.go
```diff
@@ -64,3 +66,46 @@ func (c *cliState) FormatJSONString(s string) (string, error) {
...
+
+// OutputCSV will print out the provided headers/data in CSV format if the cli state is
+// configured to output CSV
+func (c *cliState) OutputCSV(headers []string, data [][]string) error {
+	err, csv := renderAsCSV(headers, data)
+	if err != nil {
+		return err
+	}
+
+	fmt.Print(csv)
```
afiune, line 109: We prefer to control the file descriptor where we will print out, lets follow the above standard:
```suggestion
	fmt.Fprint(os.Stdout, csv)
```

cli/cmd/outputs.go
```diff
@@ -64,3 +66,46 @@ func (c *cliState) FormatJSONString(s string) (string, error) {
...
+func csvCleanData(input []string) []string {
+	var data []string
+	for _, h := range input {
+		data = append(data, strings.Replace(h, "\n", "", -1))
+	}
+	return data
+}
+
+// Used to produce CSV output
+func renderAsCSV(headers []string, data [][]string) (error, string) {
```
afiune, line 80: As a standard we always return the `error` as the last value of a function:
```suggestion
func renderAsCSV(headers []string, data [][]string) (string, error) {
```
For more info look at https://golang.org/doc/tutorial/handle-errors

cli/cmd/outputs.go
```diff
@@ -64,3 +66,46 @@ func (c *cliState) FormatJSONString(s string) (string, error) {
...
+		}
+	}
+
+	// Write any buffered data to the underlying writer (standard output).
+	csv.Flush()
+	return nil, csvOut.String()
+}
+
+// OutputCSV will print out the provided headers/data in CSV format if the cli state is
+// configured to output CSV
```
afiune, line 102: This statement is not completely true, the function doesn't check, and shouldn't check
if the CSV flag/state is enabled. This is the same behavior as the `OutputJSON()` func
```suggestion
// OutputCSV will print out the provided headers/data in CSV format
```

cli/cmd/outputs_test.go
```diff
@@ -0,0 +1,75 @@
...
+	)
+	assert.Equal(t,
+		csv,
+		expectedCsv,
+		"csv is not being formatted correctly")
+}
+
+func TestCSVDataCleanup(t *testing.T) {
+	data := csvCleanData([]string{"KEY\n HEADER\n VALUE", "VALUE,TEST\n"})
+	assert.False(t, strings.Contains(strings.Join(data, ""), "\n"), "data is not being cleaned up properly")
```
afiune, line 74: Lets use this instead: https://pkg.go.dev/github.com/stretchr/testify/assert#NotContains
```suggestion
	assert.NotContains(t, data, "\n", "data is not being cleaned up properly")
```

cli/cmd/outputs.go
```diff
@@ -64,3 +66,46 @@ func (c *cliState) FormatJSONString(s string) (string, error) {
...
+
+	for _, record := range data {
+		if err := csv.Write(csvCleanData(record)); err != nil {
+			return errors.Wrap(err, "Failed to build csv output"), ""
+		}
+	}
+
+	// Write any buffered data to the underlying writer (standard output).
+	csv.Flush()
+	return nil, csvOut.String()
```
afiune, line 98: uhmm, shouldn't we check for errors after flushing the buffer? 
```suggestion
	csv.Flush()
	return csv.Error(), csvOut.String()
```

afiune: Please, do change the order of the returned elements to be `return csvOut.String(), csv.Error()`

afiune (changes requested):
Great work so far @ipcrm !! A few more changes to make 

tenor-190380337

afiune:
Also, the pipeline is upset:
Screen Shot 2021-06-03 at 8 38 20 PM

Could you please run `make ci` on your local workstation, which will run all CI tests! 🙌🏽

afiune:
MAKE_IT_SO 

tenor-68093265

afiune (approved):
tenor-134167546

dmurray-lacework (approved):

afiune:
Updated documentation: https://github.com/lacework/go-sdk/wiki/CLI-Documentation#output-formats

PR #431, opened 2021-06-03
feat(cli): Automate migration of v2 configuration

afiune:
**User Story**
As a Lacework CLI User, I won't like to be disrupted during the
migration of the Lacework CLI from APIv1 to APIv2, So I can
continue accessing my Lacework data at all times and I don't have
to reconfigure all my profiles.

**Feature Description**
During the migration of our tools and integrations (in this case,
the Lacework CLI) we won't disrupt any of our customers, instead,
we are codifying and automating the migration by steps, tiny steps
that won't be noticed by our customers.

The first step of our plan is to automate the migration of the
configuration of existing customers so that the Lacework CLI can access
both, APIv1 and APIv2 at the same time for a short period of time (while
we slowly migrate existing functionality to APIv2). To do that, we send
an APIv1 request that describes the type of Lacework Account that our
customers have, there are two types, standalone accounts, and
organizational accounts.

For standalone accounts, the new configuration version 2 might look
practically the same as version 1, but the fact that we add the version
to the configuration will indicate that we no longer need to migrate it.

Example of version 1 standalone config:
```toml
[default]
  account = "standalone"
  api_key = "STANDALONE_13C8910D0BDCDBD7AFD4355A1C5284104AAA2AE5953938C"
  api_secret = "_123abc32a752e5f94651bcb6e6c798f1"
```

Example of version 2 standalone config:
```toml
[default]
  account = "standalone"
  api_key = "STANDALONE_13C8910D0BDCDBD7AFD4355A1C5284104AAA2AE5953938C"
  api_secret = "_123abc32a752e5f94651bcb6e6c798f1"
  version = 2
```

For organizational accounts, the new configuration version 2 will have a
new field named `subaccount`, and the field account will change to point
to the primary account (unique account URL).

Example of version 1 organizational account config:
```toml
[default]
  account = "sub-account"
  api_key = "SUBACCOUNT_13C8910D0BDCDBD7AFD4355A1C5284104AAA2AE5953938C"
  api_secret = "_123abc32a752e5f94651bcb6e6c798f1"
```

Example of version 2 standalone config:
```toml
[default]
  account = "org-account"
  subaccount = "sub-account"
  api_key = "SUBACCOUNT_13C8910D0BDCDBD7AFD4355A1C5284104AAA2AE5953938C"
  api_secret = "_123abc32a752e5f94651bcb6e6c798f1"
  version = 2
```

For more details, refer to the migration document of the EPIC.

**Acceptance Criteria**
Users using this feature will start the process of migrating their
Lacework CLI configuration. If we do this right, our customers won't
notice that this change happened. Additionally, we will be able to start
leveraging APIv2 endpoints from this new version of the CLI.

This ticket adds a new Honeycomb Feature Event (metric) to detect
when customers have been migrated to the new CLI configuration v2.

Feature Name: `migrate_config_v2`

JIRA https://lacework.atlassian.net/browse/ALLY-376

### DIFF
--- a/cli/cmd/file.go
+++ b/cli/cmd/file.go
@@ -16,34 +16,15 @@
 // limitations under the License.
 //
 
-package api
+package cmd
 
-import "github.com/lacework/go-sdk/internal/domain"
+import "os"
 
-// AccountService is a service that interacts with Account related
-// endpoints from the Lacework Server
-type AccountService struct {
-	client *Client
-}
-
-func (svc *AccountService) GetOrganizationInfo() (
-	response accountOrganizationInfoResponse,
-	err error,
-) {
-	err = svc.client.RequestDecoder("GET",
-		apiAccountOrganizationInfo,
-		nil,
-		&response,
-	)
-	return
-}
-
-type accountOrganizationInfoResponse struct {
-	OrgAccount    bool   `json:"orgAccount"`
-	OrgAccountURL string `json:"orgAccountUrl,omitempty"`
-}
-
-func (r accountOrganizationInfoResponse) AccountName() string {
-	d, _ := domain.New(r.OrgAccountURL)
-	return d.String()
+// fileExists checks if a file exists and is not a directory
+func fileExists(filename string) bool {
+	f, err := os.Stat(filename)
+	if os.IsNotExist(err) {
+		return false
+	}
+	return !f.IsDir()
 }
--- a/cli/cmd/migration.go
+++ b/cli/cmd/migration.go
@@ -1,0 +1,168 @@
+//
+// Author:: Salim Afiune Maya (<afiune@lacework.net>)
+// Copyright:: Copyright 2021, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package cmd
+
+import (
+	"fmt"
+	"os"
+	"path"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/viper"
+
+	"github.com/lacework/go-sdk/lwconfig"
+)
+
+// The name of the directory we will store backups of configuration files before migrating them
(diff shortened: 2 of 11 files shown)


cli/cmd/file.go
```diff
@@ -0,0 +1,30 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package cmd
+
+import "os"
+
+// fileExists checks if a file exists and is not a directory
+func fileExists(filename string) bool {
```
dmurray-lacework, line 24: Would this function not be better placed in internal directory?

cli/cmd/migration.go
```diff
@@ -0,0 +1,168 @@
...
+	}
+
+	backupDir := path.Join(cacheDir, ConfigBackupDir)
+	if err := os.MkdirAll(backupDir, 0755); err != nil {
+		return "", err
+	}
+
+	backupCfgPath := path.Join(backupDir,
+		fmt.Sprintf(".lacework.toml.%s.%s.bkp",
+			time.Now().Format("20060102150405"), newID()),
```
dmurray-lacework, line 161: Is there a const from the time package can be used instead of string "20060102150405"

cli/cmd/root.go
```diff
@@ -64,7 +64,13 @@ This will prompt you for your Lacework account and a set of API access keys.`,
...
 					return nil
 				}
-				return cli.NewClient()
+
+				if err := cli.NewClient(); err != nil {
+					return err
+				}
+
+				// @afiune execute any necessary migration, if any
+				return cli.Migrations()
```
dmurray-lacework, line 73: Have you thought about moving cli.Migrations() call inside cli.NewClient()?

dmurray-lacework (approved):
Have a few questions, but looks good!

cli/cmd/file.go
afiune: I can move it but it felt weird when I created a file package since there is a native one already that doesn't do this check out-of-the-box. 

Do you want me to move it? If so, help me with a package name.

cli/cmd/migration.go
afiune: I couldn't find one, do you know if there is one?

At this point I don't think it will be common to create backup files outside a migration so it is not a standard.

cli/cmd/root.go
afiune: No, mainly because the creation of an API client has nothing to do with the migration of the CLI configuration file.

cli/cmd/migration.go
dmurray-lacework: I had a look there, and as far as i can see there's no const in time package for it. 👍  Leave it as is.

cli/cmd/file.go
dmurray-lacework: I understand what you mean. Let's keep it here for now, if we come to reuse this function we can tackle that then.

afiune:
tenor-22550793

PR #438, opened 2021-06-07
refactor(cli): improved LQL service func names

hazedav:
While working on the `lacework policy` command feedback was given as it related to the redundancy of function names of an api service.  For instance we don't need `LQL.CreateQuery` as "Query" is implied.  Thus we have simplified this to `LQL.Create`...

ALLY-458


hazedav:
I'll get these conflicts fixed up and we'll put this out for final review.

hazedav:
`dev7` is 😢 so I can't verify integration tests at the moment...

afiune (approved):
tenor-201095887

PR #441, opened 2021-06-08
feat(cli): LQLv2 support

hazedav:
Updating API and CLI to support LQLv2

1. api-server is now returning `lql_id` and `query_text` instead of `LQL_ID` and `QUERY_TEXT`.
2. api-server response body for lql create/update was changed.
3. handling for lqlv2 semantics

ALLY-504


afiune (approved):
tenor-243512984

PR #443, opened 2021-06-09
fix(cli): better error messaging for policy create

hazedav:
If policy input does not `json.Unmarshal` wrap a the error with `policy must be valid JSON`

ALLY-507


afiune (approved):
tenor-73592814

PR #445, opened 2021-06-16
fix(api): Enable FirstSeenTime field in hostVulnListHostsResponse

dmurray-lacework:
Fixes https://lacework.atlassian.net/browse/ALLY-528

The response from `v1/external/vulnerabilities/host/cveId/<cve-id>` contains a field that was previously not returned in hostVulnListHostsResponse
 ` "first_seen_time": "Mon, 12 Apr 2021 20:00:00 -0700"`


lwtime/rfc1123z.go
```diff
@@ -0,0 +1,20 @@
...
+	time.Time
+}
+
+func (rfc *RFC1123Z) UnmarshalJSON(b []byte) (err error) {
+	t := strings.Trim(string(b), `"`)
+	rfc.Time, err = time.Parse(time.RFC1123Z, t)
+	return
+}
+
+func (rfc *RFC1123Z) MarshalJSON() ([]byte, error) {
```
afiune, line 18: Could we return the stand time format for all Time types? We use https://datatracker.ietf.org/doc/html/rfc3339

afiune (changes requested):

lwtime/rfc1123z.go
dmurray-lacework: Fixed

afiune (approved):
tenor-267483460

PR #453, opened 2021-06-22
feat(api): propagate APIv2 error messages

afiune:
New APIv2 endpoints contain the error message inside the `message`
field at the top level of the JSON response, we will now propagate this
message.

Example API v2 error response:
```
{
  "message": "This is an APIv2 error messages. Catch it!"
}
```


afiune:
tenor-122313009

PR #455, opened 2021-06-24
refactor(cli): Switch the creation of Cloudtrail Integrations to APIv2

afiune:
As part of our plan to migrate to APIv2, we are slowly switching existing functionality
like this one. Here we are creating Cloudtail Integrations in the CLI via APIv2, our next
move will be to switch this same functionality in Terraform and other integrations.

This change is also enabling our users to be able to create CloudTrail integrations at
the organization level when they want to add an Org Account Mappings file, which
before, in APIv1, it was disallowed.

When a user tries to create such Org CloudTrail integration, they will experience:

Screen Shot 2021-06-24 at 9 26 08 AM

JIRA: https://lacework.atlassian.net/browse/ALLY-539

### DIFF
--- a/cli/cmd/cli_state.go
+++ b/cli/cmd/cli_state.go
@@ -196,6 +196,11 @@ func (c *cliState) NewClient() error {
 		apiOpts = append(apiOpts, api.WithApiV2())
 	}
 
+	if c.OrgLevel {
+		c.Log.Debug("accessing organization level data sets")
+		apiOpts = append(apiOpts, api.WithOrgAccess())
+	}
+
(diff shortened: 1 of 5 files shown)


JessicaG:
Maybe not for this request, but in general, as we add more to this SDK, how do we feel about using switch statements over if/else? Readability would be the primary driver here, I think performance is miniscule.

afiune:
@JessicaG I LOOOOVE switch statements over if/else. I am 💯   on aboard!! 

Though, could you point me to where did you find those if/else? 😉

dmurray-lacework (approved):

PR #462, opened 2021-06-29
fix(cli): non-interactive should not run daily version check

afiune:
If a user is running in non-interactive mode, the Lacework CLI should avoid
the daily version check feature.

JIRA: https://lacework.atlassian.net/browse/ALLY-551

### DIFF
--- a/cli/cmd/cli_state.go
+++ b/cli/cmd/cli_state.go
@@ -216,7 +216,7 @@ func (c *cliState) NewClient() error {
 
 // InteractiveMode returns true if the cli is running in interactive mode
 func (c *cliState) InteractiveMode() bool {
-	return !c.nonInteractive && !c.csvOutput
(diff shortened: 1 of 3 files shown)


cli/cmd/cli_state.go
```diff
@@ -216,7 +216,7 @@ func (c *cliState) NewClient() error {
 
 // InteractiveMode returns true if the cli is running in interactive mode
 func (c *cliState) InteractiveMode() bool {
-	return !c.nonInteractive && !c.csvOutput
+	return !c.nonInteractive && !c.csvOutput && !c.jsonOutput
```
dmurray-lacework, line 219: This is kind of a different task than what this ticket intended. I think it's probably correct, but i'd maybe consider either adding more tests for this use case or moving it out of this PR.

dmurray-lacework (approved):
Looks good, just a comment on adding the json check along side this change.

PR #466, opened 2021-07-02
feat(cli): 'compliance google list' returns all projects and org IDs

dmurray-lacework:
Relates to: https://lacework.atlassian.net/browse/ALLY-526

Usage: `lacework compliance google list`
Outputs 
```
  ORGANIZATION ID           PROJECT ID
------------------+------------------------------
  n/a               gcp-project-one
  n/a               gcp-project-two
```


cli/cmd/compliance_gcp.go
```diff
@@ -32,10 +32,46 @@ import (
...
 var (
+	// complianceGcpListCmd represents the list sub-command inside the gcp command
+	complianceGcpListCmd = &cobra.Command{
+		Use:   "list",
+		Short: "list gcp projects and organizations",
+		Long:  `List all GCP projects and organization IDs.`,
+		RunE: func(_ *cobra.Command, args []string) error {
+			var (
+				gcpAccounts   []gcpAccount
+				response, err = cli.LwApi.Integrations.ListGcpCfg()
```
dmurray-lacework, line 43: This just checks gcp_cfg integration. Need to add audit log too.

cli/cmd/compliance_gcp.go
```diff
@@ -32,10 +32,46 @@ import (
...
+
+			if err != nil {
+				return errors.Wrap(err, "unable to list gcp projects/organizations")
+			}
+
+			for _, gcp := range response.Data {
+				gcpIntegration := gcpAccount{}
+				// if organization account, fetch the project ids
+				if gcp.Data.IDType == "ORGANIZATION" {
+					gcpAccounts = append(gcpAccounts, getGcpAccounts(gcpIntegration.OrganizationID)...)
```
afiune, line 55: `gcpIntegration.OrganizationID` will always be an empty string since you just initialized at
line `52`, I think you meant to use `gcp.Data.ID` instead.
```suggestion
					gcpAccounts = append(gcpAccounts, getGcpAccounts(gcp.Data.ID)...)
```

Still, this shows that we need to test this a bit more, could you please fix this issue and make sure we
can add more tests?

afiune (changes requested):

cli/cmd/compliance_gcp.go
```diff
@@ -32,10 +32,46 @@ import (
...
+					gcpIntegration = gcpAccount{OrganizationID: "n/a", ProjectID: gcp.Data.ID}
+					gcpAccounts = append(gcpAccounts, gcpIntegration)
+				}
+			}
+
+			for _, gcp := range gcpAccounts {
+				rows = append(rows, []string{gcp.ProjectID, gcp.OrganizationID})
+			}
+
+			cli.OutputHuman(renderSimpleTable([]string{"Project ID", "Organization ID"}, rows))
```
afiune, line 66: nit: I feel that it is better to show ORG ID and then Project ID like this
```suggestion
			cli.OutputHuman(renderSimpleTable([]string{"Organization ID", "Project ID"}, rows))
```

why? Because of 1) the hierarchy orgs can have projects and 2) the user can copy and paste the like to download the compliance report which asks for:
```
lacework compliance google get-report <organization_id> <project_id>
```

cli/cmd/compliance_gcp.go
dmurray-lacework: Did some more tests locally. At the minute while we're making changes to our integration test automation I don't want to add a dependency on having a gcp org account(yet). But I'll make a ticket to add tf and integration test for this.

cli/cmd/compliance_gcp.go
```diff
@@ -327,6 +355,21 @@ func splitIDAndAlias(text string) (id string, alias string) {
 	return
 }
 
+func getGcpAccounts(orgID string) []gcpAccount {
+	var accounts []gcpAccount
+	projectsResponse, err := cli.LwApi.Compliance.ListGcpProjects(orgID)
+	if err != nil {
+		return accounts
```
afiune, line 362: could we at least log the error as a
```go
cli.Log.Warnf("unable to list gcp projects", "org_id", orgID, "error", err.Error())
```

cli/cmd/compliance_gcp.go
```diff
@@ -336,3 +379,37 @@ type cliComplianceIDAlias struct {
...
+		if gcp.Data.IDType == "ORGANIZATION" {
+			gcpAccounts = append(gcpAccounts, getGcpAccounts(gcp.Data.ID)...)
+		} else if !containsDuplicateProjectID(gcpAccounts, gcp.Data.ID) {
+			gcpIntegration := gcpAccount{OrganizationID: "n/a", ProjectID: gcp.Data.ID}
+			gcpAccounts = append(gcpAccounts, gcpIntegration)
+		}
+	}
+
+	sort.Slice(gcpAccounts, func(i, j int) bool {
+		return gcpAccounts[i].ProjectID < gcpAccounts[j].ProjectID
```
afiune, line 402: I am not sure about this sorting mainly because, lets say there are two orgs, and each org has 50 projects, we would sort only by project and that will mix the grouping of the orgs. From a UX perspective, is that what we want¿?

In best case scenario, we should find a way to double sort, by org and then by projects.

afiune (approved):
Nitpicks but this code looks good now. Nice work.

Could you also update the description of the PR to match the new output?

PR #467, opened 2021-07-02
feat(cli): apiv2 support for lacework query command

hazedav:
Updating `lacework query` to support apiv2 based on:
https://qan.qan.corp.lacework.net/api/v2/docs#tag/Queries

This also requires that we adopt a markup format for query definitions.

The following commands are still using v1 endpoints as v2 endpoints do not yet exist:
`lacework query validate` (v1/external/lql/compile)
`lacework query sources` (v1/external/lql/dataSources)
`lacework query describe` (v1/external/lql/describe)
`lacework query run` (v1/external/lql/query)

ALLY-537


Makefile
```diff
@@ -41,6 +41,16 @@ integration: build-cli-cross-platform integration-only
 integration-only:
 	PATH=$(PWD)/bin:${PATH} go test -v github.com/lacework/go-sdk/integration -timeout 30m
 
+integration-lql: build-cli-cross-platform integration-lql-only
+
+integration-lql-only:
+	PATH=$(PWD)/bin:${PATH} go test -v github.com/lacework/go-sdk/integration -run TestQuery* -timeout 30m
```
afiune, line 47: Do you need the `*`? 🤔  Just curious since I never used it before.

api/client.go
```diff
@@ -49,7 +49,7 @@ type Client struct {
 	Compliance      *ComplianceService
 	Events          *EventsService
 	Integrations    *IntegrationsService
-	LQL             *LQLService
+	Query           *QueryService
```
afiune, line 52: If Query is using V2 endpoints, could we move it to the V2 service that groups them all?

api/lql.go
```diff
@@ -172,75 +65,78 @@ func (q LQLQuery) ValidateRange(allowEmptyTimes bool) (err error) {
...
+	response QueriesResponse,
+	err error,
+) {
+	err = svc.client.RequestDecoder("GET", apiV2Queries, nil, &response)
+	return
 }
 
-func (svc *LQLService) GetByID(queryID string) (
-	response LQLQueryResponse,
+func (svc *QueryService) GetByID(queryID string) (
```
afiune, line 107: Have you tested that the GET request here returns 404 when the query does NOT exist? 

I say this since many endpoints don't work as expected, would be good that you check that this does work fine.

api/lql.go
```diff
@@ -19,151 +19,44 @@
...
-		return t.UTC().Format(time.RFC3339), err
-	}
-	// parse time as millis
-	if i, err := strconv.ParseInt(inTime, 10, 64); err == nil {
-		return time.Unix(0, i*int64(time.Millisecond)).UTC().Format(time.RFC3339), err
-	}
-	return "", errors.New(fmt.Sprintf("unable to parse time (%s)", inTime))
+type QueriesResponse struct {
+	Data    []Query `json:"data"`
+	Message string  `json:"message"`
```
afiune, line 56: What is this `Message`field? I don't see it in the API v2 specifications

api/lql.go
```diff
@@ -19,151 +19,44 @@
...
-		if end, err = time.Parse(time.RFC3339, q.EndTimeRange); err != nil {
-			return
-		}
-	} else if allowEmptyTimes {
-		end = time.Now()
-	} else {
-		err = errors.New("end time must not be empty")
-		return
-	}
+func validateQueryRange(start, end time.Time) (err error) {
```
afiune, line 59: nit: curious if this time validation exists already inside `lwtime`, if not, disregard.

api/lql.go
```diff
@@ -172,75 +65,78 @@ func (q LQLQuery) ValidateRange(allowEmptyTimes bool) (err error) {
...
+		fmt.Sprintf("%s/%s", apiV2Queries, url.QueryEscape(queryID)),
+		uq,
+		&response,
+	)
 	return
 }
 
-func (svc *LQLService) GetQueries() (LQLQueryResponse, error) {
-	return svc.GetByID("")
+func (svc *QueryService) GetQueries() (
```
afiune, line 99: We adopted the function name`List()` for listing all "resources" rather than `GetResources()`.

Example: `client.V2.AgentAccessTokens.List()` vs `client.V2.AgentAccessTokens.GetAgentAccessTokens()`

cli/cmd/lql.go
```diff
@@ -240,78 +244,132 @@ func inputQueryFromEditor(action string) (query string, err error) {
...
+      source {
+          CloudTrailRawEvents
+      }
+      filter {
+          EVENT_SOURCE = 's3.amazonaws.com'
+      }
+      return {
+          INSERT_ID
+      }
+  }"
```
afiune, line 322: nit: I don't think we need this `"`
```suggestion
  }
```

afiune (approved):
A few comments but the code structure looks healthy! 

tenor-40479105

Makefile
hazedav: I will verify.  It appears that `-run` is a regexp:

```
-run regexp
        Run only those tests and examples matching the regular expression.
        For tests the regular expression is split into smaller ones by
        top-level '/', where each must match the corresponding part of a
        test's identifier.
```

api/lql.go
hazedav: Confirmed; yes.

```
~/Applications/src/lacework/go-sdk ALLY-561* ⇡
❯ curl --location --request GET 'https://qan.qan.corp.lacework.net/api/v2/Queries/MyExampleQuery' --header 'Authorization: Bearer _ee0d0a838d01a4e77cb3170791810722' -I 
HTTP/1.1 404 Not Found
Server: nginx
Date: Thu, 08 Jul 2021 15:51:12 GMT
Content-Type: application/json
Content-Length: 55
Connection: keep-alive
RateLimit-Remaining: 119
RateLimit-Limit: 120
RateLimit-Reset: 0
Strict-Transport-Security: max-age=31536000; includeSubDomains;; preload
```

api/lql.go
hazedav: Message comes out when error.
```
❯ curl --location --request GET 'https://qan.qan.corp.lacework.net/api/v2/Queries/MyExampleQuery' \
--header 'Authorization: Bearer _ee0d0a838d01a4e77cb3170791810722'
{"message":"Error: Query id MyExampleQuery not found."}
```

api/lql.go
hazedav: We do not have validation in lwtime to ensure that a start time precedes end time, but this would be trivial to relocate to that package if so desired.

api/lql.go
hazedav: Did we adopt `Show()` for listing single "resource" ?

Makefile
hazedav: Works without `*`... Removing...

api/client.go
hazedav: @afiune is it cool that the entire service isn't currently, but will be soon using all v2 endpoints ?

api/lql.go
hazedav: Answer:
`GetByID()`

api/lql.go
hazedav: @afiune , I'm going to merge as-is with `Message` in the Response structs and I'll visit this with you again on the next PR...

PR #471, opened 2021-07-09
feat: Introducing a caching mechanism

afiune:
The Lacework CLI helps our customers interact with the Lacework Platform
from the command line. A few of the operations it performs are inspection of
events, management of compliance reports and vulnerability assessments, and
more. Some of these tasks send API request that fetches static artifacts such as
a compliance report or vulnerability assessments from container images, these
static artifacts (JSON payloads) never change but when the user tries to change
the output format using flags and filters, the Lacework CLI fetches these static
artifacts every time a command is executed. This new caching mechanism aims
to store on disk some of these static artifacts to speed up the execution of
commands as well as reducing the load on the API Server.

Another use case for this new caching mechanism is the ability to store tokens
on disk so that the Lacework CLI does not generate a new token on every 
single command execution, instead, we will now generate a token and store it
in the cache, and when the user runs another command, the CLI will load the
cached token and proceed with the execution of the command.

If the token stored in cache has expired, the `api` go client will detect it and
automatically generate a new one and then, via a new concept called 
`LifeCycleCallbacks`, will notify back to the CLI that the token needs to be
erased from the cache.

# Usage

Here are a few examples showing some basic capabilities.

**Simple CRUD Example:**
```go
cli.Cache.WriteString("key", "something useful")  // Create
myData := cli.Cache.Read("key")                   // Read
cli.Cache.Write("key", []byte("cool update"))     // Update
cli.Cache.Erase("key")                            // Delete
```

**Store something in the global cache**
```go
cli.Cache.WriteString("global/data", "some global information")
```

**Store something in the scoped cache**: this particular cache location
is dedicated to things we need to cache that are scoped to the account,
subaccount (or `"standalone"` if none), and API key id.
```go
cli.Cache.WriteString("token", "secret token")
```


integration/compliance_test.go
```diff
@@ -67,6 +67,7 @@ Global Flags:
   -s, --api_secret string   secret access key
       --debug               turn on debug logging
       --json                switch commands output from human-readable to json format
+      --nocache             turn off chaching
```
dmurray-lacework, line 70: typo
```suggestion
      --nocache             turn off caching
```

integration/help_test.go
```diff
@@ -77,6 +77,7 @@ Global Flags:
   -s, --api_secret string   secret access key
       --debug               turn on debug logging
       --json                switch commands output from human-readable to json format
+      --nocache             turn off chaching
```
dmurray-lacework, line 80: typo
```suggestion
      --nocache             turn off caching
```

integration/help_test.go
```diff
@@ -159,6 +160,7 @@ Flags:
...
```
dmurray-lacework, line 163: typo
```suggestion
      --nocache             turn off caching
```

api/auth.go
```diff
@@ -81,15 +97,14 @@ func WithToken(token string) Option {
...
 		c.auth.expiration = t
-		c.auth.expiresAt = time.Now().Add(time.Duration(t) * time.Second)
+		c.auth.expiresAt = time.Now().UTC().Add(time.Duration(t) * time.Second)
 		return nil
 	})
 }
 
 func (c *Client) TokenExpired() bool {
-	return time.Until(c.auth.expiresAt) <= 0
+	return c.auth.expiresAt.Sub(time.Now().UTC()) <= 0
```
dmurray-lacework, line 107: I'm not sure this is correct

Should be either  
```suggestion
	return c.auth.expiresAt.Sub(time.Now().UTC()) >= 0
```
or 
```suggestion
	return time.Now().UTC().Sub(c.auth.expiresAt.)) >= 0
```

dmurray-lacework (approved):

integration/compliance_test.go
afiune: Fixing this locally since I found more of these typos, nice catch! 💯

dmurray-lacework (approved):

PR #472, opened 2021-07-12
feat(cli): apiv2 support for lacework policy

hazedav:
- Adds YAML support
- Simplifies API contract
- Sets up V2 service

ALLY-561


afiune (approved):
tenor-97520708

Issue #440, opened 2021-06-08
lacework configure command doesn't remove subaccount


afiune:
If an org user runs `lacework configure` and selects the `PRIMARY` account,
the Lacework CLI does not remove the `subaccount` configuration from the
`~/.lacework.toml` file.

PR #474, opened 2021-07-15
fix(cli): remove subaccount during configure

afiune:
If an org user runs `lacework configure` and selects the PRIMARY account,
the Lacework CLI will now remove the subaccount from the configuration.

Closes https://github.com/lacework/go-sdk/issues/440

### DIFF
--- a/cli/cmd/configure.go
+++ b/cli/cmd/configure.go
@@ -307,7 +307,7 @@ func getSubAccountForOrgAdmins() (string, error) {
 	user, err := cli.LwApi.V2.UserProfile.Get()
 	cli.StopProgress()
 	if err != nil {
-		return "", errors.Wrap(err, "unable to access UserProfile endpoit")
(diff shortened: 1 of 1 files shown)


cli/cmd/configure.go
```diff
@@ -309,12 +307,7 @@ func getSubAccountForOrgAdmins() (string, error) {
 	user, err := cli.LwApi.V2.UserProfile.Get()
 	cli.StopProgress()
 	if err != nil {
-		cli.Log.Warnw("unable to access UserProfile endpoint",
-			"error", err,
-		)
-		// We do NOT error here since API v2 is sending 500 errors
-		// for mortal users, we need to fix this on the server side
-		return "", nil
+		return "", errors.Wrap(err, "unable to access UserProfile endpoit")
```
dmurray-lacework, line 310:
```suggestion
		return "", errors.Wrap(err, "unable to access UserProfile endpoint")
```

dmurray-lacework (approved):

PR #479, opened 2021-07-15
refactor(cli): policy update api args

hazedav:
Cleaner approach to handling PolicyID for update

ALLY-570


api/policy.go
```diff
@@ -141,14 +141,17 @@ func (svc *PolicyService) Get(policyID string) (
...
 	response PolicyResponse,
 	err error,
 ) {
-	if policyID == "" {
+	if up.PolicyID == "" {
 		err = errors.New("policy ID must be provided")
 		return
 	}
+	var policyID = up.PolicyID
+	up.PolicyID = "" // omit this for PATCH
```
afiune, line 153: why do you need to omit the ID for PATCH?

afiune: oh, because the ID is in the URL, gotcha... 🤔

hazedav: That's correct.

hazedav:
@afiune / @dmurray-lacework any additional concerns here?

afiune (approved):
Not from my side! 

tenor-46308957

PR #482, opened 2021-07-16
fix: Output gcp comp list with --json flag

dmurray-lacework:

### DIFF
--- a/cli/cmd/compliance_gcp.go
+++ b/cli/cmd/compliance_gcp.go
@@ -55,7 +55,7 @@ var (
 			}
 
 			if cli.JSONOutput() {
-				return cli.OutputJSON(rows)
+				return cli.OutputJSON(gcpAccounts)
 			}
 
 			cli.OutputHuman(renderSimpleTable([]string{"Organization ID", "Project ID"}, rows))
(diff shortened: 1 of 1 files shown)


cli/cmd/compliance_gcp.go
```diff
@@ -54,6 +54,10 @@ var (
 				rows = append(rows, []string{gcp.OrganizationID, gcp.ProjectID})
 			}
 
+			if cli.JSONOutput() {
+				return cli.OutputJSON(rows)
```
afiune, line 58: will this be a valid JSON? I thought we needed to fix `[]` this output... 🤔

dmurray-lacework: We need to discuss the ux on this. [] is valid json and so is wrapping it in an object {[]}. So we need to find out what the better output is

afiune: I could not agree more with your, @dmurray-lacework !!! Lets talk about this.

afiune: @dmurray-lacework could you please update this PR so that the JSON
output matches this structure:
```json
[
  {
    "organization_id": "",
    "project_id": ""
  },
  {
    "organization_id": "",
    "project_id": ""
  }
]
```

afiune (changes requested):

afiune (approved):
tenor-199316860

PR #485, opened 2021-07-19
feat(cli): apiv2 support for query run

hazedav:
ExecuteByID is now restful
Eliminated InputQueryFromEnv
Simplified --validate_only

ALLY-562


api/lql_execute.go
```diff
@@ -0,0 +1,127 @@
...
+import (
+	"fmt"
+	"net/url"
+	"time"
+
+	"github.com/pkg/errors"
+)
+
+const (
+	RFC3339Milli = "2006-01-02T15:04:05.000Z"
```
afiune, line 30: Could we define this time type inside our `lwtime` package?

cli/cmd/lql.go
```diff
@@ -61,8 +61,9 @@ NOTE: This feature is not yet available!`,
 
 	// queryRunCmd represents the lql run command
 	queryRunCmd = &cobra.Command{
-		Use:   "run [query_id]",
-		Short: "run a query",
+		Aliases: []string{"execute"},
+		Use:     "run [query_id]",
```
afiune, line 65: The standard for arguments is `<arg>` not `[arg]`:
```suggestion
		Use:     "run <query_id>",
```

afiune (approved):
A couple of things that need to change, out of that, this change looks good! 💯

api/lql_execute.go
hazedav: Yes; absolutely...

cli/cmd/lql.go
hazedav: I'm pretty sure you told me that optional arguments use square brackets...

PR #489, opened 2021-07-21
fix(cli): evalFrequency optional for policyCreate

hazedav:
ALLY-585


afiune (approved):
I had to go to Jira to understand this PR, could we have a practice to
add a description to every PR? 🙌🏽  Thank you @hazedav 

tenor-212362065

hazedav:
> could we have a practice to add a description to every PR

I typically do, but incorrectly thought the title alone explained it...

PR #501, opened 2021-07-30
test(cli): updating policy integration resource

hazedav:
Policy guardrails dictate we specify partial policy id

ALLY-584


afiune (approved):
tenor-71463007

PR #502, opened 2021-08-03
feat: Add apiV2AlertChannelTest

dmurray-lacework:
https://lacework.atlassian.net/browse/ALLY-331

Add V2 Alert Channel Test function to api.go

### DIFF
--- a/api/alert_channels.go
+++ b/api/alert_channels.go
@@ -141,22 +141,7 @@ func (svc *AlertChannelsService) Test(guid string) error {
 	}
 
 	apiPath := fmt.Sprintf(apiV2AlertChannelTest, guid)
-	request, err := svc.client.NewRequest("POST", apiPath, nil)
-	if err != nil {
-		return err
-	}
-
-	resp, err := svc.client.Do(request)
-	if err != nil {
-		return err
-	}
-	err = checkErrorInResponse(resp)
-
-	if err != nil {
-		return err
-	}
-
-	return nil
+	return svc.client.RequestDecoder("POST", apiPath, nil, nil)
 }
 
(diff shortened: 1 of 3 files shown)


afiune (commented):
Is there a way we can add tests?

api/alert_channels_test.go
```diff
@@ -238,6 +238,44 @@ func generateAlertChannels(guids []string, iType string) string {
...
+		apiPath    = fmt.Sprintf("AlertChannels/%s/test", intgGUID)
+		fakeServer = lacework.MockServer()
+	)
+	fakeServer.UseApiV2()
+	fakeServer.MockToken("TOKEN")
+	defer fakeServer.Close()
+
+	fakeServer.MockAPI(apiPath,
+		func(w http.ResponseWriter, r *http.Request) {
+			if assert.Equal(t, "POST", r.Method, "Post() should be a POST method") {
```
afiune, line 253: I think you meant to say `Test()` should be `POST`
```suggestion
			if assert.Equal(t, "POST", r.Method, "Test() should be a POST method") {
```

api/alert_channels.go
```diff
@@ -134,6 +134,31 @@ func (svc *AlertChannelsService) Delete(guid string) error {
...
+	if err != nil {
+		return err
+	}
+	err = checkErrorInResponse(resp)
+
+	if err != nil {
+		return err
+	}
+
+	return nil
```
afiune, line 159: nit: I think this would have the same behavior. But I leave it to you to decide what is best.
```suggestion
	return svc.client.RequestDecoder("POST", apiPath, nil, nil)
```

afiune (approved):
Thank you so much for adding unit tests! 💯

tenor-263344365

PR #511, opened 2021-08-10
test(cli): go test modularity via build tags

hazedav:
Adds go build tags to integration tests
Specifies build tags in makefile
Disables policy testing

ALLY-603

### DIFF
--- a/integration/configure_unix_test.go
+++ b/integration/configure_unix_test.go
@@ -1,5 +1,5 @@
-// +build !windows
-//
+// +build configure configure,!windows
+
 // Author:: Salim Afiune Maya (<afiune@lacework.net>)
 // Copyright:: Copyright 2020, Lacework Inc.
 // License:: Apache License, Version 2.0
--- a/integration/configure_windows_test.go
+++ b/integration/configure_windows_test.go
@@ -1,5 +1,5 @@
-// +build !linux
-//
+// +build configure configure,!linux
+
 // Author:: Salim Afiune Maya (<afiune@lacework.net>)
 // Copyright:: Copyright 2020, Lacework Inc.
 // License:: Apache License, Version 2.0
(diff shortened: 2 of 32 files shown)


dmurray-lacework (approved):

integration/configure_windows_test.go
```diff
@@ -1,5 +1,5 @@
-// +build !linux
-//
+// +build configure configure,!linux
```
afiune, line 1: is this right? or should it be 
```suggestion
// +build configure,!linux
```

integration/configure_unix_test.go
```diff
@@ -1,5 +1,5 @@
-// +build !windows
-//
+// +build configure configure,!windows
```
afiune, line 1: is this right? or should it be

```suggestion
// +build configure,!windows
```

integration/framework_test.go
```diff
@@ -1,4 +1,5 @@
-//
+// +build core
+
```
afiune, line 2: Since no test would work without this "core" code, let's not tag it
```suggestion
```

integration/configure_windows_test.go
hazedav: `spaces are OR and commas are AND`

So this reads "include me when specifying configure, or when specifying configure and not linux"

integration/configure_unix_test.go
hazedav: `spaces are OR and commas are AND`

So this reads "include me when specifying configure, or when specifying configure and not windows"

integration/framework_test.go
hazedav: I don't think this works like that, but I will try it.

integration/configure_unix_test.go
afiune: TIL 🙌🏽

integration/configure_windows_test.go
afiune: TIL 🙌🏽

afiune (approved):
tenor-115716271

PR #519, opened 2021-08-16
feat(cli): switch over to API v2 by default

afiune:

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -21,6 +21,7 @@ package cmd
 import (
 	"fmt"
 	"os"
+	"strings"
 
 	homedir "github.com/mitchellh/go-homedir"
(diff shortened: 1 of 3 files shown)


dmurray-lacework (approved):

cli/cmd/root.go
```diff
@@ -65,12 +66,13 @@ This will prompt you for your Lacework account and a set of API access keys.`,
 					return nil
 				}
 
-				if err := cli.NewClient(); err != nil {
-					return err
+				err := cli.NewClient()
+				if err != nil && strings.Contains(err.Error(), "Invalid Account") {
```
afiune, line 70: we are slowing the error from `NewClient()` if it doesn't contain `Invalid Account`

PR #521, opened 2021-08-18
fix(cli): Detect if running in cloudshell to output correct update command

dmurray-lacework:
issue: https://lacework.atlassian.net/browse/ALLY-567

When cli is running in cloudshell the update command should output:
`curl https://raw.githubusercontent.com/lacework/go-sdk/main/cli/install.sh | bash -s -- -d $HOME/bin
`

### DIFF
--- a/cli/cmd/cli_unix.go
+++ b/cli/cmd/cli_unix.go
@@ -33,6 +33,13 @@ var promptIconsFunc = func(icons *survey.IconSet) {
 	icons.Question.Text = "▸"
 }
 
+// Env variables found in GCP, AWS and Azure cloudshell
+var (
+	gcpCloudEnv   = "CLOUD_SHELL"
+	awsCloudEnv   = "AWS_EXECUTION_ENV"
+	AzureCloudEnv = "POWERSHELL_DISTRIBUTION_CHANNEL"
+)
+
 // UpdateCommand returns the command that a user should run to update the cli
 // to the latest available version (unix specific command)
 func (c *cliState) UpdateCommand() string {
@@ -41,7 +48,17 @@ func (c *cliState) UpdateCommand() string {
   brew upgrade lacework-cli
 `
 	}
+
+	if isCloudshell() {
+		return `
+  curl https://raw.githubusercontent.com/lacework/go-sdk/main/cli/install.sh | bash -s -- -d $HOME/bin
+`
+	}
 	return `
   curl https://raw.githubusercontent.com/lacework/go-sdk/main/cli/install.sh | bash
 `
+}
+
+func isCloudshell() bool {
+	return os.Getenv(gcpCloudEnv) == "true" || os.Getenv(awsCloudEnv) == "Cloudshell" || os.Getenv(AzureCloudEnv) == "CloudShell"
 }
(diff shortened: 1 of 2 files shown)


cli/cmd/cli_unix.go
```diff
@@ -33,6 +33,13 @@ var promptIconsFunc = func(icons *survey.IconSet) {
 	icons.Question.Text = "▸"
 }
 
+// Env variables found in GCP, AWS and Azure cloudshell
+var (
+	gcpCloudEnv   = "CLOUD_SHELL"
+	awsCloudEnv   = "AWS_EXECUTION_ENV"
+	AzureCloudEnv = "POWERSHELL_DISTRIBUTION_CHANNEL"
```
rmoles, line 40: Is this available both in the Azure powershell and bash terminals?

afiune: tenor-44026482

dmurray-lacework: It does indeed. Confirmed this yesterday. Salim raised the same question.

cli/cmd/cli_unix.go
```diff
@@ -33,6 +33,13 @@ var promptIconsFunc = func(icons *survey.IconSet) {
 	icons.Question.Text = "▸"
 }
 
+// Env variables found in GCP, AWS and Azure cloudshell
+var (
```
afiune, line 37: nit
```suggestion
const (
```

cli/cmd/cli_unix.go
```diff
@@ -33,6 +33,13 @@ var promptIconsFunc = func(icons *survey.IconSet) {
 	icons.Question.Text = "▸"
 }
 
+// Env variables found in GCP, AWS and Azure cloudshell
```
afiune, line 36: let us make sure that this comment explains the motivation/reasoning why we look for these variables

dmurray-lacework, line 36:
```suggestion
// Env variables found in GCP, AWS and Azure cloudshell. Used to determine if cli is running on cloudshell.
```

cli/cmd/cli_unix.go
```diff
@@ -41,7 +48,17 @@ func (c *cliState) UpdateCommand() string {
...
+  curl https://raw.githubusercontent.com/lacework/go-sdk/main/cli/install.sh | bash -s -- -d $HOME/bin
+`
+	}
 	return `
   curl https://raw.githubusercontent.com/lacework/go-sdk/main/cli/install.sh | bash
 `
 }
+
+func isCloudshell() bool {
+	return os.Getenv(gcpCloudEnv) == "true" || os.Getenv(awsCloudEnv) == "Cloudshell" || os.Getenv(AzureCloudEnv) == "CloudShell"
```
afiune, line 63: Could we split these checks into three different functions?
* `isAwsCloudShell()`
* `isGcpCloudShell()`
* `isAzureCloudShell()`

These functions has to explain what it does (check if running on XYZ cloud shell) and what to we expect.

Then this function should call them all:
```suggestion
	return isAwsCloudShell() || isGcpCloudShell() || isAzureCloudShell()
```

afiune (commented):
write a few tests to verify this behavior please 🙌🏽

afiune (approved):
tenor-179775173

PR #522, opened 2021-08-18
feat(cli): More detailed CSV compliance reports

ipcrm:
Updating the CLI CSV output of the compliance reports to closely resemble the CSV exports from UI

JIRA: https://lacework.atlassian.net/browse/ALLY-534

### DIFF
--- a/cli/cmd/compliance.go
+++ b/cli/cmd/compliance.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"sort"
 	"strings"
+	"time"
 
 	"github.com/olekukonko/tablewriter"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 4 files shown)


dmurray-lacework:
make it so

cli/cmd/compliance.go
```diff
@@ -213,6 +214,90 @@ func complianceReportRecommendationsTable(recommendations []api.ComplianceRecomm
...
+	return record
+}
+
+func complianceCSVReportRecommendationsTable(
+	tenantAlias string,
+	accountAlias string,
+	reportType string,
+	reportTime time.Time,
+	recommendations []api.ComplianceRecommendation) [][]string {
+	out := [][]string{}
```
dmurray-lacework, line 259: I like the idea to make this generic across the 3 clouds. But rather than having param name that don't relate when the function is called (eg. when called from compliace_aws.go there is no tenant alias) Would it be possible to create a struct container these values and pass it in?

ipcrm: Of course! I rework this.

dmurray-lacework (approved):

afiune (approved):
Pulled down and tested, looks GREAT!!!

tenor-39278194

afiune:
make it so

afiune:
make it sooooooooooo

PR #523, opened 2021-08-19
fix(cli): removing policyUi from policy schema

hazedav:
Removing policyUi from all structures and tests

ALLY-612


hazedav:
I'm pretty sure this is solid, but I can't validate via automated testing as the running of integration tests is invoking 429 api rate limits exceeded every time I run it.  Will attempt on 4.20 when it rolls out.

dmurray-lacework (approved):

hazedav:
I was able to validate query/policy integration testing on customerdemo!

afiune (approved):
tenor-33061758

PR #524, opened 2021-08-19
fix(cli): query run via editor

hazedav:
Relocating cli start/stop progress for lacework query run

ALLY-615


afiune (approved):
tenor-196790344

afiune:
Changed the title just as a suggestion to avoid adding "fix" twice

PR #525, opened 2021-08-23
fix(cli): handle scan status 'Partial'

dmurray-lacework:
Add new scan status case, "Partial"


cli/cmd/vulnerability.go
```diff
@@ -392,6 +392,8 @@ func checkScanStatus(requestID string) (*api.VulnContainerAssessment, error, boo
 	switch status {
 	case "Success":
 		return &scan.Data, nil, false
+	case "Partial":
+		return &scan.Data, nil, false
```
dmurray-lacework, line 396: Should "scanning" be returned as true. Do we return the "Partial" result?

afiune: Return partial results

afiune (approved):
tenor-260416263

PR #528, opened 2021-08-23
feat(cli): stdin support for query/policy

hazedav:
Adds stdin support for lacework query
Adds stdin support for lacework policy
Adds integration testing

Example usage:
```
./sigmac -t lacework -c rules/cloud/aws_cloudtrail_disable_logging.yml -O query | lacework query run
```

```
lacework-dev query run < /Users/davidhazekamp/Applications/src/lacework/go-sdk/integration/test_resources/lql/LW_CLI_AWS_CTA_IntegrationTest.yaml
```

ALLY-616


cli/cmd/policy.go
```diff
@@ -183,6 +184,12 @@ func inputPolicy(cmd *cobra.Command) (string, error) {
 	if policyCmdState.URL != "" {
 		return inputPolicyFromURL(policyCmdState.URL)
 	}
+	// if running via stdin
+	stat, _ := os.Stdin.Stat()
```
afiune, line 188: As a best practice, we always check errors, even if it means that we couldn't read the standard input. At least log a debug message.

cli/cmd/policy.go
```diff
@@ -183,6 +184,12 @@ func inputPolicy(cmd *cobra.Command) (string, error) {
...
+	if (stat.Mode() & os.ModeCharDevice) == 0 {
```
afiune, line 189: I can see what we are trying to do here, support both "pipes" and "code injection" (via <)

I think I like the pipe idea, but the "code injection" not so much since we already have a flag for passing a file from disk. 

Could we limit this PR to enable only pipes using `os.ModeNamedPipe` rather than `os.ModeCharDevice`?

afiune (commented):
Im loving this approach, I have a couple comments though.

cli/cmd/policy.go
afiune: What I have used in the past is `if (stat.Mode() & os.ModeNamedPipe) != 0 {}` but if it works with your implementation Im ok with it.

cli/cmd/lql.go
```diff
@@ -198,6 +199,14 @@ func inputQuery(cmd *cobra.Command) (string, error) {
 	if queryCmdState.URL != "" {
 		return inputQueryFromURL(queryCmdState.URL)
 	}
+	// if running via stdin
+	stat, err := os.Stdin.Stat()
+	if err != nil {
+		cli.Log.Debugw("error retrieving stdin mode")
+	} else if (stat.Mode() & os.ModeCharDevice) == 0 {
```
afiune, line 206: I have never used `os.ModeCharDevice` but instead this:
```suggestion
	} else if (stat.Mode() & os.ModeNamedPipe) != 0 {
```

From the official Golang documentation https://pkg.go.dev/os
```
	ModeDevice     = fs.ModeDevice     // D: device file
	ModeNamedPipe  = fs.ModeNamedPipe  // p: named pipe (FIFO)
	ModeCharDevice = fs.ModeCharDevice // c: Unix character device, when ModeDevice is set
```

I have used `ModeDevice` and `ModeCharDevice` for files, not pipes. But if it works, Im +1

cli/cmd/lql.go
```diff
@@ -198,6 +199,14 @@ func inputQuery(cmd *cobra.Command) (string, error) {
 	if queryCmdState.URL != "" {
 		return inputQueryFromURL(queryCmdState.URL)
 	}
+	// if running via stdin
+	stat, err := os.Stdin.Stat()
+	if err != nil {
+		cli.Log.Debugw("error retrieving stdin mode")
```
afiune, line 205:
```suggestion
		cli.Log.Debugw("error retrieving stdin mode", "error", err.Error())
```

afiune (approved):
tenor-43838116

cli/cmd/lql.go
hazedav: I will try this again; thanks!

hazedav: So `os.ModeNamedPipe` is having a problem with the test setup as well.  I think I'm going to stick with the slightly more robust `else if (stat.Mode() & os.ModeCharDevice) == 0`...

PR #543, opened 2021-09-08
fix: Output correct resource group json

dmurray-lacework:
Fix resource groups cli json output


afiune (approved):
tenor-93550135

PR #546, opened 2021-09-09
fix: Change lastUpdated type to lwtime.Epoch

dmurray-lacework:

### DIFF
--- a/api/resource_groups_azure.go
+++ b/api/resource_groups_azure.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"strconv"
 
+	"github.com/lacework/go-sdk/lwtime"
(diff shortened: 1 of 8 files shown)


api/resource_groups_azure.go
```diff
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"strconv"
 
+	"github.com/lacework/go-sdk/lwtime"
 	"github.com/pkg/errors"
```
afiune, line 26: nit: group imports
```suggestion
	"github.com/pkg/errors"

	"github.com/lacework/go-sdk/lwtime"
```

afiune (approved):
Would be nice to add some tests for the `lwtime.Epoch` type.

PR #551, opened 2021-09-16
feat(telemetry): send information about flags

afiune:
With this change we are now sending telemetry about the flags that the
user use on every command execution, this will help us understand how
the Lacework CLI is used, what are the most used flags, and more
information to discover.

### DIFF
--- a/cli/cmd/honeyvent.go
+++ b/cli/cmd/honeyvent.go
@@ -95,6 +95,7 @@ type Honeyvent struct {
 	Arch          string      `json:"arch"`
 	Command       string      `json:"command,omitempty"`
 	Args          []string    `json:"args,omitempty"`
+	Flags         []string    `json:"flags,omitempty"`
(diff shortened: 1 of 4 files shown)


cli/cmd/honeyvent.go
```diff
@@ -218,3 +219,18 @@ func installMethod() string {
 	}
 	return ""
 }
+
+// parseFlags is a helper used to parse all the flags that the user provided
+func parseFlags(args []string) (flags []string) {
+	for len(args) > 0 {
+		arg := args[0]
+		args = args[1:]
+		if len(arg) == 0 || arg[0] != '-' || len(arg) == 1 {
```
dmurray-lacework, line 228: nit can reduce this. 
```suggestion
		if len(arg) <= 1 || arg[0] != '-' {
```

dmurray-lacework (approved):
Approved with 1 nit

Issue #540, opened 2021-08-30
Pass-through used profile for command recommendations


alannix-lw:
I'd like to have the currently used profile passed through in command recommendations.  For example, the container vuln scan command output currently looks like this:
```
lacework vuln container scan index.docker.io alannix/my-awesome-container latest -p my-awesome-profile
A new vulnerability scan has been requested. (request_id: 151e6d4c-3948-4430-80ed-b07f0c8d548e)

To track the progress of the scan, use the command:
  $ lacework vulnerability container scan-status 151e6d4c-3948-4430-80ed-b07f0c8d548e
```

I think we should add the profile if possible to make it more copy/paste - like this:
```
lacework vuln container scan index.docker.io alannix/my-awesome-container latest -p my-awesome-profile
A new vulnerability scan has been requested. (request_id: 151e6d4c-3948-4430-80ed-b07f0c8d548e)

To track the progress of the scan, use the command:
  $ lacework vulnerability container scan-status 151e6d4c-3948-4430-80ed-b07f0c8d548e -p my-awesome-profile
```

afiune:
JIRA: https://lacework.atlassian.net/browse/ALLY-634

PR #554, opened 2021-09-21
feat(cli): pass-through used profile in recommendations

afiune:
***Issue***
Closes https://github.com/lacework/go-sdk/issues/540
JIRA: https://lacework.atlassian.net/browse/ALLY-634

***Description:***
Pass through the currently used profile in command recommendations.

***Additional Info:***

Screen Shot 2021-09-21 at 5 12 49 PM


dmurray-lacework (approved):
LGTM

Doesn't have to be for this  PR, but we should consider a named func for cli outputs that are "suggestions". The func will still perform the same action but it'll be apparent where we are suggestion actions to the user

PR #565, opened 2021-10-01
feat(api): Webhook alert channel(v2)

dmurray-lacework:
***Issue***: https://lacework.atlassian.net/browse/ALLY-651

***Description:***
Migrate the Webhook alert channel to use the V2 API.

### DIFF
--- a/api/alert_channels_webhook.go
+++ b/api/alert_channels_webhook.go
@@ -47,5 +47,5 @@ type WebhookAlertChannelV2 struct {
 }
 
 type WebhookDataV2 struct {
-	Url string `json:"webhookUrl"`
+	WebhookUrl string `json:"webhookUrl"`
 }
--- a/api/alert_channels_webhook.go
+++ b/api/alert_channels_webhook.go
@@ -1,0 +1,51 @@
+//
+// Author:: Darren Murray (<darren.murray@lacework.net>)
+// Copyright:: Copyright 2021, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package api
+
+// GetWebhook gets a single Webhook alert channel matching the
+// provided integration guid
+func (svc *AlertChannelsService) GetWebhook(guid string) (
+	response WebhookAlertChannelResponseV2,
+	err error,
+) {
+	err = svc.get(guid, &response)
+	return
+}
+
+// UpdateWebhook updates a single Webhook integration on the Lacework Server
+func (svc *AlertChannelsService) UpdateWebhook(data AlertChannel) (
+	response WebhookAlertChannelResponseV2,
+	err error,
+) {
+	err = svc.update(data.ID(), data, &response)
+	return
+}
+
+type WebhookAlertChannelResponseV2 struct {
+	Data WebhookAlertChannelV2 `json:"data"`
+}
+
+type WebhookAlertChannelV2 struct {
+	v2CommonIntegrationData
+	Data WebhookDataV2 `json:"data"`
+}
+
+type WebhookDataV2 struct {
+	Url string `json:"webhookUrl"`
+}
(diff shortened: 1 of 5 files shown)


api/alert_channels_webhook.go
```diff
@@ -0,0 +1,51 @@
...
+	Data WebhookAlertChannelV2 `json:"data"`
+}
+
+type WebhookAlertChannelV2 struct {
+	v2CommonIntegrationData
+	Data WebhookDataV2 `json:"data"`
+}
+
+type WebhookDataV2 struct {
+	Url string `json:"webhookUrl"`
```
vatasha, line 50: nit: I feel like the field name should match the json field name.

vatasha, line 50:
```suggestion
	webhookUrl string `json:"webhookUrl"`
```

api/alert_channels_webhook_test.go
```diff
@@ -0,0 +1,135 @@
...
+	)
+	assert.Nil(t, err)
+
+	response, err := c.V2.AlertChannels.GetWebhook(intgGUID)
+	assert.Nil(t, err)
+	assert.NotNil(t, response)
+	assert.Equal(t, intgGUID, response.Data.IntgGuid)
+	assert.Equal(t, "integration_name", response.Data.Name)
+	assert.True(t, response.Data.State.Ok)
+	assert.Contains(t, response.Data.Data.Url, "https://hooks.webhook.com/?api-token=12345")
```
vatasha, line 61: nit: I think you can use an assert.Equals here not sure why there is a containment check.

vatasha, line 61:
```suggestion
	assert.Equal(t, "https://hooks.webhook.com/?api-token=12345", response.Data.Data.Url)
```

api/alert_channels_webhook_test.go
```diff
@@ -0,0 +1,135 @@
...
+	assert.Equal(t, "Webhook", webhookAlertChan.Type, "a new Webhook alert channel should match its type")
+	assert.Equal(t, 1, webhookAlertChan.Enabled, "a new Webhook alert channel should be enabled")
+	webhookAlertChan.IntgGuid = intgGUID
+
+	response, err := c.V2.AlertChannels.UpdateWebhook(webhookAlertChan)
+	assert.Nil(t, err)
+	assert.NotNil(t, response)
+	assert.Equal(t, intgGUID, response.Data.IntgGuid)
+	assert.True(t, response.Data.State.Ok)
+	assert.Contains(t, response.Data.Data.Url, "https://hooks.webhook.com/?api-token=12345")
```
vatasha, line 111: same comment as above

api/alert_channels_webhook_test.go
```diff
@@ -0,0 +1,135 @@
...
+		},
+	)
+	assert.Equal(t, "integration_name", webhookAlertChan.Name, "Webhook alert channel name mismatch")
+	assert.Equal(t, "Webhook", webhookAlertChan.Type, "a new Webhook alert channel should match its type")
+	assert.Equal(t, 1, webhookAlertChan.Enabled, "a new Webhook alert channel should be enabled")
+	webhookAlertChan.IntgGuid = intgGUID
+
+	response, err := c.V2.AlertChannels.UpdateWebhook(webhookAlertChan)
+	assert.Nil(t, err)
+	assert.NotNil(t, response)
```
vatasha, line 108:
```suggestion
	assert.NoError(t, err)
```

api/alert_channels_webhook_test.go
vatasha, line 111:
```suggestion
	assert.Equals(t, "https://hooks.webhook.com/?api-token=12345", response.Data.Data.Url)
```

api/alert_channels_webhook_test.go
```diff
@@ -0,0 +1,135 @@
...
+	assert.Equal(t, intgGUID, response.Data.IntgGuid)
+	assert.True(t, response.Data.State.Ok)
+	assert.Contains(t, response.Data.Data.Url, "https://hooks.webhook.com/?api-token=12345")
+}
```
vatasha, line 112: nit: Can use fmt.Sprintf here instead of string concatenation

vatasha (approved):
LGTM 🎉 . I have some minor nits but nothing blocking.

api/alert_channels_webhook_test.go
afiune: I don't think this is right

vatasha: I think this is correct. You can refer to a more complete way of doing it like so 

` actualObj, err := SomeFunction()
  if assert.NoError(t, err) {
	   assert.Equal(t, expectedObj, actualObj)
  }`

and the docs can be found here https://pkg.go.dev/github.com/stretchr/testify/assert#NoError

afiune: NoError is only for errors, we can change the NotNil above for this NotError but here we are trying to check that the response is not nil, which I think is the right thing.

afiune (approved):
There are some suggestions that need to be addressed.

api/alert_channels_webhook_test.go
dmurray-lacework: There's no concatenated string in this segment of code.

PR #567, opened 2021-10-04
docs: improve pull request template

afiune:
## Summary

Improve Pull Request template to render correctly and add useful information like,
add tests, how to run those tests, how to format code, run linter, built the CLI, etc.

Today, this templates look like this:
Screen Shot 2021-10-04 at 11 00 15 AM

I have not tested it 😂 since we need to merge first to see if Github renders it correctly 🤞🏽

## Issue

N/A

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,38 @@
----
-name: Pull request template
-about: 'type(scope): Subject of the pull request '
----
+<!--
+  Thanks for submitting a pull request!
+  We appreciate you spending the time to work on these changes.
+  Please provide enough information so that others can review your pull request.
 
-***Issue***: Include link to the Jira/Github Issue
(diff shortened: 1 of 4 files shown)


.github/pull_request_template.md
```diff
@@ -1,12 +1,38 @@
----
-name: Pull request template
-about: 'type(scope): Subject of the pull request '
----
+<!--
```
dmurray-lacework, line 1: Are you sure you want to include this in the template and not in the contributing/developer guidelines?

afiune: Yes. We should also update the developer guidelines but regardless, this is the place we know with certitude our users will read, more detailed information could be added to other docs and we should link them to those docs but being honest, how many times have you (personally) follow those links?

dmurray-lacework (approved):

PR #570, opened 2021-10-05
Vatasha/ally 639/migrate dd alert channel api v2 fix

vatasha:
## Summary

We're currently migrating alert channels to v2 to be able to support alert channels at the org level. 

There are updated tests

## Issue

https://lacework.atlassian.net/browse/ALLY-639?atlOrigin=eyJpIjoiMjM4NmRkMzFmNzNlNDMwMGFhYmM2N2YwZjg0ZjFkYWYiLCJwIjoiaiJ9


afiune (approved):
Please, squash and merge this PR with a good message.

api/alert_channel_datadog.go
```diff
@@ -0,0 +1,47 @@
...
+// UpdateDatadog updates a single instance of a Datadog integration on the Lacework server
+func (svc *AlertChannelsService) UpdateDatadog(data AlertChannel) (response DatadogAlertChannelResponseV2, err error) {
+	err = svc.update(data.ID(), data, &response)
+	return
+}
+
+type DatadogDataV2 struct {
+	APIKey      string `json:"apiKey"`
+	DatadogSite string `json:"datadogSite,omitempty"`
+	DatadogType string `json:"datadogType,omitempty"`
```
dmurray-lacework, line 37: Is omitempty needed here? Schema says they are required fields

api/alert_channel_datadog.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// UpdateDatadog updates a single instance of a Datadog integration on the Lacework server
+func (svc *AlertChannelsService) UpdateDatadog(data AlertChannel) (response DatadogAlertChannelResponseV2, err error) {
+	err = svc.update(data.ID(), data, &response)
+	return
+}
+
+type DatadogDataV2 struct {
+	APIKey      string `json:"apiKey"`
```
dmurray-lacework, line 35: Elsewhere in the go-sdk do we use "API" OR "Api"?

dmurray-lacework (changes requested):
We should also include the CLI Create integration changes in these migrations?

cli/cmd/integration_datadog.go:83
cli/cmd/integration_datadog.go:92

api/alert_channel_datadog.go
vatasha: There is a default value which is why I have omitempty

PR #575, opened 2021-10-07
ci: fix downgrading of go packages

vatasha:
## Summary

Fix the downgrading of go packages by using `go install`

Ran manually and verified that `make go-vendor` was properly verified. 

## Issue
N/A


afiune (approved):
tenor-172800820

PR #583, opened 2021-10-14
feat(api): Alert Rules v2 service

dmurray-lacework:

## Summary

New Alert Rules service added to go-sdk .

New unit tests
`/api/alert_rules_test.go`

Run example in 
`api/_examples/alert-rules/main.go`

## Issue

https://lacework.atlassian.net/browse/ALLY-680

### DIFF
--- a/api/alert_rules.go
+++ b/api/alert_rules.go
@@ -34,7 +34,7 @@ type alertRuleSeverity int
 
 type AlertRuleSeverities []alertRuleSeverity
 
-const alertRuleEventType = "Event"
+const AlertRuleEventType = "Event"
 
 func (sevs AlertRuleSeverities) toInt() []int {
 	var res []int
@@ -52,9 +52,9 @@ const (
 	AlertRuleSeverityInfo     alertRuleSeverity = 5
 )
 
-// NewAlertRule returns an instance of the AlertRuleData struct
+// NewAlertRule returns an instance of the AlertRule struct
 //
-// Basic usage: Initialize a new AlertRuleData struct, then
+// Basic usage: Initialize a new AlertRule struct, then
 //              use the new instance to do CRUD operations
 //
 //   client, err := api.NewClient("account")
@@ -64,7 +64,7 @@ const (
 //
 //   alertRule := api.NewAlertRule(
 //		"Foo",
-//		api.AlertRule{
+//		api.AlertRuleConfig{
 //		Description: "My Alert Rule"
 //		Severities: api.AlertRuleSeverities{api.AlertRuleSeverityHigh,
 //		Channels: []string{"TECHALLY_000000000000AAAAAAAAAAAAAAAAAAAA"},
@@ -75,10 +75,10 @@ const (
 //
 //   client.V2.AlertRules.Create(alertRule)
 //
-func NewAlertRule(name string, rule AlertRule) AlertRuleData {
-	return AlertRuleData{
+func NewAlertRule(name string, rule AlertRuleConfig) AlertRule {
+	return AlertRule{
 		Channels: rule.Channels,
-		Type:     alertRuleEventType,
+		Type:     AlertRuleEventType,
 		Filter: AlertRuleFilter{
 			Name:            name,
 			Enabled:         1,
@@ -97,7 +97,7 @@ func (svc *AlertRulesService) List() (response AlertRulesResponse, err error) {
 }
 
 // Create creates a single Alert Rule
-func (svc *AlertRulesService) Create(rule AlertRuleData) (
+func (svc *AlertRulesService) Create(rule AlertRule) (
 	response AlertRuleResponse,
 	err error,
 ) {
@@ -120,7 +120,7 @@ func (svc *AlertRulesService) Delete(guid string) error {
 }
 
 // Update updates a single Alert Rule of the provided guid.
-func (svc *AlertRulesService) Update(data AlertRuleData) (
+func (svc *AlertRulesService) Update(data AlertRule) (
 	response AlertRuleResponse,
 	err error,
 ) {
@@ -142,7 +142,7 @@ func (svc *AlertRulesService) Get(guid string, response interface{}) error {
 	return svc.client.RequestDecoder("GET", apiPath, nil, &response)
 }
 
-type AlertRule struct {
+type AlertRuleConfig struct {
 	Channels        []string
 	Description     string
 	Severities      AlertRuleSeverities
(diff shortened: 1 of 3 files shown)


api/alert_rules.go
```diff
@@ -0,0 +1,201 @@
...
+	client *Client
+}
+
+type alertRuleType string
+
+type alertRuleSeverity int
+
+type AlertRuleSeverities []alertRuleSeverity
+
+// Alert Rule type is an enum in api v2, with only 1 supported value "Event"
```
afiune, line 39: why? are we going to have more? 🤔 This seems like a lot of overhead for a hardcoded type.

api/alert_rules.go
```diff
@@ -0,0 +1,201 @@
...
+//              use the new instance to do CRUD operations
+//
+//   client, err := api.NewClient("account")
+//   if err != nil {
+//     return err
+//   }
+//
+//   alertRule := api.NewAlertRule(
+//		"Foo",
+//     api.AlertRule{
```
afiune, line 75: nit: indentation

api/alert_rules.go
```diff
@@ -0,0 +1,201 @@
...
+
+type AlertRuleData struct {
+	Guid     string          `json:"mcGuid,omitempty"`
+	Type     string          `json:"type"`
+	Channels []string        `json:"intgGuidList"`
+	Filter   AlertRuleFilter `json:"filters"`
+}
+
+type AlertRuleFilter struct {
+	AlertRuleComputedData
```
afiune, line 161: What is the benefit of this extra struct?

api/alert_rules.go
```diff
@@ -0,0 +1,201 @@
...
+	err = svc.client.RequestDecoder("GET", apiV2AlertRules, nil, &response)
+	return
+}
+
+// Create creates a single Alert Rule
+func (svc *AlertRulesService) Create(rule AlertRuleData) (
+	response AlertRuleResponse,
+	err error,
+) {
+	err = svc.create(rule, &response)
```
afiune, line 113: we created private functions when we had multiple types and every type was calling the underlying service function, this is a unique type so, having separate functions for `create()` `update()` etc. feels like a copy-paste that is not needed for this Alert Rule service.

Can we just do the API call directly here and delete the other private functions?
```suggestion
	err = svc.client.RequestEncoderDecoder("POST", apiV2AlertRules, &rule, &response)
```
Something like that 👆🏽

api/alert_rules.go
```diff
@@ -0,0 +1,201 @@
...
+		nil,
+	)
+}
+
+// Update updates a single Alert Rule of the provided guid.
+func (svc *AlertRulesService) Update(data AlertRuleData) (
+	response AlertRuleResponse,
+	err error,
+) {
+	err = svc.update(data.Guid, data, &response)
```
afiune, line 136: same here, no need to have a private `update()` function.

api/alert_rules.go
```diff
@@ -0,0 +1,201 @@
...
+	return
+}
+
+// Get returns a raw response of the Alert Rule with the matching guid.
+func (svc *AlertRulesService) Get(guid string, response interface{}) error {
+	return svc.get(guid, &response)
```
afiune, line 142: same. no need for `get()`

api/alert_rules.go
```diff
@@ -0,0 +1,201 @@
...
+}
+
+type AlertRule struct {
+	Channels        []string
+	Description     string
+	Severities      AlertRuleSeverities
+	ResourceGroups  []string
+	EventCategories []string
+}
```
afiune, line 151: I like this struct 💯 - I just feel it might be confusing between `AlertRuleData` and `AlertRule`. But I like it since the user can customize it directly and pass it to `NewAlertRule()`

api/schemas.go
```diff
@@ -29,6 +29,7 @@ type integrationSchema int
 const (
 	None integrationSchema = iota
 	AlertChannels
+	AlertRules
```
afiune, line 32: can we inject constants like this? Im concerned we are going to change the number from the `iota` and something could get messed up 🤔

api/alert_rules.go
```diff
@@ -0,0 +1,201 @@
...
+func (sevs AlertRuleSeverities) toInt() []int {
+	var res []int
+	for _, i := range sevs {
+		res = append(res, int(i))
+	}
+	return res
+}
+
+const (
+	AlertRuleSeverityCritical alertRuleSeverity = iota + 1
```
afiune, line 56: Even though I like you used `iota` here, I don't think it is a good idea. What if someone re-orders this enum or, add something in between severities, just like you did below in the schema service.

At least put a comment that says that the enums MUST be 1,2,3,4,5 or, you could just hardcode them. (hardcoding them would be my suggestion)

afiune (commented):
Looking preeeetty goog! 👍🏽

api/schemas.go
dmurray-lacework: We don't make use the numerical value, we're really just trying to create an enumerated type for supported services.

api/alert_rules.go
dmurray-lacework: None, I've removed it.

api/alert_rules.go
dmurray-lacework: I've left the name as is for now.

api/alert_rules.go
```diff
@@ -0,0 +1,177 @@
...
+//		Severities: api.AlertRuleSeverities{api.AlertRuleSeverityHigh,
+//		Channels: []string{"TECHALLY_000000000000AAAAAAAAAAAAAAAAAAAA"},
+//		ResourceGroups: []string{"TECHALLY_111111111111AAAAAAAAAAAAAAAAAAAA"}
+//       },
+//     },
+//   )
+//
+//   client.V2.AlertRules.Create(alertRule)
+//
+func NewAlertRule(name string, rule AlertRule) AlertRuleData {
```
afiune, line 78: I feel that the `NewAlertRule()` function should return the struct `AlertRule` so that means that we can rename the current `AlertRule` struct which acts like parameters of the function as:
* `AlertRuleConfig`
* `AlertRuleParams`
* `AlertRuleParameters`
* `AlertRuleDetails`
* `AlertRuleInfo`

NOTE: I like the first two. cc @vatasha thoughts? 

```suggestion
func NewAlertRule(name string, rule AlertRuleConfig) AlertRule {
```

afiune (approved):
Great work, just one comment to name the structs differently. Besides that, ship it!
tenor-48817522

api/alert_rules.go
dmurray-lacework: I like it! `AlertRuleConfig` would be my preference

PR #604, opened 2021-11-08
ci: Update go conditional build comment to 1.17 preferred format

dmurray-lacework:
## Summary

Fix formatting changes from go1.17

Make fmt-check in ci passes

## Issue

https://lacework.atlassian.net/browse/ALLY-657


afiune (approved):
Is there a way we can move the Codefresh YAML to be inside the repository? It would be easier to run these changes with both, the actual code change and the pipeline change together.

PR #614, opened 2021-11-16
feat(api): implement crud operations for team members

vatasha:
## Summary
Implement team members. 

There are some new tests. 

## Issue
https://lacework.atlassian.net/browse/ALLY-736


api/team_members.go
```diff
@@ -0,0 +1,139 @@
...
+	return svc.client.RequestDecoder("DELETE", fmt.Sprintf(apiV2TeamMembersFromGUID, guid), nil, nil)
+}
+
+// Update updates a single team member with the corresponding guid
+func (svc *TeamMembersService) Update(tm TeamMember) (res TeamMemberResponse, err error) {
+	if tm.CustGuid == "" {
+		err = errors.New("please specify a guid")
+		return
+	}
+	err = svc.client.RequestEncoderDecoder("PATCH", fmt.Sprintf(apiV2TeamMembersFromGUID, tm.UserGuid), tm, &res)
```
afiune, line 97: For updates, verify that you are able to send the `UserGuid` as part of the request payload, some endpoints complain about it since you are already injecting it via URL parameter.

afiune (commented):
Looking pretty good!

tenor-5332604

api/team_members.go
```diff
@@ -0,0 +1,139 @@
...
+// Create creates a single team member
+func (svc *TeamMembersService) Create(tm TeamMember) (res TeamMemberResponse, err error) {
+	err = svc.client.RequestEncoderDecoder("POST", apiV2TeamMembers, tm, &res)
+	return
+}
+
+// Delete deletes a single team member with the corresponding guid
+func (svc *TeamMembersService) Delete(guid string) error {
+	if guid == "" {
+		errors.New("please specify a guid")
```
afiune, line 85:
```suggestion
		return errors.New("please specify a guid")
```

api/team_members.go
```diff
@@ -0,0 +1,128 @@
...
+	return svc.client.RequestDecoder("GET", fmt.Sprintf(apiV2TeamMembersFromGUID, guid), nil, &res)
+
+}
+
+type TeamMemberProps struct {
+	AccountAdmin           bool   `json:"accountAdmin,omitempty"`
+	Company                string `json:"company"`
+	CreatedTime            string `json:"createdTime,omitempty"`
+	FirstName              string `json:"firstName"`
+	JitCreated             bool   `json:"jitCreated,omitempty"`
```
vatasha, line 104: I'm worried about a couple fields here that are strings but the API returns integers. For example here is the API response for getting a single team member. `{
  "data": {
    "custGuid": "<redacted>",
    "props": {
      "accountAdmin": true,
      "company": "CUSTOMERDEMO",
      "createdTime": "2020-04-09T17:39:51.572Z",
      "firstName": "<redacted>",
      "jitCreated": false,
      "lastName": "<redacted>",
      "lastSessionCreatedTime": 0,
      "orgAdmin": true,
      "orgUser": false,
      "updatedTime": 0
    },
    "userEnabled": 1,
    "userGuid": "<redacted>",
    "userName": "<redacted>"
  }
}
`

How should we handle the API returning ints for `lastSessionCreatedTime` and `updatedTime`? I suspect that we're going to have issues deserializing.

dmurray-lacework: It seems when something hasn't been updated updatedTime/lastSessionCreatedTime returns as 0<int> but when populated it returns as string time stamp.  "updatedTime": "2021-02-25T15:25:01.691Z".

This may not actually cause any issues. int value will fail to deserialize, but updatedTime/lastSessionCreatedTime would be empty strings anyway. But would be good to test this to confirm.

api/team_members_test.go
```diff
@@ -0,0 +1,340 @@
...
+            "company": "Lacework",
+            "createdTime": "2021-11-16T16:33:17.573Z",
+            "firstName": "Vatasha",
+            "jitCreated": %t,
+            "lastLoginTime": "0",
+            "lastName": "White",
+			"lastSessionCreatedTime": "0",
+			"orgAdmin": false,
+      		"orgUser": false,
+      		"updatedTime": "0"
```
afiune, line 333: from what @dmurray-lacework already mentioned, what happens if this field is 0?
```
      		"updatedTime": 0
```

afiune (approved):
tenor-172213508

PR #619, opened 2021-11-23
feat(cli): logic to store assets with expiration time

afiune:
## Summary

Introducing a simple private struct that acks as an envelope for storing
assets that has an expiration time into the local cache named `cliAsset`.
The `Data` field is an interface on purpose to allow developers to store
any kind of asset, from primitives such as strings, ints, bools, to JSON
objects.

The new function called `WriteAssetToCache` stores an asset with an
expiration time.

**Simple Example:** Having a struct named `vulnReport`

```go
cli.WriteAssetToCache("my-report", time.Now().Add(time.Hour * 1), vulnReport{Foo: "bar"})
```

The funciton `ReadCachedAsset` tries to reads an asset with an expiration
time, if the asset has expired, it returns `true`, otherwise it returns
`false`.

**Simple Example:** Having a struct named `vulnReport`

```go
var report vulnReport
if expired := cli.ReadCachedAsset("my-report", &report); !expired {
    fmt.Printf("My report: %v\n", report)
}
```

Tests were added.

## Issue

https://lacework.atlassian.net/browse/ALLY-773 - **Research and development.**

### DIFF
--- a/cli/cmd/cache.go
+++ b/cli/cmd/cache.go
@@ -22,6 +22,7 @@ import (
 	"encoding/json"
 	"path"
 	"strings"
+	"time"
 
 	"github.com/lacework/go-sdk/internal/format"
 	homedir "github.com/mitchellh/go-homedir"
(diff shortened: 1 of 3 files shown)


ipcrm (approved):
LGTM!

ipcrm:
Examples:

Pasted_Image_11_23_21__5_50_PM
Pasted_Image_11_23_21__5_51_PM

ipcrm (approved):

PR #624, opened 2021-11-29
feat(cli): cache compliance reports for 30 minutes

afiune:
## Summary

When a user is trying to visualize a compliance report via the CLI, we
shouldn’t ask the APIs for the same report on every command execution,
especially not since the report won’t change between two commands run
within a few minutes. We should be able to use the Caching mechanism
inside the Lacework CLI to cache compliance reports and let users
navigate them faster and reduce the load to our API server.

We have already done this for host vulnerability assessments.

I want to inspect compliance reports via the CLI multiple times without getting to API rate limits,
So I can create workflows internally and navigate my reports with additional flags and filters

>As a Lacework engineer,
I want tools and integrations to cache data as much as possible without compromising consistency,
So other services, such as the API server, don't get overwhelmed with the same API requests that are immutable

Ran a few commands in sequence against our test account.

Screen Shot 2021-11-29 at 9 39 58 AM

## Issue
https://lacework.atlassian.net/browse/ALLY-770

### DIFF
--- a/cli/cmd/compliance_azure.go
+++ b/cli/cmd/compliance_azure.go
@@ -204,7 +204,7 @@ To run an ad-hoc compliance assessment use the command:
 
 			var (
 				report   api.ComplianceAzureReport
-				cacheKey = fmt.Sprintf("compliance/aws/%s/%s/%s",
+				cacheKey = fmt.Sprintf("compliance/azure/%s/%s/%s",
 					config.TenantID, config.SubscriptionID, config.Type)
 			)
 			expired := cli.ReadCachedAsset(cacheKey, &report)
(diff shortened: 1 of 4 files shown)


dmurray-lacework (approved):

cli/cmd/compliance_azure.go
```diff
@@ -202,18 +202,29 @@ To run an ad-hoc compliance assessment use the command:
...
 
-			cli.StartProgress(" Getting compliance report...")
-			response, err := cli.LwApi.Compliance.GetAzureReport(config)
-			cli.StopProgress()
-			if err != nil {
-				return errors.Wrap(err, "unable to get azure compliance report")
-			}
+			var (
+				report   api.ComplianceAzureReport
+				cacheKey = fmt.Sprintf("compliance/aws/%s/%s/%s",
```
afiune, line 207: @dmurray-lacework crap, this is wrong 😢 , this should be `compliance/gcp/` or `compliance/google`

PR #625, opened 2021-11-30
feat(cli): Manage Report Rules

dmurray-lacework:
## Summary

Manage Report Rules in the Lacework cli

Manually run `lacework report-rule create | list | show | delete`

## Issue

https://lacework.atlassian.net/browse/ALLY-765

### DIFF
--- a/cli/cmd/report_rules.go
+++ b/cli/cmd/report_rules.go
@@ -1,0 +1,410 @@
+//
+// Author:: Darren Murray(<darren.murray@lacework.net>)
+// Copyright:: Copyright 2021, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package cmd
+
+import (
+	"fmt"
+	"sort"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/fatih/structs"
+	"github.com/olekukonko/tablewriter"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/lacework/go-sdk/api"
+)
+
+var (
+	// report-rules command is used to manage lacework report rules
+	reportRulesCommand = &cobra.Command{
+		Use:     "report-rule",
+		Aliases: []string{"report-rules", "rr"},
+		Short:   "Manage report rules",
+		Long: `Manage report rules to route reports to one or more email alert channels.		
+
+A report rule has four parts:
+
+  1. Email alert channel(s) that should receive the report
+  2. One or more severities to include
+  3. Resource group(s) containing the subset of your environment to consider
+  4. Notification types containing which report information to send
+`,
+	}
+
+	// list command is used to list all lacework report rules
+	reportRulesListCommand = &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "List all report rules",
+		Long:    "List all report rules configured in your Lacework account.",
+		Args:    cobra.NoArgs,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			cli.StartProgress(" Fetching report rules...")
+			reportRules, err := cli.LwApi.V2.ReportRules.List()
+			cli.StopProgress()
+			if err != nil {
+				return errors.Wrap(err, "unable to get report rules")
+			}
+			if len(reportRules.Data) == 0 {
+				msg := `There are no report rules configured in your account.
+
+Get started by configuring your report rules using the command:
+
+    lacework report-rule create
+
+If you prefer to configure report rules via the WebUI, log in to your account at:
+
+    https://%s.lacework.net
+
+Then navigate to Settings > Report Rules.
+`
+				cli.OutputHuman(fmt.Sprintf(msg, cli.Account))
+				return nil
+			}
+			if cli.JSONOutput() {
+				return cli.OutputJSON(reportRules)
+			}
+
+			var rows [][]string
+			for _, rule := range reportRules.Data {
+				rows = append(rows, []string{rule.Guid, rule.Filter.Name, rule.Filter.Status()})
+			}
+
+			cli.OutputHuman(renderSimpleTable([]string{"GUID", "NAME", "ENABLED"}, rows))
+			return nil
+		},
+	}
+	// show command is used to retrieve a lacework report rule by guid
+	reportRulesShowCommand = &cobra.Command{
+		Use:   "show",
+		Short: "Show a report rule by id",
+		Long:  "Show a single report rule by it's id.",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(_ *cobra.Command, args []string) error {
+			var response api.ReportRuleResponse
+			cli.StartProgress(" Fetching report rule...")
+
+			err := cli.LwApi.V2.ReportRules.Get(args[0], &response)
+			if err != nil {
+				cli.StopProgress()
+				return errors.Wrap(err, "unable to get report rule")
+			}
+			cli.StopProgress()
+
+			if cli.JSONOutput() {
+				return cli.OutputJSON(response)
+			}
+
+			reportRule := response.Data
+			headers := [][]string{
+				[]string{reportRule.Guid, reportRule.Filter.Name, reportRule.Filter.Status()},
+			}
+
+			cli.OutputHuman(renderSimpleTable([]string{"GUID", "NAME", "ENABLED"}, headers))
+			cli.OutputHuman("\n")
+			cli.OutputHuman(buildReportRuleDetailsTable(reportRule))
+
+			return nil
+		},
+	}
+
+	// delete command is used to remove a lacework report rule by id
+	reportRulesDeleteCommand = &cobra.Command{
+		Use:   "delete",
+		Short: "Delete a report rule",
+		Long:  "Delete a single report rule by it's ID.",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(_ *cobra.Command, args []string) error {
+			cli.StartProgress(" Deleting report rule...")
+			err := cli.LwApi.V2.ReportRules.Delete(args[0])
+			cli.StopProgress()
+			if err != nil {
+				return errors.Wrap(err, "unable to delete report rule")
+			}
+			cli.OutputHuman("The report rule with GUID %s was deleted\n", args[0])
+			return nil
+		},
+	}
+
+	// create command is used to create a new lacework report rule
+	reportRulesCreateCommand = &cobra.Command{
+		Use:   "create",
+		Short: "Create a new report rule",
+		Args:  cobra.NoArgs,
+		RunE: func(_ *cobra.Command, args []string) error {
+			if !cli.InteractiveMode() {
+				return errors.New("interactive mode is disabled")
+			}
+
+			response, err := promptCreateReportRule()
+			if err != nil {
+				return errors.Wrap(err, "unable to create report rule")
+			}
+
+			cli.OutputHuman("The report rule was created with GUID %s\n", response.Data.Guid)
+			return nil
+		},
+	}
+)
+
+func init() {
+	// add the report-rule command
+	rootCmd.AddCommand(reportRulesCommand)
+
+	// add sub-commands to the report-rule command
+	reportRulesCommand.AddCommand(reportRulesListCommand)
+	reportRulesCommand.AddCommand(reportRulesShowCommand)
+	reportRulesCommand.AddCommand(reportRulesCreateCommand)
+	reportRulesCommand.AddCommand(reportRulesDeleteCommand)
+}
+
+func buildReportRuleDetailsTable(rule api.ReportRule) string {
+	var (
+		details       [][]string
+		notifications [][]string
+		updatedTime   string
+	)
+	severities := api.NewReportRuleSeveritiesFromIntSlice(rule.Filter.Severity).ToStringSlice()
+
+	if nano, err := strconv.ParseInt(rule.Filter.CreatedOrUpdatedTime, 10, 64); err == nil {
+		updatedTime = time.Unix(nano/1000, 0).Format(time.RFC3339)
+	}
+	details = append(details, []string{"SEVERITIES", strings.Join(severities, ", ")})
+	details = append(details, []string{"DESCRIPTION", rule.Filter.Description})
+	details = append(details, []string{"UPDATED BY", rule.Filter.CreatedOrUpdatedBy})
+	details = append(details, []string{"LAST UPDATED", updatedTime})
+
+	detailsTable := &strings.Builder{}
+	detailsTable.WriteString(renderOneLineCustomTable("ALERT RULE DETAILS",
+		renderCustomTable([]string{}, details,
+			tableFunc(func(t *tablewriter.Table) {
+				t.SetBorder(false)
+				t.SetColumnSeparator(" ")
+				t.SetAutoWrapText(false)
+				t.SetAlignment(tablewriter.ALIGN_LEFT)
+			}),
+		),
+		tableFunc(func(t *tablewriter.Table) {
+			t.SetBorder(false)
+			t.SetAutoWrapText(false)
+		}),
+	),
+	)
+
+	notifcationsMap := rule.ReportNotificationTypes.ToMap()
+	// sort keys
+	keys := make([]string, 0, len(notifcationsMap))
+	for k := range notifcationsMap {
+		keys = append(keys, k)
+	}
+	sort.Strings(keys)
+
+	for _, key := range keys {
+		enabled := strings.Title(strconv.FormatBool(notifcationsMap[key]))
+		notifications = append(notifications, []string{key, enabled})
+	}
+
+	detailsTable.WriteString(renderCustomTable([]string{"NOTIFICATION TYPES", "ENABLED"}, notifications,
+		tableFunc(func(t *tablewriter.Table) {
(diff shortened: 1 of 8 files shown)


cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+
+var (
+	// report-rules command is used to manage lacework report rules
+	reportRulesCommand = &cobra.Command{
+		Use:     "report-rule",
+		Aliases: []string{"report-rules", "rr"},
+		Short:   "Manage report rules",
+		Long: `Manage report rules to route reports to one or more email alert channels.		
+
+An report rule has Four parts:
```
afiune, line 45:
```suggestion
A report rule has four parts:
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+
+  1. Email Alert channel(s) that should receive the report
```
afiune, line 47:
```suggestion
  1. Email alert channel(s) that should receive the report
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+  2. Severities to include
```
afiune, line 48:
```suggestion
  2. One or more severities to include
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+		Args:    cobra.NoArgs,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			reportRules, err := cli.LwApi.V2.ReportRules.List()
+			if err != nil {
+				return errors.Wrap(err, "unable to get report rules")
+			}
+			if len(reportRules.Data) == 0 {
+				msg := `There are no report rules configured in your account.
+
+Get started by integrating your report rules to manage reporting using the command:
```
afiune, line 69:
```suggestion
Get started by creating your report rules using the command:
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+			var rows [][]string
+			for _, rule := range reportRules.Data {
+				rows = append(rows, []string{rule.Guid, rule.Filter.Name, rule.Filter.Status()})
+			}
+
+			cli.OutputHuman(renderSimpleTable([]string{"GUID", "NAME", "ENABLED"}, rows))
+			return nil
+		},
+	}
+	// show command is used to retrieve a lacework report rule by resource id
```
afiune, line 95:
```suggestion
	// show command is used to retrieve a lacework report rule by guid
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+
+			cli.OutputHuman(renderSimpleTable([]string{"GUID", "NAME", "ENABLED"}, headers))
+			cli.OutputHuman("\n")
+			cli.OutputHuman(buildReportRuleDetailsTable(reportRule))
+
+			return nil
+		},
+	}
+
+	// delete command is used to remove a lacework report rule by resource id
```
afiune, line 124: either guid or id
```suggestion
	// delete command is used to remove a lacework report rule by id
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+			}
+
+			cli.OutputHuman(renderSimpleTable([]string{"GUID", "NAME", "ENABLED"}, rows))
+			return nil
+		},
+	}
+	// show command is used to retrieve a lacework report rule by resource id
+	reportRulesShowCommand = &cobra.Command{
+		Use:   "show",
+		Short: "Show an report rule by id",
```
afiune, line 98:
```suggestion
		Short: "Show a report rule by id",
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+		Long:  "Show a single report rule by it's ID.",
```
afiune, line 99: either `ID` or `id` but be consistent on all other commands
```suggestion
		Long:  "Show a single report rule by it's id.",
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+				return errors.Wrap(err, "unable to get report rule")
+			}
+
+			if cli.JSONOutput() {
+				return cli.OutputJSON(response)
+			}
+
+			reportRule := response.Data
+			var headers [][]string
+			headers = append(headers, []string{reportRule.Guid, reportRule.Filter.Name, reportRule.Filter.Status()})
```
afiune, line 114: nit
```suggestion
			headers := [][]string{
				[]string{reportRule.Guid, reportRule.Filter.Name, reportRule.Filter.Status()}
			}
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+			}
+			cli.OutputHuman(fmt.Sprintf("The report rule with GUID %s was deleted \n", args[0]))
+			return nil
+		},
+	}
+
+	// create command is used to create a new lacework report rule
+	reportRulesCreateCommand = &cobra.Command{
+		Use:   "create",
+		Short: "Create a new report rule",
```
afiune, line 143:
```suggestion
		Short: "Create a new report rule",
		Args:    cobra.NoArgs,
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+		Use:   "delete",
+		Short: "Delete a report rule",
+		Long:  "Delete a single report rule by it's ID.",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(_ *cobra.Command, args []string) error {
+			err := cli.LwApi.V2.ReportRules.Delete(args[0])
+			if err != nil {
+				return errors.Wrap(err, "unable to delete report rule")
+			}
+			cli.OutputHuman(fmt.Sprintf("The report rule with GUID %s was deleted \n", args[0]))
```
afiune, line 135:
```suggestion
			cli.OutputHuman(fmt.Sprintf("The report rule with GUID %s was deleted\n", args[0]))
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+			if !cli.InteractiveMode() {
+				return errors.New("interactive mode is disabled")
+			}
+
+			response, err := promptCreateReportRule()
+			if err != nil {
+				return errors.Wrap(err, "unable to create report rule")
+			}
+
+			cli.OutputHuman(fmt.Sprintf("The report rule was created with GUID %s \n", response.Data.Guid))
```
afiune, line 154:
```suggestion
			cli.OutputHuman(fmt.Sprintf("The report rule was created with GUID %s\n", response.Data.Guid))
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+			ResourceGroups:     groups,
+			EmailAlertChannels: channels,
+			NotificationTypes:  api.ReportRuleNotifications{notifications},
+		})
+
+	if err != nil {
+		return api.ReportRuleResponse{}, err
+	}
+
+	cli.StartProgress(" Creating report rule...")
```
afiune, line 348: I really like we use the progress widget, can we please add it to other commands? (`show`, `delete`, and `list`)

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+	response, err := cli.LwApi.V2.ReportRules.Create(reportRule)
+
+	cli.StopProgress()
+	return response, err
```
afiune, line 352:
```suggestion
	defer cli.StopProgress()

	return cli.LwApi.V2.ReportRules.Create(reportRule)
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+
+	return detailsTable.String()
+}
+
+func promptCreateReportRule() (api.ReportRuleResponse, error) {
+	channelList, channelMap := getEmailAlertChannels()
+	notificationFields := structs.Names(api.ReportRuleNotificationTypes{})
+	notificationsMap := make(map[string]bool)
+
+	if len(channelList) < 1 {
```
afiune, line 258: I am worried that `getEmailAlertChannels()` can return `nil, nil` in which case this `len(nil)` will panic.

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+	err := survey.Ask(questions, &answers,
+		survey.WithIcons(promptIconsFunc),
+	)
+	if err != nil {
+		return api.ReportRuleResponse{}, err
+	}
+
+	var channels []string
+	for _, channel := range answers.Channels {
+		channels = append(channels, channelMap[channel])
```
afiune, line 315: same here, `channelMap` could be `nil` 😟

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+	}
+
+	resourceGroups, resourceGroupMap := promptAddResourceGroupsToReportRule()
```
afiune, line 318: same here, if functions "can" return `nil` values, you will have to check them. One way to fix this is by always returning an empty array or something other than `nil`

afiune (changes requested):

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,399 @@
...
+	"github.com/lacework/go-sdk/api"
+)
+
+var (
+	// report-rules command is used to manage lacework report rules
+	reportRulesCommand = &cobra.Command{
+		Use:     "report-rule",
+		Aliases: []string{"report-rules", "rr"},
+		Short:   "Manage report rules",
+		Long: `Manage report rules to route reports to one or more email alert channels.
```
vatasha, line 43: Can report rules only be routed to email alert channels?

dmurray-lacework: Yes, they're only available to email alert channels

afiune: Yes.

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,400 @@
...
+			cli.OutputHuman(buildReportRuleDetailsTable(reportRule))
+
+			return nil
+		},
+	}
+
+	// delete command is used to remove a lacework report rule by id
+	reportRulesDeleteCommand = &cobra.Command{
+		Use:   "delete",
+		Short: "Delete a report rule",
```
vatasha, line 128: Do we want to introduce a `del` alias as it's short for delete.

afiune: I thought about `rm` once 🤔 - not sure about `del` 

Let us talk about this in standup, if we all agree, we could do it, but if we do, we need to add this alias for ALL `delete` commands.

cli/cmd/report_rules.go
dmurray-lacework: The len of a nil slice is 0

dmurray-lacework: Tested this to confirm, when nil we enter the if condition block and return err

cli/cmd/report_rules.go
dmurray-lacework: The channelList is used to populate the survey. Which is a required field, if the channelList is empty this line is not reachable

cli/cmd/report_rules.go
dmurray-lacework: If we'd like to add more aliases we should create a ticket to standardize across commands. We use 'bash' syntactical commands elsewhere like 'ls' for aliases so 'rm' would fit this pattern

cli/cmd/report_rules.go
dmurray-lacework: A range of nil is 0. In this scenario there is no issue. But I agree with the sentiment of nil checking. But here we're not trying to access a nil value

cli/cmd/report_rules.go
dmurray-lacework: Done

cli/cmd/report_rules.go
vatasha: That's a good point, I like rm.

vatasha (approved):
LGTM 🎉

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,409 @@
...
+	// list command is used to list all lacework report rules
+	reportRulesListCommand = &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "List all report rules",
+		Long:    "List all report rules configured in your Lacework account.",
+		Args:    cobra.NoArgs,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			cli.StartProgress(" Fetching report rules...")
+			defer cli.StopProgress()
```
afiune, line 63: I am not sure you can defer this function because maybe it will conflict with `OutputHuman()`

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,409 @@
...
+	// show command is used to retrieve a lacework report rule by guid
+	reportRulesShowCommand = &cobra.Command{
+		Use:   "show",
+		Short: "Show a report rule by id",
+		Long:  "Show a single report rule by it's id.",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(_ *cobra.Command, args []string) error {
+			var response api.ReportRuleResponse
+			cli.StartProgress(" Fetching report rule...")
+			defer cli.StopProgress()
```
afiune, line 107: same here...

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,409 @@
...
+			if !cli.InteractiveMode() {
+				return errors.New("interactive mode is disabled")
+			}
+
+			response, err := promptCreateReportRule()
+			if err != nil {
+				return errors.Wrap(err, "unable to create report rule")
+			}
+
+			cli.OutputHuman(fmt.Sprintf("The report rule was created with GUID %s\n", response.Data.Guid))
```
afiune, line 165: You don't need to format the string, the underlying function does it automatically.
```suggestion
			cli.OutputHuman("The report rule was created with GUID %s\n", response.Data.Guid)
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,409 @@
...
+	}
+
+	cli.StartProgress(" Creating report rule...")
+	defer cli.StopProgress()
+
+	return cli.LwApi.V2.ReportRules.Create(reportRule)
+}
+
+func getEmailAlertChannels() ([]string, map[string]string) {
+	response, err := cli.LwApi.V2.AlertChannels.List()
```
afiune, line 366: You could add a Start and Stop Progress here with a defer. Not sure how fast we can return Alert Channels.

cli/cmd/report_rules.go
afiune: ok

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,409 @@
...
+		Message: "Add Resource Groups to Report Rule?",
+	}, &addResourceGroups)
+
+	if err != nil {
+		return nil, nil
+	}
+
+	if addResourceGroups {
+		var groups []string
+		groupList, groupMap := getResourceGroups()
```
afiune, line 396: would be nice to also add a progress inside this function, it doesn't have it

afiune (approved):
A few things to consider but the code looks good. Great work!

cli/cmd/report_rules.go
dmurray-lacework: 👍  Yeh, will just have to call stopProgress twice inside the error and after success

cli/cmd/report_rules.go
dmurray-lacework: I just tested this and there's something not quite righ about it. As we fetch the email alert channels prior to the survey starting. It didn't seem right to have a loader for email alert channels but we don't prompt the user to use them til later in the survey.

We could put in the loader without any message though.

cli/cmd/report_rules.go
dmurray-lacework: 👍

cli/cmd/report_rules.go
dmurray-lacework: Done

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,412 @@
...
+		Long:    "List all report rules configured in your Lacework account.",
+		Args:    cobra.NoArgs,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			cli.StartProgress(" Fetching report rules...")
+			reportRules, err := cli.LwApi.V2.ReportRules.List()
+			if err != nil {
+				cli.StopProgress()
+				return errors.Wrap(err, "unable to get report rules")
+			}
+			cli.StopProgress()
```
afiune, line 68:
```suggestion
			cli.StartProgress(" Fetching report rules...")
			reportRules, err := cli.LwApi.V2.ReportRules.List()
			cli.StopProgress()
			if err != nil {
				return errors.Wrap(err, "unable to get report rules")
			}
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,412 @@
...
+		Args:  cobra.ExactArgs(1),
+		RunE: func(_ *cobra.Command, args []string) error {
+			cli.StartProgress(" Deleting report rule...")
+			defer cli.StopProgress()
+
+			err := cli.LwApi.V2.ReportRules.Delete(args[0])
+			if err != nil {
+				return errors.Wrap(err, "unable to delete report rule")
+			}
+			cli.OutputHuman(fmt.Sprintf("The report rule with GUID %s was deleted\n", args[0]))
```
afiune, line 146:
```suggestion
			cli.StartProgress(" Deleting report rule...")
			err := cli.LwApi.V2.ReportRules.Delete(args[0])
			cli.StopProgress()
			if err != nil {
				return errors.Wrap(err, "unable to delete report rule")
			}
			cli.OutputHuman("The report rule with GUID %s was deleted\n", args[0])
```

cli/cmd/report_rules.go
```diff
@@ -0,0 +1,412 @@
...
+	}
+
+	cli.StartProgress(" Creating report rule...")
+	defer cli.StopProgress()
+
+	return cli.LwApi.V2.ReportRules.Create(reportRule)
+}
+
+func getEmailAlertChannels() ([]string, map[string]string) {
+	cli.StartProgress("")
```
afiune, line 367: ?

dmurray-lacework: The comment on previous review comment. This is left empty as the transition in the prompt looks weird when there is text in the loader. The text here would also refer to email alert channels which don't appear til further down in the prompt. So I felt it better to leave an empty loader

PR #626, opened 2021-12-01
feat(cli): generate & execute aws terraform code

ipcrm:
## Summary

Enable the automated generation & execution of Terraform code for integrating Lacework with AWS

See tests

https://github.com/ipcrm/go-sdk/blob/ipcrm/ally-392/cli/integration/aws_generation_test.go
https://github.com/ipcrm/go-sdk/blob/ipcrm/ally-392/cli/cli/cmd/generate_aws_test.go

## Issue

https://lacework.atlassian.net/browse/ALLY-392

#### wip

documentation (wiki)


Makefile
```diff
@@ -49,6 +49,13 @@ test: prepare ## Run all go-sdk tests
...
 
+.PHONY: integration-generation
+integration-generation: build-cli-cross-platform integration-generation-only ## Build and run integration tests
+
+.PHONY: integration-generation-only
+integration-generation-only: ## Run integration tests
+	PATH=$(PWD)/bin:${PATH} go test -v github.com/lacework/go-sdk/integration -timeout 30m -run "^TestGeneration" -tags="generation"
+
 .PHONY: integration-only
 integration-only: ## Run integration tests
```
ipcrm, line 60: @dmurray-lacework I think I need to update this to include generation as well?

integration/aws_generation_test.go
```diff
@@ -0,0 +1,472 @@
...
+		},
+		"iac",
+		"aws",
+	)
+
+	// Ensure CLI errored properly
+	assert.Nil(t, runError)
+}
+
+// Test barebones geneation with no customization
```
dmurray-lacework, line 52: nit typo
```suggestion
// Test barebones generation with no customization
```

cli/cmd/generate_execute.go
```diff
@@ -0,0 +1,392 @@
...
+		deleted: resourceDelete,
+		update:  resourceDelete,
+		replace: resourceReplace,
+	}, nil
+}
+
+func TerraformInit(tf *tfexec.Terraform) error {
+	cli.StartProgress("Running terraform init")
+	if err := tf.Init(context.Background(), tfexec.Upgrade(true)); err != nil {
+		return errors.Wrap(err, "failed to execute terraform init")
```
dmurray-lacework, line 251: nit: this prints on the same line as the progress. Would be nice to have a newline before the error

`\Running terraform init ERROR failed to run terraform apply: failed to execute terraform init: exit status 1`

cli/cmd/generate_execute.go
```diff
@@ -0,0 +1,392 @@
...
+	laceworkCmd := "  lacework integration list\n\n"
+	if laceworkProfile != "" {
+		laceworkCmd = fmt.Sprintf("  lacework -p %s integration list\n\n", laceworkProfile)
+	}
+	fmt.Fprint(out, laceworkCmd)
+
+	cli.OutputHuman(out.String())
+}
+
+// this helper function is called when existing the Terraform flow without applying to provide guidance on next steps
```
dmurray-lacework, line 333: Is this comment missing some words, I don't think it quite makes sense

cli/cmd/generate_execute.go
```diff
@@ -0,0 +1,392 @@
...
+//
+// - Run plan
+// - Get plan file details (returned)
+func TerraformExecApply(tf *tfexec.Terraform) error {
+	// Plan
+	cli.StartProgress("Running terraform apply")
+	if err := tf.Apply(context.Background()); err != nil {
+		return err
+	}
+	cli.StopProgress()
```
dmurray-lacework, line 284: I believe you need to defer the stopProgress here. If line 282 gets hit the progress spinner won't get stopped.

Makefile
dmurray-lacework: Yeh, if we want anything with the generation build tag to run in the full integration suite

cli/cmd/generate_aws.go
```diff
@@ -0,0 +1,349 @@
...
+	value := reflect.ValueOf(val)
+
+	// if the value passed is not a string
+	if value.Kind() != reflect.String {
+		return errors.New("value must be a string")
+	}
+
+	// if value doesn't match regex, return invalid arn
+	// original source: https://regex101.com/r/pOfxYN/1
+	matchRegEx := `^arn:(?P<Partition>[^:\n]*):(?P<Service>[^:\n]*):(?P<Region>[^:\n]*):(?P<AccountID>[^:\n]*):(?P<Ignore>(?P<ResourceType>[^:\/\n]*)[:\/])?(?P<Resource>.*)$`
```
dmurray-lacework, line 58: Would very much like a unit test for this regex

cli/cmd/generate.go
```diff
@@ -15,6 +17,7 @@ import (
 type AwsGenerateCommandExtraState struct {
 	Output                string
 	UseExistingCloudtrail bool
+	AwsSubAccounts        string
```
dmurray-lacework, line 20: Any reason not to use []string here?

cli/cmd/generate.go
```diff
@@ -142,6 +145,28 @@ var (
...
 
+			// Parse passed in subaccounts (if any)
+			if GenerateAwsCommandExtraState.AwsSubAccounts != "" {
+				// validate consolidated_cloudtrail is enabled - otherwise this flag cannot be used
+				if ok, _ := cmd.Flags().GetBool("consolidated_cloudtrail"); !ok {
+					return errors.New("aws subaccounts can only be supplied with consolidated cloudtrail enabled")
+				}
+
+				// validate the format of supplied value is correct
+				matchRegEx := `(\w+:\w+(,)?)+`
```
dmurray-lacework, line 156: Similar to the arn regex it would be nice to see a unit test for this.

dmurray-lacework (approved):
Approved but have some comments to address.

afiune:
make it so

afiune:
Pipeline Error:
```
                                                                                                                                                                      
                                Diff:                                                                                                                                 
                                --- Expected                                                                                                                          
                                +++ Actual                                                                                                                            
                                @@ -22,2 +22,3 @@                                                                                                                     
                                   event          Inspect Lacework events                                                                                             
                                +  iac-generate   create iac code                                                                                                     
                                   integration    Manage external integrations                                                                                        
                Test:           TestNoCommandProvided  
```

Also, you need to rebase from main.

cli/cmd/generate_aws.go
```diff
@@ -0,0 +1,375 @@
...
+// entry point for launching a survey to build out the required generation parameters
+func promptAwsGenerate(
+	config *aws.GenerateAwsTfConfigurationArgs,
+	existingIam *aws.ExistingIamRoleDetails,
+	extraState *AwsGenerateCommandExtraState,
+) error {
+	// Cache for later use if generation is abandon and in interactive mode
+	if cli.InteractiveMode() {
+		defer cli.WriteAssetToCache("iac-aws-generate-params", time.Now().Add(time.Hour*1), config)
+		defer cli.WriteAssetToCache("extra-state", time.Now().Add(time.Hour*1), extraState)
```
afiune, line 319: I don't like the name `extra-state` but I understand what it is, could we at least add a prefix in case other providers also have an "extra state"?
```suggestion
		defer cli.WriteAssetToCache("iac-aws-extra-state", time.Now().Add(time.Hour*1), extraState)
```

afiune: Also, could we move these names to the top and have them as constants?

afiune (changes requested):
The code looks great, my only major ask is to not ignore any error and instead, check it and handle it properly.

Things like:
```go
thing, _ := Do()
```
Needs to be checked.

cli/cmd/generate.go
```diff
@@ -0,0 +1,348 @@
...
+var (
+	QuestionRunTfPlan            = "Run Terraform plan now?"
+	GenerateAwsCommandState      = &aws.GenerateAwsTfConfigurationArgs{}
+	GenerateAwsExistingRoleState = &aws.ExistingIamRoleDetails{}
+	GenerateAwsCommandExtraState = &AwsGenerateCommandExtraState{}
+
+	// iac-generate command is used to create IaC code for various environments
+	generateTfCommand = &cobra.Command{
+		Use:     "iac-generate",
+		Aliases: []string{"iac-generate", "iac"},
```
rmoles, line 32: I wonder if we don't need to specify `iac-generate` in the Aliases field?
When I run it locally I see it duplicated in the help output
image

cli/cmd/generate.go
```diff
@@ -0,0 +1,348 @@
...
+					return errors.New("aws subaccounts can only be supplied with consolidated cloudtrail enabled")
+				}
+
+				// validate the format of supplied value is correct
+				matchRegEx := `(\w+:\w+(,)?)+`
+				if ok, _ := regexp.MatchString(matchRegEx, GenerateAwsCommandExtraState.AwsSubAccounts); !ok {
+					return errors.New("supplied aws subaccounts in invalid format")
+				}
+
+				awsSubaccounts := []aws.AwsSubAccount{}
```
rmoles, line 163: nitpick: Could this var be changed to camelCase? `awsSubaccounts` --> `awsSubAccounts`

cli/cmd/generate.go
```diff
@@ -0,0 +1,348 @@
...
+					awsSubaccounts = append(awsSubaccounts, aws.NewAwsSubAccount(accountDetails[0], accountDetails[1]))
+				}
+
+				GenerateAwsCommandState.SubAccounts = awsSubaccounts
+			}
+
+			// Collect and/or confirm parameters
+			err := promptAwsGenerate(GenerateAwsCommandState, GenerateAwsExistingRoleState, GenerateAwsCommandExtraState)
+			if err != nil {
+				return err
```
rmoles, line 175: Should we wrap the error here?
```
return errors.Wrap(err, "Error when collecting/confirming parameters")
```

cli/cmd/generate.go
```diff
@@ -0,0 +1,348 @@
...
+		&GenerateAwsCommandExtraState.Output, "output", "", "location to write generated content")
+
+	// add flags to sub commands
+	// TODO Share the help with the interactive generation
+	generateAwsTfCommand.PersistentFlags().BoolVar(
+		&GenerateAwsCommandState.Cloudtrail, "cloudtrail", false, "enable cloudtrail integration")
+	generateAwsTfCommand.PersistentFlags().BoolVar(
+		&GenerateAwsCommandState.Config, "config", false, "enable config integration")
+	generateAwsTfCommand.PersistentFlags().StringVar(
+		&GenerateAwsCommandState.AwsRegion, "aws_region", "", "specify aws region")
```
rmoles, line 198: I wonder whether we want/need some validation on this field against a list of valid regions? Currently, you can enter anything potentially with small typos that could cause issues.
Perhaps even a future goal would be to provide a list of regions that the user can select from rather than allowing free text?

cli/cmd/generate_aws.go
```diff
@@ -0,0 +1,375 @@
...
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/lacework/go-sdk/lwgenerate/aws"
+	"github.com/pkg/errors"
+)
+
+var (
+	// Define question text here so they can be reused in testing
+	QuestionEnableConfig                = "Enable Config Integration?"
+	QuestionEnableCloudtrail            = "Enable Cloudtrail Integration?"
+	QuestionAwsRegion                   = "Specify the AWS region Cloudtrail, SNS, and S3 resources should use:"
```
rmoles, line 21: Nit: Personally I prefer this wording.
`Specify the AWS region to be used by Cloudtrail, SNS, and S3:`
(non blocking)

cli/cmd/generate_aws.go
```diff
@@ -0,0 +1,375 @@
...
+	QuestionConsolidatedCloudtrail      = "Use consolidated Cloudtrail?"
+	QuestionUseExistingCloudtrail       = "Would you like to use an existing Cloudtrail?"
```
rmoles, line 23: Another nitpick: IMO it would be good to be consistent with the wording here, as we are fairly consistent throughout the rest of the questions. 
Either both to start with `Use` or `Would you like to use`

ipcrm: Good catch @rmoles

cli/cmd/generate_aws.go
ipcrm: 👍

cli/cmd/generate.go
ipcrm: The original thinking here was that we only wanted to do some light validation and let the TF errors come through to the user; that said it would be trivial to add a simple regex test against the region format.  I think giving a list to choose from may be more maintenance than necessary.

cli/cmd/generate.go
ipcrm: Good catch; will update

cli/cmd/generate.go
ipcrm: 👍

cli/cmd/generate.go
ipcrm: Definitely; will update

cli/cmd/generate.go
ipcrm: 👍

cli/cmd/generate_aws.go
ipcrm: 👍  will do

cli/cmd/generate_aws.go
ipcrm: 👍  yep & yep; will update

cli/cmd/generate_execute.go
ipcrm: 👍

cli/cmd/generate_execute.go
ipcrm: 👍

cli/cmd/generate_execute.go
ipcrm: lol; will fix

cli/cmd/generate.go
ipcrm: In this context, the value of `AwsSubAccounts` gets sets to the input value from the flag `--aws_subaccounts`; so at this point it's just a string that is used on L236: https://github.com/lacework/go-sdk/blob/86e614b4408858bbc4bb5fc9582b1e4263ffe21b/cli/cmd/generate.go#L236-L240

cli/cmd/generate.go
```diff
@@ -0,0 +1,348 @@
...
+
+var (
+	QuestionRunTfPlan            = "Run Terraform plan now?"
+	GenerateAwsCommandState      = &aws.GenerateAwsTfConfigurationArgs{}
+	GenerateAwsExistingRoleState = &aws.ExistingIamRoleDetails{}
+	GenerateAwsCommandExtraState = &AwsGenerateCommandExtraState{}
+
+	// iac-generate command is used to create IaC code for various environments
+	generateTfCommand = &cobra.Command{
+		Use:     "iac-generate",
```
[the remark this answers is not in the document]
ipcrm: That I'm not sure on; will discuss with @afiune and @dmurray-lacework

cli/cmd/generate.go
```diff
@@ -0,0 +1,348 @@
...
+		Aliases: []string{"iac-generate", "iac"},
+		Short:   "create iac code",
```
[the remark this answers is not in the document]
ipcrm: 👍

cli/cmd/generate.go
```diff
@@ -0,0 +1,348 @@
...
+		Long:    "Create IaC content for various different cloud environments and configurations",
+		PreRunE: func(cmd *cobra.Command, _ []string) error {
+			// If output location was supplied, validate it exists
+			dirname, err := cmd.Flags().GetString("output")
+			if err == nil {
+				_, err := os.Stat(dirname)
+				if err != nil {
+					return errors.Wrap(err, "could not access specified output location!")
+				}
```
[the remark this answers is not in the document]
ipcrm: 👍

Makefile
ipcrm: 👍

cli/cmd/generate_aws.go
```diff
@@ -0,0 +1,375 @@
...
+	QuestionCloudtrailExistingBucketArn = "Specify an existing bucket ARN used for Cloudtrail logs:"
```
rmoles, line 24: Nit: `Specify an existing bucket ARN used for Cloudtrail logs:` --> `Specify an existing bucket ARN to be used for Cloudtrail logs:`

rmoles, line 24: nit:
```suggestion
	QuestionCloudtrailExistingBucketArn = "Specify an existing bucket ARN to be used for Cloudtrail logs:"
```

cli/cmd/generate_aws.go
```diff
@@ -0,0 +1,375 @@
...
+	QuestionConfigAdvanced              = "Configure advanced integration options?"
+	QuestionAnotherAdvancedOpt          = "Configure another advanced integration option"
+	QuestionCustomizeOutputLocation     = "Provide the location for the output to be written:"
+
+	// select options
+	AdvancedOptDone        = "Done"
+	AdvancedOptCloudTrail  = "Additional Cloudtrail Options"
+	AdvancedOptIamRole     = "Configure Lacework integration with an existing IAM role"
+	AdvancedOptAwsAccounts = "Add Additional AWS Accounts to Lacework"
+	AdvancedOptLocation    = "Customize Output Location"
```
rmoles, line 43:
```suggestion
	AdvancedOptAwsAccounts = "Add additional AWS Accounts to Lacework"
	AdvancedOptLocation    = "Customize output location"
```

cli/cmd/generate_aws.go
rmoles: Is this a REGEX you have generated, or was this pulled from AWS docs? If the latter, could we get a source link?

cli/cmd/generate.go
afiune: Lets follow the same pattern we have in other CLI commands, instead of having a single flag that requires a CSV of things, we use a flag that can be used multiple times, so instead of:
```
lacework iac --aws_subaccounts "profile:region,profile:region"
```
It would be better to have this instead:
```
lacework iac --aws_subaccount "profile:region" --aws_subaccount "profile:region"
```

cli/cmd/generate.go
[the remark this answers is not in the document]
afiune: After a cool zoom/codeshare session we have agreed that we want these commands and aliases:
```
# Commands
lacework cloud-account iac-generate gcp
lacework cloud-account iac-generate aws
lacework cloud-account iac-generate azure
lacework cloud-account list
lacework cloud-account show [id]
lacework cloud-account delete [id]
lacework cloud-account create

# Alias
lacework cloud iac gcp
lacework cloud iac aws
lacework cloud iac azure
lacework cloud ls
lacework cloud show [id]
lacework cloud delete [id]
lacework cloud create
```

cli/cmd/generate.go
rmoles: A regex could be the way to go on this for some light validation. Agreed on the point of maintenance.

integration/aws_generation_test.go
ipcrm: 👍

cli/cmd/generate_execute.go
ipcrm: Reworked these to not use the shorthand if, but rather just define `err` and allow to run the start/stop around that line.

cli/cmd/generate_aws.go
ipcrm: @rmoles the source link is in the comment above this line

cli/cmd/generate_aws.go
ipcrm: This one is actually OK; this is ONLY valid if you have an existing trail setup; so it's not a future state.

cli/cmd/generate.go
ipcrm: 👍

cli/cmd/generate.go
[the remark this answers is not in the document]
ipcrm: 👍

Makefile
```diff
@@ -49,6 +49,13 @@ test: prepare ## Run all go-sdk tests
 .PHONY: integration
 integration: build-cli-cross-platform integration-only ## Build and run integration tests
 
+.PHONY: integration-generation
+integration-generation: build-cli-cross-platform integration-generation-only ## Build and run integration tests
+
+.PHONY: integration-generation-only
+integration-generation-only: ## Run integration tests
+	PATH=$(PWD)/bin:${PATH} go test -v github.com/lacework/go-sdk/integration -timeout 30m -run "^TestGeneration" -tags="generation"
```
rmoles, line 57: Can we bump the version of gotestsum to v1.7.0 and use that here? Just adds a bit of color to the formatting which helps at a glance IMO.
```suggestion
	PATH=$(PWD)/bin:${PATH} gotestsum -- -v github.com/lacework/go-sdk/integration -timeout 30m -run "^TestGeneration" -tags="generation"
```

cli/cmd/generate_aws.go
```diff
@@ -0,0 +1,399 @@
...
+	if !ok {
+		return errors.New(errorString)
+	}
+
+	return nil
+}
+
+// survey.Validator for aws ARNs
+//
+// This isn't service/type specific but rather just validates that an ARN was entered that matches valid ARN formats
```
rmoles, line 75: nitpick: My IDE complains about top-level function comments that aren't prefixed with the function name. I can't claim to know if this is the standard practice in Go, so feel free to ignore.
IDE error:
`Comment should have the following format '<func_name>...' (with an optional leading article) `

Makefile
ipcrm: @rmoles seems reasonable to me; that said looks like we don't do this for the other integration tests so perhaps we should have a maint PR to bump version as well as change utility.  

@afiune @dmurray-lacework thoughts on this?

cli/cmd/generate_aws.go
ipcrm: @rmoles 😬  yuk.  what IDE is that, intellij?

dmurray-lacework:
make it so

Makefile
dmurray-lacework: Sounds good!

rmoles: Shall I fire a PR up to main with this change?

rmoles: https://github.com/lacework/go-sdk/pull/632

cli/cmd/generate_aws.go
rmoles: As discussed on slack, it seems it is actually the IntelliJ IDE Go `Inspector` settings flagging this. This check does point to the [GoLang Docs](https://github.com/golang/go/wiki/CodeReviewComments#comment-sentences), however seems to go above and beyond golangci linter.
image (4)

rmoles: This comment is definitely non blocking

rmoles: Just realising that this comment is on the wrong file the issue was showing in `lwgenerate/aws/aws.go`

afiune:
MAKE_IT_SO!

cli/cmd/cloud_account.go
```diff
@@ -0,0 +1,20 @@
...
+	"github.com/spf13/cobra"
+)
+
+var (
+	// top-level cloud-account command
+	cloudAccountCommand = &cobra.Command{
+		Use:     "cloud-account",
+		Aliases: []string{"cloud", "ca"},
+		Short:   "Manage cloud account(s)",
+		Long:    "Manage cloud account integration with Lacework",
```
afiune, line 13:
```suggestion
		Aliases: []string{"cloud-accounts", "cloud", "ca"},
		Short:   "Manage cloud accounts",
		Long:    "Manage cloud account integrations with Lacework",
```

cli/cmd/generate.go
```diff
@@ -0,0 +1,461 @@
...
+					accountDetails := strings.Split(account, ":")
+					awsSubAccounts = append(awsSubAccounts, aws.NewAwsSubAccount(accountDetails[0], accountDetails[1]))
+				}
+				GenerateAwsCommandState.SubAccounts = awsSubAccounts
+			}
+
+			// Collect and/or confirm parameters
+			err = promptAwsGenerate(GenerateAwsCommandState, GenerateAwsExistingRoleState, GenerateAwsCommandExtraState)
+			if err != nil {
+				return errors.Wrap(err, "error when collecting/confirming parameters")
```
afiune, line 224: This looks odd since it says error twice
```
lacework cloud-account iac-generate aws
▸ Enable Config Integration? No
▸ Enable Cloudtrail Integration? No

ERROR error when collecting/confirming parameters: must enable cloudtrail or config
```
```suggestion
				return errors.Wrap(err, "collecting/confirming parameters")
```

cli/cmd/cloud_account.go
ipcrm: 👍

cli/cmd/generate.go
ipcrm: 👍

afiune (approved):
tenor-189291708

afiune:
Make it so!!

tenor-234508091

PR #627, opened 2021-12-01
feat(api): Vulnerability Exceptions v2 Service

dmurray-lacework:

## Summary
New Vulnerability Exceptions service added to go-sdk .

api/vulnerability_exceptions_test.go
api/examples/vulnerability-exceptions/main.go

## Issue
https://lacework.atlassian.net/browse/ALLY-749

### DIFF
--- a/api/vulnerability_exceptions.go
+++ b/api/vulnerability_exceptions.go
@@ -1,0 +1,450 @@
+//
+// Author:: Darren Murray (<darren.murray@lacework.net>)
+// Copyright:: Copyright 2021, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package api
+
+import (
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/pkg/errors"
+
+	"github.com/lacework/go-sdk/lwtime"
+)
+
+// VulnerabilityExceptionsService is the service that interacts with
+// the VulnerabilityExceptions schema from the Lacework APIv2 Server
+type VulnerabilityExceptionsService struct {
+	client *Client
+}
+
+// vulnerabilityExceptionResourceScope is an interface for the 2 types of vulnerability exceptions resource scopes:
+// 'VulnerabilityExceptionContainerResourceScope' or 'VulnerabilityExceptionHostResourceScope'
+type vulnerabilityExceptionResourceScope interface {
+	Type() vulnerabilityExceptionType
+	Scope() VulnerabilityExceptionResourceScope
+}
+
+// vulnerabilityExceptionReason represents the types of vulnerability exceptions reasons:
+//'False Positive', 'Accepted Risk', 'Compensating Controls', 'Fix Pending' or 'Other'
+type vulnerabilityExceptionReason int
+
+const (
+	VulnerabilityExceptionReasonAcceptedRisk vulnerabilityExceptionReason = iota
+	VulnerabilityExceptionReasonAcceptedFalsePositive
+	VulnerabilityExceptionReasonCompensatingControls
+	VulnerabilityExceptionReasonFixPending
+	VulnerabilityExceptionReasonOther
+	VulnerabilityExceptionReasonUnknown
+)
+
+var VulnerabilityExceptionReasons = map[vulnerabilityExceptionReason]string{
+	VulnerabilityExceptionReasonAcceptedRisk:          "Accepted Risk",
+	VulnerabilityExceptionReasonAcceptedFalsePositive: "False Positive",
+	VulnerabilityExceptionReasonCompensatingControls:  "Compensating Controls",
+	VulnerabilityExceptionReasonFixPending:            "Fix Pending",
+	VulnerabilityExceptionReasonOther:                 "Other",
+	VulnerabilityExceptionReasonUnknown:               "Unknown",
+}
+
+func (i vulnerabilityExceptionReason) String() string {
+	return VulnerabilityExceptionReasons[i]
+}
+
+func NewVulnerabilityExceptionReason(reason string) vulnerabilityExceptionReason {
+	switch reason {
+	case "Accepted Risk":
+		return VulnerabilityExceptionReasonAcceptedRisk
+	case "False Positive":
+		return VulnerabilityExceptionReasonAcceptedFalsePositive
+	case "Compensating Controls":
+		return VulnerabilityExceptionReasonCompensatingControls
+	case "Fix Pending":
+		return VulnerabilityExceptionReasonFixPending
+	case "Other":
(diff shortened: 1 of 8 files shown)


api/vulnerability_exceptions.go
```diff
@@ -0,0 +1,453 @@
...
+	ExceptionName         string                              `json:"exceptionName"`
+	ExceptionType         string                              `json:"exceptionType"`
+	ExceptionReason       string                              `json:"exceptionReason"`
+	Props                 VulnerabilityExceptionProps         `json:"props"`
+	VulnerabilityCriteria VulnerabilityExceptionCriteria      `json:"vulnerabilityCriteria"`
+	ResourceScope         VulnerabilityExceptionResourceScope `json:"resourceScope,omitempty"`
+	CreatedTime           string                              `json:"createdTime,omitempty"`
+	UpdatedTime           string                              `json:"updatedTime,omitempty"`
+	// ExpiryTime is an int(unixmillis) on request & string(rfc3339) on response
+	ExpiryTime interface{} `json:"expiryTime,omitempty"`
```
afiune, line 410: do we have a ticket for fixing this?

api/vulnerability_exceptions_test.go
```diff
@@ -0,0 +1,305 @@
...
+
+func singleMockVulnerabilityException(id string) string {
+	return fmt.Sprintf(`
+{
+        "createdTime": "2021-12-02T18:47:24.897Z",
+        "envGuid": "TECHALLY_3F75784E9C7660B1F8B4A72E80B2EECCE8EDA92DBE37E5D",
+        "exceptionGuid": %q,
+        "exceptionName": "Example Vuln Exception",
+        "exceptionReason": "Accepted Risk",
+        "exceptionType": "Host",
```
afiune, line 250: should we have a container test?

api/vulnerability_exceptions.go
```diff
@@ -0,0 +1,453 @@
...
+	VulnerabilityExceptionReasonUnknown
+)
+
+var VulnerabilityExceptionReasons = map[vulnerabilityExceptionReason]string{
+	VulnerabilityExceptionReasonAcceptedRisk:          "Accepted Risk",
+	VulnerabilityExceptionReasonAcceptedFalsePositive: "False Positive",
+	VulnerabilityExceptionReasonCompensatingControls:  "Compensating Controls",
+	VulnerabilityExceptionReasonFixPending:            "Fix Pending",
+	VulnerabilityExceptionReasonOther:                 "Other",
+	VulnerabilityExceptionReasonUnknown:               "",
```
afiune, line 61:
```suggestion
	VulnerabilityExceptionReasonUnknown:               "Unknown",
```

api/vulnerability_exceptions.go
```diff
@@ -0,0 +1,453 @@
...
+//       },
+//       ExpiryTime: time.Now().AddDate(0, 1, 0),
+//   }
+//
+//	 vulnerabilityException := api.NewVulnerabilityException("vulnerabilityException", exception)
+//
+//   client.V2.VulnerabilityExceptions.Create(vulnerabilityException)
+//
+func NewVulnerabilityException(name string, exception VulnerabilityExceptionConfig) VulnerabilityException {
+	packages := AggregatePackages(exception.Package)
```
afiune, line 217: Is `AggregatePackages` public on purpose? Can we prefix the function to be something like `VulnerabilityExceptionAggregatePackages()` just because there could be other "aggregation of packages" that is not for vulnerability exceptions

api/vulnerability_exceptions.go
afiune, line 61: Should we send something to the APIs so that they tell us that we need to set it, I think if we send an empty string it will just say that an exception name is required. 🤔
```suggestion
	VulnerabilityExceptionReasonUnknown:               "Unknown",
```

api/vulnerability_exceptions.go
```diff
@@ -0,0 +1,453 @@
...
+	}
+}
+
+const (
+	VulnerabilityExceptionSeverityCritical vulnerabilityExceptionSeverity = 1
+	VulnerabilityExceptionSeverityHigh     vulnerabilityExceptionSeverity = 2
+	VulnerabilityExceptionSeverityMedium   vulnerabilityExceptionSeverity = 3
+	VulnerabilityExceptionSeverityLow      vulnerabilityExceptionSeverity = 4
+	VulnerabilityExceptionSeverityInfo     vulnerabilityExceptionSeverity = 5
+	VulnerabilityExceptionSeverityUnknown  vulnerabilityExceptionSeverity = 0
```
afiune, line 183: why do we need numbers?

api/vulnerability_exceptions.go
dmurray-lacework: I will change it to private.

api/vulnerability_exceptions_test.go
dmurray-lacework: Yes!

api/vulnerability_exceptions.go
dmurray-lacework: 👍

api/vulnerability_exceptions.go
dmurray-lacework: We do not need this, the severities here are different that other schemas

integration/alert_rules_test.go
```diff
@@ -28,7 +28,7 @@ import (
 	"github.com/stretchr/testify/assert"
 )
 
-func TestAlertRuleRead(t *testing.T) {
+func _TestAlertRuleRead(t *testing.T) {
```
afiune, line 31: no need to do this
```suggestion
func TestAlertRuleRead(t *testing.T) {
```

api/vulnerability_exceptions.go
```diff
@@ -0,0 +1,448 @@
...
+}
+
+// List returns a list of Vulnerability Exceptions
+func (svc *VulnerabilityExceptionsService) List() (response VulnerabilityExceptionsResponse, err error) {
+	err = svc.client.RequestDecoder("GET", apiV2VulnerabilityExceptions, nil, &response)
+	return
+}
+
+// Create creates a single Vulnerability Exception
+func (svc *VulnerabilityExceptionsService) Create(vuln VulnerabilityException) (
```
afiune, line 281: I would suggest removing `Create()` and `Update()`  functions since we have specific functions for every type and that way we will follow a pattern.

api/vulnerability_exceptions_container.go
```diff
@@ -0,0 +1,85 @@
...
+	err = svc.client.RequestDecoder("GET", apiPath, nil, &response)
+	return
+}
+
+func (svc *VulnerabilityExceptionsService) ListVulnerabilityExceptionsContainers() (response VulnerabilityExceptionContainerResponse, err error) {
+	err = svc.client.RequestDecoder("GET", apiV2VulnerabilityExceptions, nil, &response)
+	return
+}
+
+func (svc *VulnerabilityExceptionsService) UpdateVulnerabilityExceptionsContainer(data VulnerabilityException, id string) (
```
afiune, line 48: Why do we ask for the `id` if the`VulnerabilityException` struct already has the `Guid` field? feels redundant

dmurray-lacework: Unfortunately, the request body is invalid if the guid field is present. I believe there is another resource we had to do something similar.

afiune: @dmurray-lacework I think that is how we have done it in LQL, TeamMembers, and other v2 schemas. We should be consistent and follow the adopted pattern. (which is set the id and strip it in the update func)

afiune (approved):
**One comment but out of that, this is ready to go. Just curious why CI is upset... 🤔**

tenor-51055055

PR #628, opened 2021-12-02
feature(api): implement crud operations for org team member

vatasha:
## Summary
Implement CRUD operations on org level team members. 

There are tests in addition to an example file that calls most of the methods. 

## Issue
https://lacework.atlassian.net/browse/ALLY-766?atlOrigin=eyJpIjoiNmFjYWE2NjkzZDdmNDQ1MWExNWFjMTdhOTc3NmFlMGMiLCJwIjoiaiJ9

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -217,6 +217,12 @@ func (c *Client) ValidAuth() bool {
 	return c.auth.token != ""
 }
 
+// OrgAccess verifies that the client is managing data at the org level.
+func (c *Client) OrgAccess() bool {
+	return c.headers["Org-Access"] == "true" // check if Org-Access is set to true, if so, the client is configured to manage org level dataset
+
+}
+
 // newID generates a new client id, this id is useful for logging purposes
 // when there are more than one client running on the same machine
 func newID() string {
--- a/api/client.go
+++ b/api/client.go
@@ -217,10 +217,10 @@ func (c *Client) ValidAuth() bool {
 	return c.auth.token != ""
 }
 
-// OrgAccess verifies that the client is managing data at the org level.
+// OrgAccess check if the Org-Access header is set to 'true', if so,
+// the client is configured to manage org level dataset
 func (c *Client) OrgAccess() bool {
-	return c.headers["Org-Access"] == "true" // check if Org-Access is set to true, if so, the client is configured to manage org level dataset
-
+	return c.headers["Org-Access"] == "true"
 }
 
 // newID generates a new client id, this id is useful for logging purposes
--- a/api/team_members.go
+++ b/api/team_members.go
@@ -20,9 +20,10 @@ package api
 
 import (
 	"encoding/json"
-	"errors"
 	"fmt"
 	"io/ioutil"
+
+	"github.com/pkg/errors"
 )
 
 type TeamMembersService struct {
@@ -166,14 +167,12 @@ func (svc *TeamMembersService) UpdateOrg(tm TeamMemberOrg) (res TeamMemberOrgRes
 		return
 	}
 	var userGuid string
-	tms, err := svc.SearchUsername(tm.UserName)
-	if err != nil {
-		return res, err
-	} else if len(tms.Data) > 0 {
-		userGuid = tms.Data[0].UserGuid
-	} else {
-		return res, errors.New("unable to find user with specified username")
+	tms, errSearch := svc.SearchUsername(tm.UserName)
+	if errSearch != nil || len(tms.Data) == 0 {
+		err = errors.Wrap(err, "unable to find user with specified username")
+		return
 	}
+	userGuid = tms.Data[0].UserGuid
 	// Omit UserGuid from the patch body as it cannot be modified
 	tm.UserGuid = ""
 	// Omit userEnabled field from the patch body as it cannot be modified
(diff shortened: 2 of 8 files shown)


api/client.go
```diff
@@ -217,6 +217,12 @@ func (c *Client) ValidAuth() bool {
 	return c.auth.token != ""
 }
 
+// OrgAccess verifies that the client is managing data at the org level.
+func (c *Client) OrgAccess() bool {
+	return c.headers["Org-Access"] == "true" // check if Org-Access is set to true, if so, the client is configured to manage org level dataset
+
+}
```
afiune, line 224:
```suggestion
// OrgAccess check if the Org-Access header is set to 'true', if so,
// the client is configured to manage org level dataset
func (c *Client) OrgAccess() bool {
	return c.headers["Org-Access"] == "true"
}
```

afiune: Also, can we write unit tests for this?

api/team_members.go
```diff
@@ -64,26 +99,139 @@ func (svc *TeamMembersService) List() (res TeamMembersResponse, err error) {
...
+	}
+	var userGuid string
+	tms, err := svc.SearchUsername(tm.UserName)
+	if err != nil {
+		return res, err
+	} else if len(tms.Data) > 0 {
+		userGuid = tms.Data[0].UserGuid
+	} else {
+		return res, errors.New("unable to find user with specified username")
+	}
```
afiune, line 176: You are using named returns and as a best practice it is not recommended to mask the same variables with the operator `:=` - One suggestion could be:
```suggestion
	tms, errSearch := svc.SearchUsername(tm.UserName)
	if errSearch != nil || len(tms.Data) == 0 {
		err = errors.Wrap(errSearch, "unable to find user with specified username")
		return
	}
	userGuid := tms.Data[0].UserGuid

```

afiune: If you don't like that then just avoid using named returns and pass the variables to every `return x, y`

api/team_members.go
```diff
@@ -64,26 +99,139 @@ func (svc *TeamMembersService) List() (res TeamMembersResponse, err error) {
...
+	tm.UserGuid = ""
+	// Omit userEnabled field from the patch body as it cannot be modified
+	tm.UserEnabled = 0
+	// Omit userName field from the patch body as it cannot be modified
+	tm.UserName = ""
+	// Omit Company field from the patch body as it cannot be modified
+	tm.Props.Company = ""
+	patchBody, err := jsonReader(tm)
+	if err != nil {
+		return res, err
```
afiune, line 187: same, you are using named returns so, we need to stick to naked returns like this
```suggestion
		return
```

api/team_members.go
```diff
@@ -64,26 +99,139 @@ func (svc *TeamMembersService) List() (res TeamMembersResponse, err error) {
...
+	}
+
+	req, err := svc.client.NewRequest("PATCH", fmt.Sprintf(apiV2TeamMembersFromGUID, userGuid), patchBody)
+	if err != nil {
+		return res, err
```
afiune, line 192:
```suggestion
		return
```

api/team_members.go
```diff
@@ -64,26 +99,139 @@ func (svc *TeamMembersService) List() (res TeamMembersResponse, err error) {
...
+	}
+
+	response, err := svc.client.Do(req)
+	if err != nil {
+		return res, err
```
afiune, line 197:
```suggestion
		return
```

api/team_members.go
```diff
@@ -64,26 +99,139 @@ func (svc *TeamMembersService) List() (res TeamMembersResponse, err error) {
...
+
+	if err != nil {
+		return res, err
+	}
+
+	err = json.Unmarshal(body, &res)
+	if err != nil {
+		return res, err
+	}
+	return
```
afiune, line 212: this if is useless
```suggestion
	return
```

api/team_members.go
```diff
@@ -64,26 +99,139 @@ func (svc *TeamMembersService) List() (res TeamMembersResponse, err error) {
...
+	if err != nil {
+		return res, err
+	}
+
+	defer response.Body.Close()
+
+	body, err := ioutil.ReadAll(response.Body)
+
+	if err != nil {
+		return res, err
```
afiune, line 205:
```suggestion
		return
```

afiune (changes requested):
One thing I noticed while reviewing this PR is the fact that having the functions be `CreateOrg()` and `DeleteOrg()` feels to me that we are doing something to the organization, what do you think about switching them all to be `OrgCreate()`, `OrgDelete()`, etc.? Thoughts?

api/team_members.go
```diff
@@ -64,26 +100,142 @@ func (svc *TeamMembersService) List() (res TeamMembersResponse, err error) {
...
+		err = errors.Wrap(bodyErr, "unable to read json body")
+		return
+	}
+
+	unmarshalErr := json.Unmarshal(body, &res)
+	if unmarshalErr != nil {
+		err = errors.Wrap(unmarshalErr, "unable to deserialize response body into team member struct")
+		return
+	}
+	return
```
afiune, line 216: Can this be replaced by:
```suggestion
	err = svc.client.RequestEncoderDecoder("PATCH", fmt.Sprintf(apiV2TeamMembersFromGUID, userGuid), tm, &res)
```

or are we doing something special?

api/team_members.go
```diff
@@ -64,26 +98,100 @@ func (svc *TeamMembersService) List() (res TeamMembersResponse, err error) {
...
+	if tm.UserName == "" {
+		err = errors.New("please specify a username")
+		return
+	}
+	tms, errSearch := svc.SearchUsername(tm.UserName)
+	if errSearch != nil || len(tms.Data) == 0 {
+		err = errors.Wrap(err, "unable to find user with specified username")
+		return
+	}
+	return svc.UpdateOrgById(tm)
```
afiune, line 172: Uhmmm I feel there is something missing here, you are just passing tm without modifying the struct with the userGuid found on the search. Am I wrong?

afiune (changes requested):

afiune (approved):
tenor-173588404

PR #629, opened 2021-12-03
test: disable failing alert rule tests

dmurray-lacework:


afiune:
No need to do this anymore since APIs are working again. Thanks @dmurray-lacework

PR #632, opened 2021-12-08
chore(deps): Bump gotestsum version and use instead of go test

rmoles:

## Summary
Previously gotestsum was added to add some colorful output on test runs. It seems this hasn't been used across the make commands and therefore we have missed bumping this after the upgrade to Go 1.17

make test ✅ 
make fmt ✅ 

## Issue
https://lacework.atlassian.net/browse/ALLY-791

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -57,7 +57,7 @@ integration-generation-only: ## Run integration tests
 	PATH=$(PWD)/bin:${PATH} go test -v github.com/lacework/go-sdk/integration -timeout 30m -run "^TestGeneration" -tags="generation"
 
 .PHONY: integration-only
-integration-only: ## Run integration tests
+integration-only: install-tools ## Run integration tests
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -50,8 +50,8 @@ test: prepare ## Run all go-sdk tests
 integration: build-cli-cross-platform integration-only ## Build and run integration tests
 
 .PHONY: integration-only
-integration-only: ## Run integration tests
-	PATH=$(PWD)/bin:${PATH} go test -v github.com/lacework/go-sdk/integration -timeout 30m -tags="\
+integration-only: install-tools ## Run integration tests
```
dmurray-lacework, line 53: why move install-tools under this command? CI already runs `make prepare` prior to run integration-only.

afiune: I guess we need it now since this directive depends on `gotestsum` and if the developer doesn't have it installed, it will fail. I am ok with this.

afiune (approved):
tenor-30161288

PR #636, opened 2021-12-09
refactor(test_resources): remove Terraform code dep

afiune:
## Summary
We have reached the limit of VPCs in our AWS account and instead of creating new ones, this code is trying to reuse an existing VPC for multiple environments.

About to run the prepare-integrations pipeline in Codefresh...

## Issue
N/A


dmurray-lacework (approved):

integration/aws_generation_test.go
```diff
@@ -565,7 +565,6 @@ func runGenerateTest(t *testing.T, conditions func(*expect.Console), args ...str
 	// create a temporal directory where we will check that the
 	// configuration file is deployed (.lacework.toml)
 	dir := createDummyTOMLConfig()
-	fmt.Println(dir)
```
afiune, line 568: @ipcrm this is the code I mentioned in standup.

PR #641, opened 2021-12-14
test(iac-generation): add tests for tf install/detection

ipcrm:
## Summary

Improve testing for automated terraform detection/installation 

See changes (all tests)

## Issue

ALLY-395

### DIFF
--- a/integration/aws_generation_test.go
+++ b/integration/aws_generation_test.go
@@ -19,9 +19,9 @@ import (
 )
 
 func expectString(c *expect.Console, str string, runError *error) {
-	out, err := c.Expect(expect.WithTimeout(time.Second), expect.String(str))
+	_, err := c.Expect(expect.WithTimeout(time.Second), expect.String(str))
 	if err != nil {
-		fmt.Println(out) // To see the errored line, you can enable this and update _ above to out
+		// fmt.Println(out) // To see the errored line, you can enable this and update _ above to out
 		*runError = err
 	}
 }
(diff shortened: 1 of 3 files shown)


integration/aws_generation_test.go
```diff
@@ -19,9 +19,9 @@ import (
 )
 
 func expectString(c *expect.Console, str string, runError *error) {
-	out, err := c.Expect(expect.WithTimeout(time.Second), expect.String(str))
+	_, err := c.Expect(expect.WithTimeout(time.Second), expect.String(str))
 	if err != nil {
-		fmt.Println(out) // To see the errored line, you can enable this and update _ above to out
+		// fmt.Println(out) // To see the errored line, you can enable this and update _ above to out
```
afiune, line 24: A better pattern I would suggest is this:
```go
	out, err := c.Expect(expect.WithTimeout(time.Second), expect.String(str))
	if err != nil {
		if os.Getenv("DEBUG") != "" {
			fmt.Println(out) // To see the errored line, you can run tests with DEBUB=true
		}
```

integration/generation_terraform_test.go
```diff
@@ -0,0 +1,149 @@
...
+	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/hashicorp/go-version"
+	"github.com/hashicorp/hc-install/product"
+	"github.com/hashicorp/hc-install/releases"
+	"github.com/lacework/go-sdk/cli/cmd"
```
afiune, line 16: ugh...... We have never done this..... 😟 I don't know if I like the way we are importing the CLI.

PR #642, opened 2021-12-15
feat(cli): add apply flag for aws iac generation & fix cache issues

ipcrm:
## Summary

For AWS IaC generation; add apply flag, add cached state protections

See tests

## Issue

ALLY-395

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -18,6 +18,7 @@ type AwsGenerateCommandExtraState struct {
 	Output                string
 	UseExistingCloudtrail bool
 	AwsSubAccounts        []string
+	TerraformApply        bool
 }
 
 var (
@@ -120,10 +121,9 @@ This command can also be run in noninteractive mode however, only generation is 
 			}
 
 			// Prompt to execute
-			execute := false
 			err = SurveyQuestionInteractiveOnly(SurveyQuestionWithValidationArgs{
-				Prompt:   &survey.Confirm{Default: execute, Message: QuestionRunTfPlan},
-				Response: &execute,
+				Prompt:   &survey.Confirm{Default: GenerateAwsCommandExtraState.TerraformApply, Message: QuestionRunTfPlan},
+				Response: &GenerateAwsCommandExtraState.TerraformApply,
 			})
 
 			if err != nil {
@@ -142,14 +142,14 @@ This command can also be run in noninteractive mode however, only generation is 
 
 			// Execute
 			locationDir := filepath.Dir(location)
-			if execute {
+			if GenerateAwsCommandExtraState.TerraformApply {
 				if err := TerraformPlanAndExecute(locationDir); err != nil {
 					return errors.Wrap(err, "failed to run terraform apply")
 				}
 			}
 
 			// Output where code was generated
-			if !execute {
+			if !GenerateAwsCommandExtraState.TerraformApply {
 				provideGuidanceAfterExit(false, false, locationDir, "terraform")
 			}
 
@@ -320,6 +320,12 @@ func init() {
 		"aws_subaccount",
 		[]string{},
 		"configure an additional aws account; value format must be <aws profile>:<region>")
+	generateAwsTfCommand.PersistentFlags().BoolVar(
+		&GenerateAwsCommandExtraState.TerraformApply,
+		"apply",
+		false,
+		"run terraform apply without executing plan or prompting",
+	)
 
 	// add sub-commands to the iac-generate command
 	generateTfCommand.AddCommand(generateAwsTfCommand)
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -139,22 +139,23 @@ This command can also be run in noninteractive mode however, only generation is 
 			})
 
 			if err != nil {
-				return errors.Wrap(err, "failed to run terraform execution")
-			}
-
-			// Execution pre-run check
-			ok, err = TerraformExecutePreRunCheck(dirname)
-			if err != nil {
-				return errors.Wrap(err, "failed to check for existing terraform state")
-			}
-
-			if !ok {
-				return errors.Wrap(err, "aborting to avoid overwriting existing terraform state")
+				return errors.Wrap(err, "failed to promopt for terraform execution")
 			}
 
 			// Execute
 			locationDir := filepath.Dir(location)
 			if GenerateAwsCommandExtraState.TerraformApply {
+				// Execution pre-run check
+				ok, err = TerraformExecutePreRunCheck(dirname)
+				if err != nil {
+					return errors.Wrap(err, "failed to check for existing terraform state")
+				}
+
+				if !ok {
+					cli.OutputHuman(provideGuidanceAfterExit(false, false, locationDir, "terraform"))
+					return nil
+				}
+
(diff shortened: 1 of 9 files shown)


cli/cmd/generate_execute.go
```diff
@@ -370,30 +370,33 @@ func TerraformPlanAndExecute(workingDir string) error {
 		return err
 	}
 
-	// Write plan
-	changes, err := TerraformExecPlan(tf)
-	if err != nil {
-		return err
-	}
+	// If running in nonInteractive mode, dont run a plan or prompt to proceed; just proceed
+	if !cli.nonInteractive {
```
afiune, line 374: could we please use the function, rather than one single variable?
```suggestion
	if !cli.InteractiveMode() {
```

cli/cmd/generate_execute.go
```diff
@@ -370,30 +370,33 @@ func TerraformPlanAndExecute(workingDir string) error {
...
+			return err
+		}
 
-	// If not proceed; display guidance on how to continue outside of this session
-	if !proceed {
-		provideGuidanceAfterExit(true, true, tf.WorkingDir(), tf.ExecPath())
-		return nil
+		// If not proceed; display guidance on how to continue outside of this session
+		if !proceed {
+			provideGuidanceAfterExit(true, true, tf.WorkingDir(), tf.ExecPath())
```
afiune, line 389: apologies for bringing this up but, can we add unit tests for this function? I don't thin we are testing it and is building some valuable message to the user.

cli/cmd/generate_execute.go
```diff
@@ -370,30 +370,33 @@ func TerraformPlanAndExecute(workingDir string) error {
...
+		}
 	}
 
 	// Apply plan
 	if err := TerraformExecApply(tf); err != nil {
 		return err
 	}
-	provideGuidanceAfterSuccess(tf.WorkingDir(), GenerateAwsCommandState.LaceworkProfile)
 
+	provideGuidanceAfterSuccess(tf.WorkingDir(), GenerateAwsCommandState.LaceworkProfile)
```
afiune, line 399: Same for this one, we want to make sure that this guidance is helpful messages that don't break. We have done some tests in other places where we leave breadcrumbs to the user.

afiune (changes requested):
Almost done, left a few comments.

tenor-58718559

cli/cmd/generate.go
```diff
@@ -106,10 +107,9 @@ var (
 			}
 
 			// Prompt to execute
-			execute := false
 			err = SurveyQuestionInteractiveOnly(SurveyQuestionWithValidationArgs{
-				Prompt:   &survey.Confirm{Default: execute, Message: QuestionRunTfPlan},
-				Response: &execute,
+				Prompt:   &survey.Confirm{Default: GenerateAwsCommandExtraState.TerraformApply, Message: QuestionRunTfPlan},
```
rmoles, line 111: Wonder if the Message should be changed? `QuestionRunTfPlan` makes me think that we are just going to run the Plan action rather than Apply, which it seems will be the default for this now?

cli/cmd/generate_execute.go
ipcrm: 👍

cli/cmd/generate.go
ipcrm: So there are really two modes of operating here; one interactive and one non-interactive.

The question shown here is prompted only in interactive mode; so that's still OK (prompt for Plan execution, and then we prompt again for Apply execution).

If you pass `--apply` in interactive mode, all that happens is the default response in that prompt changes from N to Y, but you still (again, in interactive mode) have to go through a plan, review the results, and then confirm you want an apply to run.

If you pass `--apply` with `--non-interactive` then there is no option to run a plan, just the apply is run with `-auto-approve` flag set on the terraform command.  

So I think this one is OK to leave as is.  I'll resolve this, but we can unresolved this comment if there are concerns!

cli/cmd/generate_execute.go
ipcrm: 👍

cli/cmd/generate_execute.go
ipcrm: 👍

cli/cmd/generate_terraform_test.go
```diff
@@ -0,0 +1,215 @@
...
+}
+
+// Test the feedback given after running terraform to completion (success)
+func TestGenerationTfExecutionSuccess(t *testing.T) {
+	t.Run("using default lacework profile", func(t *testing.T) {
+		out := provideGuidanceAfterSuccess("/tmp", "")
+		assert.Equal(t, `Lacework integration was successful! Terraform code saved in /tmp
+
+Use the Lacework CLI to view integration status:
+  lacework integration list
```
afiune, line 202: nit: what about this format that we use in help messages and other guidance output:
```suggestion
To view integration status:
    lacework integration list
```

cli/cmd/generate_terraform_test.go
```diff
@@ -0,0 +1,215 @@
...
+
+	"github.com/hashicorp/go-version"
+	"github.com/hashicorp/hc-install/product"
+	"github.com/hashicorp/hc-install/releases"
+	"github.com/stretchr/testify/assert"
+)
+
+// Test various scenarios for detection and/or installation of Terraform binary
+func TestGenerationTfInstallation(t *testing.T) {
+	if os.Getenv("CI") == "true" {
```
afiune, line 19: Can we change the logic of this `if` statement and use `t.Skip()` instead?
```suggestion
	if os.Getenv("CI") != "" {
		t.Skip(fmt.Sprintf("skipping %s", t.Name()))
	}
```

cli/cmd/generate_terraform_test.go
```diff
@@ -0,0 +1,215 @@
...
+			installPath, err := installer.Install(context.Background())
+			if err != nil {
+				panic(err)
+			}
+			installedTempPath := filepath.Dir(installPath)
+			defer os.RemoveAll(installedTempPath)
+
+			// Update path to find the new version
+			oldPath := os.Getenv("PATH")
+			os.Setenv("PATH", fmt.Sprintf("%s:%s", installedTempPath, oldPath))
```
afiune, line 97: Does this work on Windows? 🤔

afiune (changes requested):

afiune:
Check windows! 🪟😂

cli/cmd/generate_terraform_test.go
ipcrm: 👍

cli/cmd/generate_terraform_test.go
ipcrm: 👍

cli/cmd/generate_terraform_test.go
ipcrm: No, it did not; so I updated here: https://github.com/lacework/go-sdk/pull/642/commits/4026aaaebf8b28b285140d176ab41303adb17215#diff-6cc7b544129624856ba0189d1ac4c8bb7995e5706282949c4bcc44d8b82c06a8R106

There's a little logic at the top of that test that figures out the appropriate env separator depending on platform. That allows this set of tests to pass in windows.

afiune (approved):
tenor-265848016

ipcrm:
check windows

PR #645, opened 2021-12-17
fix: handle unsupported integration types

dmurray-lacework:

## Summary

CLI panics when showing an unsupported integration type

The fix is to fetch and output the raw response using api v1 when a type is unsupported.

run `lacework int show <id>` with id being an unsupported type eg. SAML_CFG

## Issue

https://lacework.atlassian.net/browse/ALLY-775

### DIFF
--- a/cli/cmd/integration.go
+++ b/cli/cmd/integration.go
@@ -105,19 +105,22 @@ var (
 				return errors.New(msg)
 			}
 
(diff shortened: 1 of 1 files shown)


afiune (commented):
I think we talked about a different approach, I will wait for this to be reworked.

afiune (approved):
tenor-203756859

PR #649, opened 2022-01-07
feat(cli): add csv output support to vuln reports

ipcrm:
## Summary

Add CSV output support to vulnerability listings or show assessment commands.

Ran tests (see changes & ran CLI)

## Issue

ALLY-771


cli/cmd/vuln_container.go
```diff
@@ -199,20 +206,21 @@ filter on containers with vulnerabilities that have fixes available.`,
 				return cli.OutputJSON(filteredAssessments)
 			}
 
+			// Build data
 			rows := vulAssessmentsToTable(filteredAssessments)
+			headers := []string{"Registry", "Repository", "Last Scan", "Status", "Containers", "Vulnerabilities", "Image Digest"}
 
+			switch {
```
afiune, line 213: maaaann!!! I have never seen this switch statement, I LOVE IT! 🥇 

tenor-120575326

cli/cmd/vuln_container.go
```diff
@@ -246,6 +255,18 @@ To request an on-demand vulnerability scan:
 
     lacework vulnerability container scan <registry> <repository> <tag|digest>`,
 		Args: cobra.ExactArgs(1),
+		PreRunE: func(cmd *cobra.Command, args []string) error {
+			if vulCmdState.Csv {
+				cli.EnableCSVOutput()
+
+				// If --details or --packages is not passed, csv outputs nothing; defaulting to --details
```
afiune, line 262: why? if I pass `--csv` with or without those two flags, I think we should just output CSV.

This is the same behavior if someone passes `--pdf`, we just output it. (meaning it has more priority)

cli/cmd/vulnerability.go
```diff
@@ -353,23 +353,28 @@ func buildVulnContainerAssessmentReports(assessment *api.VulnContainerAssessment
...
 		filteredAssessment := assessment
 		filteredAssessment.Image.ImageLayers = details.VulnerabilityDetails.ImageLayers
 		if err := cli.OutputJSON(filteredAssessment); err != nil {
 			return err
 		}
-	} else {
+	case cli.CSVOutput():
+		if err := cli.OutputCSV(buildVulnerabilityDetailsReportCSV(details)); err != nil {
+			return err
+		}
```
afiune, line 366: For these switch statements, what happens if more than one case statement is true? to they execute in order? so if someone here passes `--cvs --json` we will display ONLY Json output? 🤔

afiune (commented):
I don't think CVS flag should change with the filters. 🤔 Do you?

cli/cmd/vulnerability.go
ipcrm: Yep; that's right.  These execute in order. 

For that situation specifically, we should probably look for conflicting output formats and error out; but as the code exists today a customer would get a different result than they might expect based on the order.

cli/cmd/vuln_container.go
ipcrm: The challenge with that is the outputs from these commands essentially output 3 tables.  Typically a host summary, the details summary, and the data relevant to CSV output is only generated if you pass `--details` or `--packages`. 

So we can either add these implicit flags as I did here, or we can build a totally different 'path' in the code for CSV output.  Let's discuss this further.

cli/cmd/vuln_container.go
ipcrm: 🙌

afiune (approved):
tenor-184182789

PR #654, opened 2022-01-12
feat(lwgenerate): Add initial hcl generation support GCP - ALLY-393

rmoles:
## Summary
Add initial support for Terraform code generation for GCP

Added unit tests. Coverage is now at 92%

## Issue

https://lacework.atlassian.net/browse/ALLY-393

### DIFF
--- a/lwgenerate/aws/aws.go
+++ b/lwgenerate/aws/aws.go
@@ -4,8 +4,9 @@ import (
 	"fmt"
 
 	"github.com/hashicorp/hcl/v2/hclwrite"
-	"github.com/lacework/go-sdk/lwgenerate"
 	"github.com/pkg/errors"
+
+	"github.com/lacework/go-sdk/lwgenerate"
 )
 
 type ExistingIamRoleDetails struct {
@@ -249,8 +250,8 @@ func (args *GenerateAwsTfConfigurationArgs) Generate() (string, error) {
 func createRequiredProviders() (*hclwrite.Block, error) {
 	return lwgenerate.CreateRequiredProviders(
 		lwgenerate.NewRequiredProvider("lacework",
-			lwgenerate.HclRequiredProviderWithSource("lacework/lacework"),
-			lwgenerate.HclRequiredProviderWithVersion("~> 0.12.2")))
+			lwgenerate.HclRequiredProviderWithSource(lwgenerate.LaceworkProviderSource),
+			lwgenerate.HclRequiredProviderWithVersion(lwgenerate.LaceworkProviderVersion)))
 }
 
 func createAwsProvider(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block, error) {
@@ -307,9 +308,6 @@ func createLaceworkProvider(args *GenerateAwsTfConfigurationArgs) (*hclwrite.Blo
 }
 
 func createConfig(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block, error) {
-	source := "lacework/config/aws"
-	version := "~> 0.1"
-
 	blocks := []*hclwrite.Block{}
 	if args.Config {
 		// Add main account
@@ -319,8 +317,11 @@ func createConfig(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block, erro
 				lwgenerate.HclModuleWithProviderDetails(map[string]string{"aws": "aws.main"}))
 		}
 
-		moduleBlock, err := lwgenerate.NewModule("aws_config", source,
-			append(moduleDetails, lwgenerate.HclModuleWithVersion(version))...).ToBlock()
+		moduleBlock, err := lwgenerate.NewModule(
+				"aws_config",
+				lwgenerate.AwsConfigSource,
+				append(moduleDetails, lwgenerate.HclModuleWithVersion(lwgenerate.AwsConfigVersion))...
+			).ToBlock()
 
 		if err != nil {
 			return nil, err
@@ -330,8 +331,8 @@ func createConfig(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block, erro
 		// Add sub accounts
 		for _, subaccount := range args.SubAccounts {
 			configModule, err := lwgenerate.NewModule(fmt.Sprintf("aws_config_%s", subaccount.AwsProfile),
-				source,
-				lwgenerate.HclModuleWithVersion(version),
+				lwgenerate.AwsConfigSource,
+				lwgenerate.HclModuleWithVersion(lwgenerate.AwsConfigVersion),
 				lwgenerate.HclModuleWithProviderDetails(map[string]string{
 					"aws": fmt.Sprintf("aws.%s", subaccount.AwsProfile),
 				})).ToBlock()
@@ -350,7 +351,7 @@ func createConfig(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block, erro
 func createCloudtrail(args *GenerateAwsTfConfigurationArgs) (*hclwrite.Block, error) {
 	if args.Cloudtrail {
 		attributes := map[string]interface{}{}
-		modDetails := []lwgenerate.HclModuleModifier{lwgenerate.HclModuleWithVersion("~> 0.1")}
+		modDetails := []lwgenerate.HclModuleModifier{lwgenerate.HclModuleWithVersion(lwgenerate.AwsCloudTrailVersion)}
 
 		if args.ForceDestroyS3Bucket && args.ExistingCloudtrailBucketArn == "" {
 			attributes["bucket_force_destroy"] = true
@@ -392,7 +393,7 @@ func createCloudtrail(args *GenerateAwsTfConfigurationArgs) (*hclwrite.Block, er
 			lwgenerate.HclModuleWithAttributes(attributes),
 		)
 
-		return lwgenerate.NewModule("main_cloudtrail", "lacework/cloudtrail/aws", modDetails...).ToBlock()
+		return lwgenerate.NewModule("main_cloudtrail", lwgenerate.AwsCloudTrailSource, modDetails...).ToBlock()
 	}
 
 	return nil, nil
--- a/lwgenerate/aws/aws.go
+++ b/lwgenerate/aws/aws.go
@@ -270,7 +270,10 @@ func createAwsProvider(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block,
 			attrs["alias"] = "main"
 		}
 
-		provider, err := lwgenerate.NewProvider("aws", lwgenerate.HclProviderWithAttributes(attrs)).ToBlock()
+		provider, err := lwgenerate.NewProvider(
+			"aws",
+			lwgenerate.HclProviderWithAttributes(attrs),
+		).ToBlock()
 		if err != nil {
 			return nil, err
 		}
@@ -285,7 +288,10 @@ func createAwsProvider(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block,
 				"profile": subaccount.AwsProfile,
 				"region":  subaccount.AwsRegion,
 			}
-			providerBlock, err := lwgenerate.NewProvider("aws", lwgenerate.HclProviderWithAttributes(attrs)).ToBlock()
+			providerBlock, err := lwgenerate.NewProvider(
+				"aws",
+				lwgenerate.HclProviderWithAttributes(attrs),
+			).ToBlock()
 
 			if err != nil {
 				return nil, err
@@ -300,7 +306,8 @@ func createAwsProvider(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block,
 
 func createLaceworkProvider(args *GenerateAwsTfConfigurationArgs) (*hclwrite.Block, error) {
 	if args.LaceworkProfile != "" {
-		return lwgenerate.NewProvider("lacework",
+		return lwgenerate.NewProvider(
+			"lacework",
 			lwgenerate.HclProviderWithAttributes(map[string]interface{}{"profile": args.LaceworkProfile}),
 		).ToBlock()
 	}
@@ -318,10 +325,10 @@ func createConfig(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block, erro
 		}
 
 		moduleBlock, err := lwgenerate.NewModule(
-				"aws_config",
-				lwgenerate.AwsConfigSource,
-				append(moduleDetails, lwgenerate.HclModuleWithVersion(lwgenerate.AwsConfigVersion))...
-			).ToBlock()
+			"aws_config",
+			lwgenerate.AwsConfigSource,
+			append(moduleDetails, lwgenerate.HclModuleWithVersion(lwgenerate.AwsConfigVersion))...,
+		).ToBlock()
 
 		if err != nil {
 			return nil, err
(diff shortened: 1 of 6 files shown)


lwgenerate/gcp/gcp.go
```diff
@@ -0,0 +1,505 @@
...
+
+// AuditLogLabels set labels to be applied to ALL newly created AuditLog resources
+func AuditLogLabels(labels map[string]string) GcpTerraformModifier {
+	return func(c *GenerateGcpTfConfigurationArgs) {
+		c.AuditLogLabels = labels
+	}
+}
+
+// BucketLabels set labels to be applied to the newly created AuditLog Bucket
+func BucketLabels(labels map[string]string) GcpTerraformModifier {
```
ipcrm, line 192: nit: I'm wondering if the `GcpTerraformModifier` functions could have a consistent naming convention?  We start by using `With` then move to `Use` and the last set of functions don't use either

lwgenerate/gcp/gcp.go
```diff
@@ -0,0 +1,505 @@
...
+	}
+
+	laceworkProvider, err := createLaceworkProvider(args)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to generate lacework provider")
+	}
+
+	configModule, err := createConfig(args)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to generate aws config module")
```
ipcrm, line 308: typo: `s/aws/gcp/g`

Makefile
```diff
@@ -12,7 +12,7 @@ CIARTIFACTS?=ci-artifacts
 COVERAGEOUT?=coverage.out
 COVERAGEHTML?=coverage.html
 PACKAGENAME?=lacework-cli
-CLINAME?=lacework
+CLINAME?=lacework-dev
```
ipcrm, line 15: I'm sure this is just for testing purposes, but just calling out so it doesn't get missed later on when opening a PR

cli/cmd/generate.go
```diff
@@ -38,11 +38,33 @@ var (
...
+			// Generate TF Code
+			cli.StartProgress("Generating Azure Terraform Code...")
+			return nil
+		},
+	}
+	// GCP command
+	generateGcpTfCommand = &cobra.Command{
+		Use:   "gcloud",
+		Short: "Generate code for gcp environment",
+		Long:  "Generate Terraform code for deploying into a new GCP environment.",
```
ipcrm, line 56: Just a note here (because I messed this up originally) these `Long` descriptions are now auto-converted to docs so we'll need to build out this section (see `main` branch for example).

ipcrm (commented):
@rmoles I think this is off to a great start!  I made some minor comments.  One thing to note is I'm not an expert on the downstream TF modules so we'll definitely need to recruit some help to validate the settings are appropriate; etc.

lwgenerate/gcp/gcp.go
ipcrm: I see the same inconsistency exists for AWS: https://github.com/lacework/go-sdk/blob/main/lwgenerate/aws/aws.go#L170

rmoles: Indeed this inconsistency exists elswhere. Will try to make a move towards standardising the naming in this class.

rmoles: My personal preference is to drop Use and With from the functions and just match the var names in the `GenerateGcpTfConfigurationArgs`

ipcrm: Honestly, I could go either way; as long as we do it the same all good with me.  We can go and update the other parts of the project.

lwgenerate/gcp/gcp.go
```diff
@@ -0,0 +1,505 @@
+package gcp
+
+import (
+	"github.com/hashicorp/hcl/v2/hclwrite"
+	"github.com/lacework/go-sdk/lwgenerate"
+	"github.com/pkg/errors"
```
dmurray-lacework, line 6: nit group imports
```suggestion
	"github.com/hashicorp/hcl/v2/hclwrite"
	"github.com/pkg/errors"

	"github.com/lacework/go-sdk/lwgenerate"
```

lwgenerate/gcp/gcp.go
```diff
@@ -0,0 +1,505 @@
...
+)
+
+type ExistingServiceAccountDetails struct {
+	// Existing Service Account Name
+	Name string
+
+	// Existing Service Account private key in JSON format, base64 encoded
+	Base64EncodedPrivateKey string
```
dmurray-lacework, line 14: Naming is a little too verbose? Given the comment about explains it's base64. Perhaps just PrivateKey?

lwgenerate/gcp/gcp.go
dmurray-lacework: I'd lean towards the 'With' prefix, It makes it clear these funcs are used as part of the creational pattern to construct a GenerateGcpTfConfigurationArgs

afiune: The use of `With[XYZ]()` is what we have across this project, I think we should stick with it. 

I do agree there are inconsistencies https://github.com/lacework/go-sdk/blob/main/cli/cmd/generate.go#L83-L90 but you can see how easy it would be to know that modifiers start with `With...()` and when you start chaining them, it just makes perfect sense. 

Still, if this package follows a different pattern, I am ok with it as well. 🙌🏽

lwgenerate/gcp/gcp.go
```diff
@@ -0,0 +1,505 @@
...
+			auditLogModule),
+	)
+	return hclBlocks, nil
+}
+
+func createRequiredProviders() (*hclwrite.Block, error) {
+	return lwgenerate.CreateRequiredProviders(
+		lwgenerate.NewRequiredProvider("lacework",
+			lwgenerate.HclRequiredProviderWithSource("lacework/lacework"),
+			lwgenerate.HclRequiredProviderWithVersion("~> 0.12.2")))
```
afiune, line 332: Somehow I feel we need a single place that has all of these versions defined as constants so that we don't have to go find them deep in the codebase when we have to do a major version bump.

cli/cmd/generate.go
```diff
@@ -51,6 +51,28 @@ var (
 		Long:    "Create IaC content for various different cloud environments and configurations",
 	}
 
+	// Azure command
+	generateAzureTfCommand = &cobra.Command{
```
afiune, line 55: this PR is marked to be a feature to the `lwgenerate` Go package, should we move any other change to CLI or any other package to a different PR? or are you going to also add the CLI bits already here?

afiune (commented):
Loooking pretty good, definitely needs testing!

tenor-126147810

cli/cmd/generate.go
rmoles: Fair point!
I can drop this out of this PR and move to the future CLI PR :)

lwgenerate/gcp/gcp.go
rmoles: I may as well keep us consistent with the patterns already in place. Will update to use `With...()`

lwgenerate/gcp/gcp.go
rmoles: Wondering if we should create a constants file at the top level of the `lwgenerate` folder and set the versions and provider sources there?

lwgenerate/gcp/gcp.go
rmoles: done

Makefile
rmoles: done

cli/cmd/generate.go
rmoles: will address this in the future CLI PRs. Dropped this file change from the PR.

lwgenerate/gcp/gcp.go
rmoles: done

lwgenerate/gcp/gcp.go
rmoles: done

cli/cmd/generate.go
rmoles: dropped from this PR 😄

lwgenerate/gcp/gcp.go
rmoles: Created an `lwgenerate` package `constants.go` and updated :)

lwgenerate/gcp/gcp.go
```diff
@@ -0,0 +1,521 @@
...
+	ServiceAccountCredentials string
+
+	// Should we configure an Organization wide integration?
+	OrganizationIntegration bool
+
+	// Supply a GCP Organization ID, only asked if OrganizationIntegration is True
+	GcpOrganizationId string
+
+	// Supply a GCP Project ID, to host the new resources
+	GcpProjectId string
```
rmoles, line 43: Would it be fair to make this a required field in the context of running this via the CLI, rather than relying on the end-user to pass this using env vars?
The Project ID must be set one way or another, but for the purposes of building the TF via the CLI it would be nice to require this to be passed in always, and then we can set it in the google provider.
I'd appreciate any input on this @afiune @ipcrm @dmurray-lacework @mikelaramie

rmoles: This would result in us building the googler Terraform provider like so
```
provider "google" {
  credentials = "/path/to/credentials"
  project     = "project1"
}
```

lwgenerate/aws/aws.go
```diff
@@ -299,17 +306,15 @@ func createAwsProvider(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block,
 
 func createLaceworkProvider(args *GenerateAwsTfConfigurationArgs) (*hclwrite.Block, error) {
 	if args.LaceworkProfile != "" {
-		return lwgenerate.NewProvider("lacework",
+		return lwgenerate.NewProvider(
+			"lacework",
```
afiune, line 310: just curious to know if any linting tool you use did this? or that was you formatting it 🤔

lwgenerate/gcp/gcp.go
mikelaramie: W.r.t. `project` in the provider: I love this idea unless we're considering creating the project as part of this workflow.  Makes it a lot easier than specifying `project=` in each resource.  

W.r.t `credentials`: Are we requiring a service account, or is that credentials field optional?  Ideally we should default to gcloud/application default credentials for auth but provide the option for a SA if necessary.

lwgenerate/aws/aws.go
rmoles: This was done manually. Interestingly the formatter doesn't format it back to the original layout.
Personally, I find it easier to parse in this format with the args stacked. I can change it back if needs be.

lwgenerate/hcl.go
```diff
@@ -204,6 +210,12 @@ func setBlockAttributeValue(block *hclwrite.Block, key string, val interface{})
 			data[attrKey] = value
 		}
 		block.Body().SetAttributeValue(key, cty.ObjectVal(data))
+	case map[string]string:
+		value, err := convertTypeToCty(v)
+		if err != nil {
+			return err
+		}
+		block.Body().SetAttributeValue(key, value)
```
rmoles, line 218: @ipcrm I've added explicit support for `map[string]string` rather than using `map[string]interface`. I want to be explicit that the labels can only be string and not allow values of any other type. Does this change seem reasonable at the hcl.go level?

ipcrm: @rmoles I don't have any issue with this; I see how your using it and it makes sense to me to be able to set `map[string]string` and not have to do any additional validation.

ipcrm (changes requested):
@rmoles looking good! I'm still taking a look but would you please add GCP to the examples? https://github.com/lacework/go-sdk/tree/ALLY-393/lwgenerate/_examples

dmurray-lacework (approved):
Great job!

lwgenerate/aws/aws.go
afiune: I am not against style changes but I have seen others do the same but backward, so we are just playing ping-pong until we have a tool we all use that is opinionated and dictates our desired format/style. Leave it, just some food for thought.

lwgenerate/gcp/gcp_test.go
```diff
@@ -0,0 +1,574 @@
+package gcp
```
afiune, line 1: Are we testing the functions exposed to the consumers? or internal functions?

I feel these are tests for consumers so we should rename the test package and import the `lwgenerate/gcp` package as a subject
```suggestion
package gcp_test
```

lwgenerate/gcp/gcp.go
```diff
@@ -0,0 +1,533 @@
...
+			configModuleName = "gcp_organization_level_config"
+			auditLogModuleName = "gcp_organization_level_audit_log"
+			attributes["org_integration"] = args.OrganizationIntegration
+			attributes["organization_id"] = args.GcpOrganizationId
+		}
+
+		if args.ExistingServiceAccount == nil && args.Config {
+			attributes["use_existing_service_account"] = true
+			attributes["service_account_name"] = lwgenerate.CreateSimpleTraversal([]string{"module", configModuleName, "service_account_name"})
+			attributes["service_account_private_key"] = lwgenerate.CreateSimpleTraversal([]string{"module", configModuleName, "service_account_private_key"})
```
afiune, line 508: nit: max line length should not be more than 100 characters for readability

afiune (approved):
tenor-125580557

lwgenerate/gcp/gcp.go
```diff
@@ -0,0 +1,533 @@
...
+
+		if args.PubSubTopicLabels != nil {
+			attributes["pubsub_topic_labels"] = args.PubSubTopicLabels
+		}
+
+		if args.ExistingLogBucketName != "" {
+			attributes["existing_bucket_name"] = args.ExistingLogBucketName
+		} else {
+			if args.LogBucketLifecycleRuleAge == nil {
+				defaultValue := -1
```
ipcrm, line 451: nit: Do we need to default this? Looks like the TF module already handles this and it would kinda be not great to have to maintain this default in two places?

ipcrm, line 451: nit: Do we need to default this? Looks like the TF module already handles this and it would kinda be not great to have to maintain this default in two places?

lwgenerate/gcp/gcp.go
```diff
@@ -0,0 +1,533 @@
...
+	Config bool
+
+	// Path to service account credentials to be used by Terraform
+	ServiceAccountCredentials string
+
+	// Should we configure an Organization wide integration?
+	OrganizationIntegration bool
+
+	// Supply a GCP Organization ID, only asked if OrganizationIntegration is True
+	GcpOrganizationId string
```
ipcrm, line 39: nit: Do we need both of these?  Could the presence of an org id set org integration to true? I noticed you have to supply both when using an org and in most cases, we are testing the presence of both of these values so it seems it may be redundant.

lwgenerate/aws/aws.go
rmoles: Interesting when you consider how opinionated Go and Go fmt is, that it accepts both styles as valid.

ipcrm (approved):
@rmoles This looks really good!  I've made a couple of comments but I don't think we need to block on them if you feel strongly about not changing. One item of note, I compared the TF output to the documentation and that looks correct (minus the default that I noted), but I'm not the expert on the GCP TF modules!

lwgenerate/gcp/gcp.go
rmoles: I guess we could just not set this var if the memory address is nil. It's a bit of a tricky one to handle via the CLI survey, because 0 is a valid option.

ipcrm: Iif I'm understanding correctly; in that situation, the `LogBucketLifecycleRuleAge` would not be nil and you'd just set the value? If it's not supplied it's nil and the code wouldn't populate it.

lwgenerate/gcp/gcp.go
rmoles: Fair question... I was simply copying the convention we have within the TF modules. 

I guess we can infer the value that we should be setting the Terraform `org_integration` var boolean to, based on whether the user supplies an organization id.
Any objections to this approach @afiune @dmurray-lacework ? I can't think of a scenario where we would be setting the organization ID and it not being an org integration 🤔

lwgenerate/gcp/gcp.go
rmoles: Indeed. Which, the more I think about it, is likely the correct way to go on this. Good catch!

lwgenerate/gcp/gcp.go
rmoles: Going to leave this one as-is for now. I may revisit when it comes to implementing the survey, as it mightn't make sense to have the boolean as part of the survey

PR #655, opened 2022-01-13
fix: non_os_package_support omitted from json request when false

dmurray-lacework:
Remove omitempty from no_os_package_support field


afiune (approved):
tenor-42207164

PR #659, opened 2022-01-19
test: fix host vuln test

dmurray-lacework:
fix logic in host vuln test

### DIFF
--- a/integration/host_vulnerability_test.go
+++ b/integration/host_vulnerability_test.go
@@ -414,7 +414,10 @@ func validRFC3339(field string, jsonString string) bool {
 
 func machineIDExists(lw *api.Client) bool {
 	resp, err := lw.Vulnerabilities.Host.GetHostAssessment(machineID)
-	return err != nil || !resp.Ok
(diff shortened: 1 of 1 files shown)


integration/host_vulnerability_test.go
```diff
@@ -414,7 +414,10 @@ func validRFC3339(field string, jsonString string) bool {
 
 func machineIDExists(lw *api.Client) bool {
 	resp, err := lw.Vulnerabilities.Host.GetHostAssessment(machineID)
-	return err != nil || !resp.Ok
+	if err != nil {
```
rmoles, line 417: Wondering if it could be reduced to `return err == nil && resp.Ok` ?

dmurray-lacework: Could do, but I think the way it is now is a little more clear. We had a 1 liner here previously.

dmurray-lacework: Gonna leave as it for now

rmoles (approved):
LGTM

PR #661, opened 2022-01-20
feat(cli): add lacework_content_library helper

hazedav:
## Summary

Similar to our LQL API library we will need a helper that abstracts our interactions with the Lacework Content Library component

Adds automated functional testing

## Issue

https://lacework.atlassian.net/browse/ALLY-811


hazedav:
@afiune perhaps you could weigh in on where you might want this code structurally.  Perhaps in a sibling of "api" and "cmd" packages (i.e. `api_component`)

afiune (approved):

afiune:
I don't see anything else using this code but the CLI, but I might be missing something. We should talk.

hazedav:
> I don't see anything else using this code but the CLI, but I might be missing something. We should talk.

We talked and agreed that it makes sense to stay as part of `cmd` package.

PR #662, opened 2022-01-21
feat(lwcomponent): add component artifacts

hazedav:
## Summary
As a lwcomponent developer I may have unique binaries for different platforms/architectures and in-so-doing each of these binaries may have unique signatures.  This dictates that we don’t have just a single Signature associated with a component, but instead a list of a architectures

Automated testing

## Issue
https://lacework.atlassian.net/browse/ALLY-830


afiune (approved):
I like this a lot, we need to update the design document (proposal).

PR #664, opened 2022-01-21
feat(cli): make policyID optional for policy create

hazedav:
## Summary
We want to discourage users from defining their own policyId. Also, I learned that the policyId's format is invalid.

Automated integration tests added

## Issue
https://lacework.atlassian.net/browse/ALLY-835

### DIFF
--- a/integration/policy_test.go
+++ b/integration/policy_test.go
@@ -65,7 +65,8 @@ policies:
 )
 
 var (
-	policyIDRE *regexp.Regexp = regexp.MustCompile(`([\w-]+-cli.*?test-1)`)
+	policyIDRE *regexp.Regexp = regexp.MustCompile(
+		`([\w-]+-(?:(?:cli.*?test-1)|(?:default-\d+)))`)
 )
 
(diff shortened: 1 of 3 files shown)


integration/policy_test.go
```diff
@@ -236,18 +238,18 @@ func TestPolicyCreateStdin(t *testing.T) {
...
 	}
+	assert.Contains(t, out.String(), "was created.")
+	assert.Empty(t, stderr.String(), "STDERR should be empty")
+	assert.Equal(t, 0, exitcode, "EXITCODE is not the expected one")
 
 	policyID, err := getPolicyIdFromStdout(out.String())
-	assert.Nil(t, err)
+	if err != nil {
+		assert.FailNow(t, err.Error())
+	}
```
afiune, line 248: why these changes?

afiune (approved):
tenor-228301063

integration/policy_test.go
hazedav: We want to fail fast when unable to extract policyID.  Subsequent calls require a valid policyID and will implicitly fail w/o them....

PR #671, opened 2022-01-24
feat(cli): Manage Team Members

dmurray-lacework:
## Summary

Manage Team Members in the CLI

```
lacework team-members create
lacework team-members list
lacework team-members show <id>
lacework team-members delete <id>
```

run lacework tm create | list | show | delete
add unit test -> go-sdk/cli/cmd/team_members_test.go

test for org level
run lacework tm create | list | show | delete --organization

## Issue

https://lacework.atlassian.net/browse/ALLY-815

### DIFF
--- a/cli/cmd/team_members.go
+++ b/cli/cmd/team_members.go
@@ -1,0 +1,403 @@
+//
+// Author:: Darren Murray(<darren.murray@lacework.net>)
+// Copyright:: Copyright 2021, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package cmd
+
+import (
+	"fmt"
+	"regexp"
+	"strconv"
+	"strings"
+
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/olekukonko/tablewriter"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+
+	"github.com/lacework/go-sdk/api"
+)
+
+var (
+	// team-members command is used to manage lacework team members
+	teamMembersCommand = &cobra.Command{
+		Use:     "team-member",
+		Aliases: []string{"team-members", "tm"},
+		Short:   "Manage team members",
+		Long: `Manage Team Members to grant or restrict access to multiple Lacework Accounts. 
+			  Team members can also be granted organization-level roles.
+`,
+	}
+
+	// list command is used to list all lacework team members
+	teamMembersListCommand = &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "List all team members",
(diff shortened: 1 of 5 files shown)


cli/cmd/team_members.go
```diff
@@ -0,0 +1,392 @@
...
+			}
+			cli.StopProgress()
+
+			if cli.JSONOutput() {
+				return cli.OutputJSON(response)
+			}
+
+			teamMember := response.Data
+			headers := [][]string{{teamMember.UserGuid, teamMember.UserName, enabled(teamMember.UserEnabled)}}
+			cli.OutputHuman(renderSimpleTable([]string{"GUID", "NAME", "ENABLED"}, headers))
```
afiune, line 111:
```suggestion
			cli.OutputHuman(renderSimpleTable([]string{"GUID", "NAME", "STATE"}, headers))
```

cli/cmd/team_members.go
```diff
@@ -0,0 +1,392 @@
...
+	detailsTable.WriteString("\n")
+	return detailsTable.String()
+}
+
+func promptCreateTeamMember() (string, error) {
+	questions := []*survey.Question{
+		{
+			Name:     "email",
+			Prompt:   &survey.Input{Message: "Email: "},
+			Validate: survey.Required,
```
afiune, line 215: Can we add a simple validation to this field, at least check that the string has an `@` and a `.`

afiune: That way we don't send the request to the APIs to have it return
```
ERROR unable to create team member:
  [POST] https://customerdemo.lacework.net/api/v2/TeamMembers
  [400] validemail: invalid value for userName; username: not a well-formed email address;
```

afiune (commented):
Should we add a way to disable team members? Sounds like a good use-case. Thoughts?

dmurray-lacework:
> Should we add a way to disable team members? Sounds like a good use-case. Thoughts?

I think in perhaps another ticket. I don't know what that command looks like atm. I think we can add some ideas to a new ticket.

cli/cmd/team_members.go
```diff
@@ -0,0 +1,403 @@
...
+		Long:  "Delete a single team member by it's ID.",
+		Args:  cobra.ExactArgs(1),
+		RunE: func(_ *cobra.Command, args []string) error {
+			cli.StartProgress(" Deleting team member...")
+			err := cli.LwApi.V2.TeamMembers.Delete(args[0])
+			cli.StopProgress()
+			if err != nil {
+				return errors.Wrap(err, "unable to delete team member")
+			}
+			cli.OutputHuman("The team member with GUID %s was deleted\n", args[0])
```
afiune, line 133: Could we agree on a pattern for delete operations? We can clearly see that many engineers print this message in different forms:
```
cli/cmd/alert_rules.go:                 cli.OutputHuman(fmt.Sprintf("The alert rule with GUID %s was deleted \n", args[0]))
cli/cmd/integration.go:                 cli.OutputHuman("The integration %s was deleted.\n", args[0])
cli/cmd/lql_delete.go:  cli.OutputHuman("The query %s was deleted.\n", args[0])
cli/cmd/policy.go:              fmt.Sprintf("The policy %s was deleted.\n", args[0]))
cli/cmd/report_rules.go:                        cli.OutputHuman("The report rule with GUID %s was deleted\n", args[0])
cli/cmd/team_members.go:                        cli.OutputHuman("The team member with GUID %s was deleted\n", args[0])
```

dmurray-lacework:
> When creating a user at the organization level I get:
> 
> Screen Shot 2022-02-01 at 8 02 03 AM
> 
> ```
> ERROR unable to create team member: client configured to manage account-level datasets, use Create()
> ```

We require the --organization flag for this. Should we set org header if the user selects the organzation in the prompt?

PR #677, opened 2022-01-26
feat(cli): add --cascade flag to delete policy and query

hazedav:
## Summary
As a Lacework user I want to be able to delete a policy and it’s associated query in “one go”.  If we implement a “force” flag for deletion we can attempt to delete the query associated with the policy.

```
lacework policy delete lwcustom-28 --cascade
```

Integration testing added

## Issue
https://lacework.atlassian.net/browse/ALLY-852

### DIFF
--- a/cli/cmd/policy_delete.go
+++ b/cli/cmd/policy_delete.go
@@ -45,8 +45,8 @@ func init() {
 	policyCmd.AddCommand(policyDeleteCmd)
 
 	policyDeleteCmd.Flags().BoolVar(
-		&policyCmdState.ForceDelete,
-		"force", false, "delete policy and its associated query",
+		&policyCmdState.CascadeDelete,
+		"cascade", false, "delete policy and its associated query",
 	)
 }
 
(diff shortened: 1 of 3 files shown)


cli/cmd/policy_delete.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+)
+
+func init() {
+	// add sub-commands to the policy command
+	policyCmd.AddCommand(policyDeleteCmd)
+
+	policyDeleteCmd.Flags().BoolVar(
+		&policyCmdState.ForceDelete,
+		"force", false, "delete policy and its associated query",
```
hazedav, line 49: Is this OK for a flag name or is something like `--recursive` more intuitive?

afiune (commented):
This is going to main..... What happens if someone tries to delete a policy that points to an out-of-the-box query coming from US (Lacework) - Can users delete such queries/policies? Do we want them to delete them?

hazedav:
> This is going to main..... 
This is intentionally going to main because it's not dependent on any component related work.

> What happens if someone tries to delete a policy that points to an out-of-the-box query coming from US (Lacework).  Can users delete such queries/policies? Do we want them to delete them?
This is controlled by the API, and the workflow being introduced in this PR isn't really impacted by this.  

We simply allow you to optionally delete the query pointed at by the policy as most policies:queries are 1:1.  If the policy can't be deleted we won't attempt to delete the query.  If the query is pointed at by another policy or can't be deleted then it won't be deleted....

afiune (approved):
tenor-136837340

PR #678, opened 2022-01-26
refactor(cli): abstract file funcs to package

hazedav:
## Summary
As we look to leverage caching and file checking in additional packages it makes sense to migrate helper functions out of the cmd package and into something more suitable for import.

Covered by existing automation and integration testing

## Issue
https://lacework.atlassian.net/browse/ALLY-844

### DIFF
--- a/cli/cmd/cache.go
+++ b/cli/cmd/cache.go
@@ -25,7 +25,7 @@ import (
 	"time"
 
 	"github.com/lacework/go-sdk/internal/format"
-	homedir "github.com/mitchellh/go-homedir"
+	lwint "github.com/lacework/go-sdk/internal/lacework"
 	"github.com/peterbourgon/diskv/v3"
 )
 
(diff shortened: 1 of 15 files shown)


cli/cmd/cache.go
```diff
@@ -25,7 +25,7 @@ import (
 	"time"
 
 	"github.com/lacework/go-sdk/internal/format"
-	homedir "github.com/mitchellh/go-homedir"
+	lwint "github.com/lacework/go-sdk/internal/lacework"
```
afiune, line 28: lets not use `lwint` since it looks/sounds like lacework int = lacework integration 🤔 

It confused me a bit. I also don't know how to call this new package..... 🤔 `lacework` sounds too broad, but if we call it `lwcache`, then I don't know how `lwcache.FileExist()` would look like. 

I would suggest two packages:
* `/internal/cache` for cache stuff
* `/internal/file` for general file functions

We already have an internal package for `array` so it makes sense

afiune (changes requested):

hazedav:
@afiune / @dmurray-lacework , any additional concerns here?

afiune (approved):
tenor-127799501

PR #679, opened 2022-01-27
feat(cli): cascading policy create/update

hazedav:
## Summary
When creating a policy attempt to create the query
```
lacework policy create --library lwcustom-18
The query LW_Custom_AWS_CTA_CloudTrailCSVMaliciousFormula was created.
The policy dev7-lwcustom-18 was created.
```

When updating a policy attempt to update the query
```
lacework policy update --library lwcustom-18
The query LW_Custom_AWS_CTA_CloudTrailCSVMaliciousFormula was updated.
The policy dev7-lwcustom-18 was updated.
```

Manually

## Issue
https://lacework.atlassian.net/browse/ALLY-853

### DIFF
--- a/cli/cmd/policy_create.go
+++ b/cli/cmd/policy_create.go
@@ -20,6 +20,7 @@ package cmd
 
 import (
 	"fmt"
+	"strings"
 
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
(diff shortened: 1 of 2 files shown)


cli/cmd/policy_create.go
```diff
@@ -92,6 +135,12 @@ func createPolicy(cmd *cobra.Command, _ []string) error {
 	if cli.JSONOutput() {
 		return cli.OutputJSON(createResponse.Data)
 	}
+	if queryExists {
+		cli.OutputHuman(fmt.Sprintf("The query %s already exists.\n", newPolicy.QueryID))
```
afiune, line 139: Should we call this out? 🤔 I don't think we should (my own opinion). I would make it a log message

afiune (approved):

cli/cmd/policy_create.go
hazedav: Well we're communicating that we're creating a query and a policy and if we don't output anything about the query it might be a wtf moment...

hazedav: Going to merge this down, but let me know what you prefer here and we can still make it happen.

PR #682, opened 2022-02-01
feat(lwgenerate): Add initial hcl generation support Azure - ALLY-394

djmctavish:
### Summary
Add initial support for HCL Terraform code generation for Azure

### How did you test this change?
Added unit tests. Coverage is now at 92.1%
Checked the generated hcl with terraform init, terraform plan

### Issue
https://lacework.atlassian.net/browse/ALLY-394


lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,250 @@
...
+
+		if args.ActivityLogIntegrationName != "" {
+			attributes["lacework_integration_name"] = args.ActivityLogIntegrationName
+		}
+
+		// Set default values
+		attributes["use_existing_ad_application"] = true
+		attributes["application_id"] = "module.az_ad_application.application_id"
+		attributes["application_password"] = "module.az_ad_application.application_password"
+		attributes["service_principal_id"] = "module.az_ad_application.service_principal_id"
```
dmurray-lacework, line 231: Are these defaults correct? https://github.com/lacework/terraform-azure-activity-log/tree/main/examples/default-activity-log

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,250 @@
...
+	if args.Config {
+		attributes := map[string]interface{}{}
+		moduleDetails := []lwgenerate.HclModuleModifier{}
+
+		if args.ConfigIntegrationName != "" {
+			attributes["lacework_integration_name"] = args.ConfigIntegrationName
+		}
+
+		// Set default values
+		attributes["use_existing_ad_application"] = true
```
dmurray-lacework, line 193: Are these defaults correct? https://github.com/lacework/terraform-azure-config/tree/main/examples/default-config

dmurray-lacework (changes requested):
A few comments. I think the defaults/tf needs looked at again
Also, eed to add examples to 'lwgenerate/_examples/azure/main.go'

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,250 @@
+package azure
+
+import (
+	"github.com/hashicorp/hcl/v2/hclwrite"
+	"github.com/lacework/go-sdk/lwgenerate"
+	"github.com/pkg/errors"
+)
+
+type GenerateAzureTfConfigurationArgs struct {
```
ipcrm, line 9: I would expect this struct to mirror (closely, likely not exactly) the options found...

here: https://registry.terraform.io/modules/lacework/activity-log/azure/latest#inputs

and here: https://registry.terraform.io/modules/lacework/config/azure/latest#inputs

Or are we handling that another way?

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,250 @@
...
+
+		if args.ConfigIntegrationName != "" {
+			attributes["lacework_integration_name"] = args.ConfigIntegrationName
+		}
+
+		// Set default values
+		attributes["use_existing_ad_application"] = true
+		attributes["application_id"] = "module.az_ad_application.application_id"
+		attributes["application_password"] = "module.az_ad_application.application_password"
+		attributes["service_principal_id"] = "module.az_ad_application.service_principal_id"
```
ipcrm, line 196: I think for these we should be able to leave them blank if there are not being overridden, based on the module source: https://github.com/lacework/terraform-azure-config/blob/main/main.tf#L10-L16

ipcrm (changes requested):
Looking good so far!  I made a few comments, I'm not sure all the module parameters required are included right now.  Those extra parameters are likely to require some expansion on the logic assigning attributes.  Happy to chat as needed!

lwgenerate/azure/azure.go
rmoles: ➕1 on adding the options from terraform registry inputs ☝️. This is what has been done for GCP

lwgenerate/azure/activity_log_with_config.tf
```diff
@@ -0,0 +1,43 @@
+terraform {
```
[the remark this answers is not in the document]
rmoles: ➕1 on this. Currently, AWS and GCP have the test examples inline in the test file. However, it may make sense to have these stored as files in a test resources dir.

lwgenerate/azure/activity_log_without_config.tf
```diff
@@ -0,0 +1,34 @@
...
+    }
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "=2.91.0"
+    }
+  }
+}
+
+provider "azuread" {
+}
```
rmoles, line 15: I'm wondering whether we need this provider? Looking at our TF registry module examples it looks like we don't use this in the examples.
Granted the customer-facing docs use this approach https://docs.lacework.com/azure-compliance-and-activity-log-integrations-terraform-using-azure-cloud-shell
It looks like we actually create an `az_ad_application` with each module. Perhaps we could change this to work similarly to the GCP examples, where we create the `Service Account` _(az_ad_application)_ in the config module and then re-use this in the activity log module via the config module outputs. eg.
```
provider "google" {}

provider "lacework" {}

module "gcp_organization_config" {
  source  = "lacework/config/gcp"
  version = "~> 1.0"

  org_integration = true
  organization_id = "my-organization-id"
}

module "gcp_organization_audit_log" {
  source  = "lacework/audit-log/gcp"
  version = "~> 2.0"

  bucket_force_destroy         = true
  org_integration              = true
  use_existing_service_account = true
  service_account_name         = module.gcp_organization_config.service_account_name
  service_account_private_key  = module.gcp_organization_config.service_account_private_key
  organization_id              = "my-organization-id"
  log_bucket                   = "lacework-logging-bucket"
}
```

rmoles: We likely should get the customer-facing docs for Azure to do something similar to this also
```
provider "azurerm" {
features {}
}

module "az_config" {
source  = "lacework/config/azure"
version = "~> 1.0"

}

module "az_activity_log" {
source  = "lacework/activity-log/azure"
version = "~> 1.0"

use_existing_ad_application = true
application_id              = module.az_config.application_id
application_password        = module.az_config.application_password
service_principal_id        = module.az_config.service_principal_id
}
```
Thoughts @ipcrm, @dmurray-lacework, @afiune, @marcosgm ?

rmoles: From talking to @afiune, it seems that the example that we have in the support docs has actually been a fairly recent update that was made by @marcosgm to make it more clear to the end users what is going on in the modules, rather than hiding away the app registration creation.
I still prefer the cleaner example suggested above, however, if we feel that the more verbose route shown in the docs is the right way to go, lets stick with what we have.

lwgenerate/azure/activity_log_with_config.tf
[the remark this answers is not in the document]
afiune: +1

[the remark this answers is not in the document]
djmctavish: Created a test-data folder under the azure folder to hold all these.

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,250 @@
...
+
+	if err != nil {
+		return nil, err
+	}
+
+	blocks = append(blocks, provider)
+	return blocks, nil
+}
+
+func createLaceworkAZADModule() ([]*hclwrite.Block, error) {
```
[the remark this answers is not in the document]
djmctavish: Done

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,250 @@
...
+
+	provider, err := lwgenerate.NewProvider(
+		"azurerm",
+	).ToBlock()
+
+	if err != nil {
+		return nil, err
+	}
+	// Create the features block
+	featuresBlock, err := lwgenerate.HclCreateGenericBlock("features", []string{}, attrs)
```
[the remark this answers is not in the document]
djmctavish: This is how it is used in the all the examples, 

```
provider "azurerm" {
  features = {}
}
```
I will add comment to the code to make this clear

lwgenerate/azure/azure.go
djmctavish: Again this is in the example, we create a an Active Directory application and reference that, I have added a flag for the user to use an existing AD. The terraform code will be generated accordingly depending on user choice

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,250 @@
...
+
+	// If Config is true, give the user the opportunity to name their integration. Defaults to "TF Config"
+	ConfigIntegrationName string
+
+	// If ActivityLog is true, give the user the opportunity to name their integration. Defaults to "TF activity log"
+	ActivityLogIntegrationName string
+}
+
+// Ensure all combinations of inputs are valid for supported spec
+func (args *GenerateAzureTfConfigurationArgs) validate() error {
```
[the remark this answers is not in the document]
djmctavish: I have updated this as I have added more options for configurations, and the validity of these is tested here now

lwgenerate/azure/azure.go
djmctavish: Yeah I have updated this to mirror more of the options that we require, but still able to create the 'no touch' terraform as per the help files example if the user does not set any config

lwgenerate/azure/activity_log_without_config.tf
djmctavish: Yeah, I have made it so that if the user does not enter any config then it will generate that terraform as per the support docs, but the user can enter details for an  existing AD, Resource Group etc

lwgenerate/azure/azure.go
djmctavish: I have changed this to use the created AD application if the user does not want to configure for a current AD

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,460 @@
...
+}
+
+type AzureTerraformModifier func(c *GenerateAzureTfConfigurationArgs)
+
+// NewTerraform returns an instance of the GenerateAzureTfConfigurationArgs struct with the provided enabled
+// settings (config/activity log).
+//
+// Note: Additional configuration details may be set using modifiers of the AzureTerraformModifier type
+//
+func NewTerraform(enableConfig bool, enableActivityLog bool, mods ...AzureTerraformModifier) *GenerateAzureTfConfigurationArgs {
```
ipcrm, line 95: When I run this without any modifiers (i.e., `azure.NewTerraform(true, true).Generate()`), I get: 

```
invalid inputs: Active directory details must be set
exit status 1
```

I believe we should make the AD settings a required input for the constructor since it cannot work without this value being set.

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,460 @@
...
+
+// WithAuditLogIntegrationName Set the Config Integration name to be displayed on the Lacework UI
+func WithAuditLogIntegrationName(name string) AzureTerraformModifier {
+	return func(c *GenerateAzureTfConfigurationArgs) {
+		c.ActivityLogIntegrationName = name
+	}
+}
+
+// WithAdIntegration Set the Config Integration name to be displayed on the Lacework UI
+func WithAdIntegration(enableAdIntegration bool) AzureTerraformModifier {
```
ipcrm, line 118: If I set this to false I get the following error:

```
invalid inputs: Active directory details must be set
```

Perhaps this shouldn't accept a value and just always be true when used?

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,460 @@
...
+	AllSubscriptions bool
+
+	// Storage Account name
+	StorageAccountName string
+
+	// Storage Account Resource Group
+	StorageAccountResourceGroup string
+
+	// Should we use existing storage account
+	ExistingStorageAccount bool
```
ipcrm, line 53: For these settings, can we just use the presence of a value instead of having these `bool` values as well?

i.e., if `ManagementGroupId` is not `""`; then we should use it.  Or for `AllSubscriptions`, if `SubscriptionIds` is empty then all subs get set.  

If not, all of that logic is going to have to be handled further 'downstream' in the CLI code when we survey the end-user.

ipcrm (commented):
Looking great!  Had a few questions/thoughts

lwgenerate/azure/azure.go
djmctavish: I think this is correct, the user can choose to create an Active Directory application, this adds hcl into the terraform to do so, if the do not use the create then they must supply the active directory attributes for an already existing application

lwgenerate/azure/azure.go
djmctavish: It is the case that the user must supply active directory attributes if this is set to false, these are 

```
	// Active Directory application Id
	AdApplicationId string

	// Active Directory password
	AdApplicationPassword string

	// Active Directory Enterprise app object id
	AdServicePrincipalId string
```
So if these are not set then the user should me made aware of the error

lwgenerate/azure/azure.go
djmctavish: Yes, certainly in the case of the `ManagementGroup` and `ExistingStorageAccount` flags, then the presence of the elements set will indicate if the the flag was set or not. The `AllSubscriptions` flag is a little different as it is set in the hcl for the config and activity log to indicate that all subscriptions within the selected tenant have read access, if this is not set it will only use the primary subscription (unless a list of subscriptions is given)

lwgenerate/azure/azure.go
ipcrm: I guess my confusion is that this is a constructor, and AD settings are required for this code to work (if I'm understanding correctly); so therefore I would think the AD settings would be a required 3rd property?  The whole point (IMO) of the constructor is to make using this large struct easier, and I would think the constructor should guide the developer to a working configuration.

lwgenerate/azure/azure.go
ipcrm: OK; so seems like this is an extension of the conversation here: https://github.com/lacework/go-sdk/pull/682#discussion_r805882285

lwgenerate/azure/azure.go
ipcrm: I'm certainly happy with leaving it as-is if that's beneficial to your other existing code.  The only reason I raised it is that these `bool` properties had caused myself and @rmoles some challenges previously (in the interactive CLI code).

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,460 @@
...
+
+	// Azure region where the storage account for loggin resides
```
rmoles, line 55:
```suggestion
	// Azure region where the storage account for logging resides
```

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,460 @@
...
+
+// Ensure all combinations of inputs are valid for supported spec
+func (args *GenerateAzureTfConfigurationArgs) validate() error {
+	// Validate one of config or activity log was enabled; otherwise error out
+	if !args.ActivityLog && !args.Config {
+		return errors.New("audit log or config integration must be enabled")
+	}
+
+	// Validate that active directory settings are correct
+	if !args.AdIntegration {
```
rmoles, line 67: Should this be a not check on `arg.AdIntegration`?

Additionally, it seems to me that this could be flattened to a single if
```
if args.AdIntegration && (args.AdApplicationId == "" || args.AdServicePrincipalId == "" || args.AdApplicationPassword == "")
```

lwgenerate/azure/azure.go
rmoles: I'm somewhat confused by `WithAdIntegration`
Is this intended to cover the `use_existing_ad_application` setting from the Terraform registry?
Or is this simply asking the user if they want to explicitly show the azure ad module in the top-level terraform code?

If the latter, why are we requiring `AdApplicationId`, `AdApplicationPassword`, `AdServicePrincipalId`.
If the former, should we rename this setting to `use_existing_ad_application` instead?

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,460 @@
...
+// WithSubscriptionIds List of subscriptions to to enable logging
+func WithSubscriptionIds(subscriptionIds []string) AzureTerraformModifier {
+	return func(c *GenerateAzureTfConfigurationArgs) {
+		c.SubscriptionIds = subscriptionIds
+	}
+}
+
+// WithAllSubscription Grant read access to ALL subscriptions within
+// the selected Tenant (overrides 'subscription_ids')
+func WithAllSubscription(allSubscriptions bool) AzureTerraformModifier {
```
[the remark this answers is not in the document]
djmctavish: Every little helps!

lwgenerate/azure/azure.go
djmctavish: Good spot

lwgenerate/azure/azure.go
djmctavish: No in this case AdIntegration means create an AD integration and use the generated details from terraform, probably should change the name, in the case where this is not set then we need to enter the details of an already existing AD integration. 
Could make it a single if statement

lwgenerate/azure/azure.go
djmctavish: I have changed this to be `CreateAdIntegration` as that is what it is, with this set we add in terraform to create an AD Application and we put in the details that are returned from the terraform which is 
`AdApplicationId`, `AdApplicationPassword`, `AdServicePrincipalId` if we do not want to create an AD Application then we need to reference an already existing one and the user must supply these attributes

lwgenerate/azure/azure.go
djmctavish: Made this a constructor now

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,458 @@
...
+
+// Ensure all combinations of inputs are valid for supported spec
+func (args *GenerateAzureTfConfigurationArgs) validate() error {
+	// Validate one of config or activity log was enabled; otherwise error out
+	if !args.ActivityLog && !args.Config {
+		return errors.New("audit log or config integration must be enabled")
+	}
+
+	// Validate that active directory settings are correct
+	if !args.CreateAdIntegration && (args.AdApplicationId == "" || args.AdServicePrincipalId == "" || args.AdApplicationPassword == "") {
```
rmoles, line 67: I'm wondering whether this arg should actually be inverted and be `useExistingAdApplication` ?
That aligns more closely with the TF module vars.

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,458 @@
...
+	}
+
+	// Validate the Mangement Group
+	if args.ManagementGroup && args.ManagementGroupId == "" {
+		return errors.New("When Group Management is enabled, then Group Id must be configured")
+	}
+
+	// Validate Storage Account
+	if args.ExistingStorageAccount {
+		if args.StorageAccountName == "" || args.StorageAccountResourceGroup == "" {
```
rmoles, line 78: Similar to a previous comment. We could probably flatten this to a single if

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,458 @@
...
+}
+
+type AzureTerraformModifier func(c *GenerateAzureTfConfigurationArgs)
+
+// NewTerraform returns an instance of the GenerateAzureTfConfigurationArgs struct with the provided enabled
+// settings (config/activity log).
+//
+// Note: Additional configuration details may be set using modifiers of the AzureTerraformModifier type
+//
+func NewTerraform(enableConfig bool, enableActivityLog bool, createAdIntegration bool, mods ...AzureTerraformModifier) *GenerateAzureTfConfigurationArgs {
```
rmoles, line 93: Similar to a comment from @ipcrm. I'm not convinced that we need to have `createAdIntegration bool` as part of the constructor.
If we changed the var to `useExistingAdApplication` we can infer whether we need to include the AzureAd module in the TF

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,458 @@
...
+
+// WithStorageAccountResourceGroup The Resource Group for the existing Storage Account
+func WithStorageAccountResourceGroup(storageAccountResourceGroup string) AzureTerraformModifier {
+	return func(c *GenerateAzureTfConfigurationArgs) {
+		c.StorageAccountResourceGroup = storageAccountResourceGroup
+	}
+}
+
+// WithLocation The Azure region where storage account for logging is
+func WithLocation(location string) AzureTerraformModifier {
```
rmoles, line 195: IMO we should rename this to `WithStorageAccountLocation` seeing as this location is only applicable to the Storage Account

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,458 @@
...
+	// Create blocks
+	requiredProviders, err := createRequiredProviders()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to generate required providers")
+	}
+
+	azureADProvider, err := createAzureADProvider()
+	if err != nil {
+		return "", errors.Wrap(err, "failed to generate AD provider")
+	}
```
rmoles, line 217: Should we be calling this here and also creating the AzureAdProvider in `createRequiredProviders()`?

lwgenerate/azure/azure.go
rmoles: I see further down in the code why this is named this way. We are setting the TF var `use_existing_ad_application` to true in each scenario.

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,458 @@
...
+			attributes["application_id"] = args.AdApplicationId
+			attributes["application_password"] = args.AdApplicationPassword
+			attributes["service_principal_id"] = args.AdServicePrincipalId
+		}
+
+		// Only set subscription ids if all subscriptions flag is not set
+		if !args.AllSubscriptions {
+			if len(args.SubscriptionIds) > 0 {
+				attributes["subscription_ids"] = args.SubscriptionIds
+			}
```
rmoles, line 359: should we raise an error here if `!args.AllSubscriptions && len(args.SubscriptionIds) == 0`

rmoles: perhaps this block could be something like this:
```
if args.AllSubscriptions {
    // Set Subscription information
    attributes["all_subscriptions"] = args.AllSubscriptions
} else if len(args.SubscriptionIds) > 0 {
    attributes["subscription_ids"] = args.SubscriptionIds
} else {
    return an error here....
}
```

lwgenerate/azure/test-data/activity-log-with-all-subscriptions.tf
```diff
@@ -0,0 +1,35 @@
+terraform {
+  required_providers {
+    azuread = {
+      source  = "hashicorp/azuread"
+      version = "~> 2.16"
+    }
+    azurerm = {
+      source  = "hashicorp/azurerm"
+      version = "=2.91.0"
```
rmoles, line 9: Any reason for us to be explicit with `"=2.91.0"` rather than approximate with `"~> 2.91"`?

lwgenerate/azure/azure.go
djmctavish: I am setting that to true in each occasion 
1) When we create the AD application, using the returned terraform attributes
2) When we supply details for an already existing AD integration

I was under the impression that there must be an AD for this to work

lwgenerate/azure/azure.go
djmctavish: Done

lwgenerate/azure/azure.go
djmctavish: His comment was that this should be a constructor, I have changed it be that rather than a modifier, do you think it should go back to what it was?

lwgenerate/azure/azure.go
djmctavish: Yeah, renamed the attribute to `StorageLocation` to make it clearer

lwgenerate/azure/azure.go
djmctavish: Don't think that this works here, we seem to return an error in the case where `AllSubscriptions` is set to `false`

If `AllSubscriptions` is set `true`, then it does not matter what is in the `SubscriptionIds` list, so we only need to look at this list in the case where it is `false`

lwgenerate/azure/azure.go
djmctavish: Don't think that this matters really, I made a create function for each individual section that is detailed in the example terraform from the help files

lwgenerate/azure/azure.go
```diff
@@ -0,0 +1,458 @@
...
+		} else {
+			// Set Subscription information
+			attributes["all_subscriptions"] = args.AllSubscriptions
+		}
```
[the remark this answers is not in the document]
djmctavish: Same comment as above, only when this is `false`, do we need to look at the `SubscriptionIds` list

lwgenerate/azure/azure.go
rmoles: I wouldn't expect that to be the behavior here. If all subscriptions is false, it will check if we have any subscription id's in the else if. Only if both all subscriptions is false and subscriptions id is empty should we hit the else and return an error.

rmoles: If all subscriptions is true it would just set that to true and skip the else if and else, right?

lwgenerate/azure/azure.go
rmoles: So that's not strictly true, but this relates to the first long comment that I left on this PR a while back. We have made the decision to show the Azure AD module at the top level. In reality, we can do the AD creation within the Config and/or the Activity log integration, similar to what we do in GCP.

Based on the decision made previously, I think what you have implemented here is correct, it just took me a bit to get the full picture...

lwgenerate/azure/azure.go
djmctavish: OK I see what you are getting at now, misunderstanding on my part. The thing is that we do not really need the else part. 
If `AllSubscriptions` is set `true`, it will use all and we only need to set the flag
If `AllSubscriptions` is set `false`, we may or may not have a list of `SubscriptionIds` , if we have we set them and they are used, if there are none then it will default to using the primary subscription for the account.

So in my mind we have the simpler logic that is in the original I think

rmoles: Fair enough. I was unaware that we had that fall-through option on the subscription id. In that case we're good as is

rmoles (approved):
All of my comments have been addressed.

Looks good to me. LGTM!

dmurray-lacework (approved):

Issue #292, opened 2021-01-12
Extend lacework agent command to be able to check agent status on given host


lw-mikeoc:
Would be great to query a host, or automatically query a list of all hosts with agents, to determine the run status of the datacollector. This could use the SSH approach used with the current install option or ideally leverage a new API that returns list of active hosts with and without agents installed.

afiune:
@lw-mikeoc Thank you for your feedback!

We are considering adding new APIs that will return the list of active hosts with and without agents installed. 🎉 Once those APIa are exposed we will add a new command like `lacework agent list` or similar.

afiune:
**JIRA: ALLY-293**

PR #684, opened 2022-02-03
feat: List all active hosts with an agent installed

afiune:
## Summary

Introducing a new command:
```
lacework agent list
```
This new command should list all active hosts with an agent installed in their Lacework account.

Installed a new agent and waited for it to be shown by the new command.

## Issue

Jira: https://lacework.atlassian.net/browse/ALLY-293
Closes https://github.com/lacework/go-sdk/issues/292


api/v2_search_filters.go
```diff
@@ -49,6 +51,6 @@ type Filter struct {
 }
 
 type TimeFilter struct {
-	StartTime string `json:"startTime,omitempty"`
-	EndTime   string `json:"endTime,omitempty"`
+	StartTime time.Time `json:"startTime,omitempty"`
+	EndTime   time.Time `json:"endTime,omitempty"`
```
dmurray-lacework, line 55: omitempty doesn't work with time. You'll need to use a pointer here if you'd like to omit the time value

api/entities.go
```diff
@@ -0,0 +1,49 @@
...
+//
+//   response := api.MachineDetailsResponse{}
+//   lacework.V2.Entities.Search(response, api.SearchFilter{})
+//
+func (svc *EntitiesService) Search(response interface{}, filters SearchFilter) error {
+	var apiPath string
+
+	switch response.(type) {
+	case *MachineDetailsResponse:
+		apiPath = fmt.Sprintf(apiV2EntitiesSearch, "MachineDetails")
```
dmurray-lacework, line 43: Would it make sense to create enums/const for "MachineDetails" and other entities we will be supporting?

api/entities_machine_details_test.go
```diff
@@ -0,0 +1,103 @@
...
+  "paging": {
+    "rows": 1,
+    "totalRows": 1,
+    "urls": {
+      "nextPage": null
+    }
+  }
+}
+	`
+}
```
dmurray-lacework, line 103: Might be good to include a test for json output of the searchFilter struct.
This test below will pass but i'm not sure this is the expected result given the omitempty's on the Start/End time fields

```
func TestEntities_JsonMarshall(t *testing.T) {
	searchFilter := api.SearchFilter{
		Filters: []api.Filter{{
			Expression: "eq",
			Field:      "tokenAlias",
			Value:      "alias",
		}},
		Returns: nil,
	}
	jsonSearchFilter, err := json.Marshal(searchFilter)

	assert.NoError(t, err)
	assert.Equal(t, mockSearchJson, string(jsonSearchFilter))
}

var mockSearchJson = `{"timeFilter":{"startTime":"0001-01-01T00:00:00Z","endTime":"0001-01-01T00:00:00Z"},"filters":[{"expression":"eq","field":"tokenAlias","value":"alias","values":null}],"returns":null}`
```

afiune:
Additionally, we need to cover the case where an account doesn't have any agent installed.
```
➜  go-sdk git:(afiune/ALLY-293/cmd/agent-list) lacework agent list --subaccount tech-ally-clean
  MID   SHORT AGENT TOKEN   HOSTNAME   NAME   IP ADDRESS   EXTERNAL IP   OS ARCH   LAST CHECKIN
------+-------------------+----------+------+------------+-------------+---------+---------------
```

api/v2.go
```diff
@@ -79,3 +88,78 @@ type V2Service interface {
...
+// 			break
+// 		}
+//
+// 		if pageOk {
+// 			continue
+// 		}
+// 		break
+// }
+// ```
+func (c *Client) NextPage(p Pagination) (bool, error) {
```
dmurray-lacework, line 138: I'm not sure about the placement of NextPage method on the Client.
I've been thinking a service could be Pageable and implement the NextPage()
Individual structs in that service could be PaginatedResponses and have functions to expose paging details.

api/_examples/pagination/main.go
```diff
@@ -0,0 +1,35 @@
...
+		// Output: Rows read: 21
+		fmt.Printf("Rows read: %d\n", response.Paging.Rows)
+
+		pageOk, err := lacework.NextPage(&response)
+		if err == nil && pageOk {
+			continue
+		}
+		break
+	}
+}
```
dmurray-lacework, line 35: Following on from our zoom I think i'd prefer an approach where we can call NextPage on the response.

```
	response, err := lacework.V2.Entities.ListMachineDetails()
	if err != nil {
		log.Fatal(err)
	}

	rowCount := response.PageInfo().Rows

	for {
		fmt.Printf("showing %v rows out of %v \n", rowCount, response.PageInfo().TotalRows)
		if response.HasNextPage() {
			continue
		}
		response.GetNextPage(&response)
		rowCount = rowCount + response.PageInfo().Rows
		break
	}
```

afiune: To do this, **every single response** would have to have a copy of the client (that is `*api.Client`).

Are we ok with that?

dmurray-lacework: No I don't think that would be a good idea. Could we shift a NextPage implementation to the entities services?

afiune: I thought about it too, why do you want that move? 

If we do that, then we won't have logging at the service level because `Client.log` is private. Either we make it public or we don't have logging. (we can definitely move it, there are just these tradeoffs)

api/entities.go
```diff
@@ -0,0 +1,83 @@
...
+//   )
+//   lacework.V2.Entities.Search(response, filters)
+//
+func (svc *EntitiesService) Search(response interface{}, filters SearchFilter) error {
+	var apiPath string
+
+	switch response.(type) {
+	case *MachineDetailsEntityResponse:
+		apiPath = fmt.Sprintf(apiV2EntitiesSearch, EntityTypes[MachineDetailsEntityType])
+
```
dmurray-lacework, line 71: Is the spacing between case statements personal preference? If so perhaps we should add it to linter settings. I've no strong but I've seen this pop on PR's and usually recommend to not space out between cases.

api/entities_machine_details.go
```diff
@@ -0,0 +1,105 @@
...
+		Name           string `json:"Name"`
+		InstanceID     string `json:"InstanceId"`
+		InternalIP     string `json:"InternalIp"`
+		LwTokenShort   string `json:"LwTokenShort"`
+		SubnetID       string `json:"SubnetId"`
+		VMInstanceType string `json:"VmInstanceType"`
+		VMProvider     string `json:"VmProvider"`
+		VpcID          string `json:"VpcId"`
+		Zone           string `json:"Zone"`
+		Arch           string `json:"arch"`
```
dmurray-lacework, line 102: The cases of the json tags are all different. I'm guessing this is because of the api?

cli/cmd/agent_list.go
```diff
@@ -0,0 +1,208 @@
...
+		Filters []string
+	}{}
+
+	agentListCmd = &cobra.Command{
+		Use:     "list",
+		Aliases: []string{"ls"},
+		Short:   "List all hosts with a running agent",
+		Long: `List all hosts that have a running agent in your environment.
+
+You can use 'key:value' pairs to filter the list of hosts with the --filter flag.
```
dmurray-lacework, line 48: Reading this description, it sounds like i should not be able to pass in a value without a `:` however this runs without errpr

```
❯ lacework agent list --filter invalid
- Fetching list of agents with filters (invalid)... 
```

cli/cmd/agent_list.go
```diff
@@ -0,0 +1,208 @@
...
+					"tag", tag, "expected_format", "key:value",
+				)
+				continue
+			}
+
+			cli.Log.Infow("adding filter", "key", kv[0], "value", kv[1])
+			filters.Filters = append(filters.Filters, api.Filter{
+				Field:      kv[0],
+				Expression: cli.lqlOperator, // @afiune we use rlike to allow user to pass regex
+				Value:      kv[1],
```
dmurray-lacework, line 100: Should we check for empty string here?

ERROR unable to list agents via MachineDetails search: 
  [POST] https://customerdemo.lacework.net/api/v2/Entities/MachineDetails/search
  [400] Invalid filters: expression 'rlike' requires non-empty 'value' field

dmurray-lacework (approved):
Approved, but there are a few comment that may need cleared up

api/entities_machine_details.go
afiune: Unfortunately, yes.

PR #687, opened 2022-02-04
fix(lwgenerate): support lists for module attributes

ipcrm:
## Summary
Prior to this commit, module attributes could not contain lists. This
adds support for both string lists as well as lists of interfaces.

See test/example changes.

## Issue

Supports ALLY-394


djmctavish (approved):
I have used this latest version in my tests for generating the Azure terraform lists and can confirm that this creates the correct list elements

dmurray-lacework (approved):

rmoles (approved):
Looks good. Thanks @ipcrm

PR #691, opened 2022-02-08
refactor(cli): Refactor generate & generate_aws

rmoles:

## Summary

The generate.go & generate_aws.go files are geared towards AWS only.

This PR aims to refactor `generate.go` to be more generic to work with GCP & Azure. 
This PR also moves AWS specific vars to the `generate_aws.go` file

make test ✅ 
make fmt ✅ 

## Issue

https://lacework.atlassian.net/browse/ALLY-865

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -5,12 +5,9 @@ import (
 	"os"
 	"path/filepath"
 	"regexp"
-	"strings"
 	"time"
 
 	"github.com/AlecAivazis/survey/v2"
-	"github.com/imdario/mergo"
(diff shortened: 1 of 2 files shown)


djmctavish:
This looks much cleaner Ross, I can see without this how adding Azure and GCP would increase the clutter.

I was wondering would we want to have the same `executionPreRunChecks` call for all 3 or would we want to be able to differentiate between and do something specific for each rather than just call `TerraformExecutePreRunCheck` ?

djmctavish (approved):
LGTM

dmurray-lacework (approved):

PR #692, opened 2022-02-10
fix(cli): remove text wrapping on list-sources output

dmurray-lacework:
## Summary

Fix formatting on table output. Remove text wrapping.

`lacework query list-sources`

## Issue
https://lacework.atlassian.net/browse/ALLY-863

### DIFF
--- a/cli/cmd/lql_sources.go
+++ b/cli/cmd/lql_sources.go
@@ -65,12 +65,6 @@ func getListQuerySourcesTable(datasources []api.Datasource) (out [][]string) {
 
 func listQuerySources(_ *cobra.Command, args []string) error {
 	cli.Log.Debugw("retrieving LQL data sources")
-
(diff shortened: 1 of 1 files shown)


cli/cmd/lql_sources.go
```diff
@@ -65,6 +66,11 @@ func getListQuerySourcesTable(datasources []api.Datasource) (out [][]string) {
 func listQuerySources(_ *cobra.Command, args []string) error {
 	cli.Log.Debugw("retrieving LQL data sources")
 
+	tableSettings := tableFunc(func(t *tablewriter.Table) {
+		t.SetAutoWrapText(false)
+		t.SetBorder(false)
+	})
+
```
afiune, line 73: Why do we define these settings all the way up here? Why not just nested right when we use them below?

afiune (approved):
Code looks good but.....

tenor-126586777

PR #693, opened 2022-02-10
feat(policies): add tags support

hazedav:
## Summary
As a Lacework user I want to use the CLI to interact (CRU) with policy tags

- Add list-tags command
- Add --tag filter for policy list
- Add tag to appropriate structs

Automated unit and integration testing

## Issue
https://lacework.atlassian.net/browse/ALLY-878

## Examples
List policies with a specific tag
```
lacework policy list --tag control:1.1 
```

List all policy tags
```
lacework policy list-tags                         
```

### DIFF
--- a/api/policy.go
+++ b/api/policy.go
@@ -35,19 +35,20 @@ type PolicyService struct {
 var ValidPolicySeverities = []string{"critical", "high", "medium", "low", "info"}
 
 type NewPolicy struct {
-	EvaluatorID   string `json:"evaluatorId,omitempty" yaml:"evaluatorId,omitempty"`
-	PolicyID      string `json:"policyId,omitempty" yaml:"policyId,omitempty" `
-	PolicyType    string `json:"policyType" yaml:"policyType"`
-	QueryID       string `json:"queryId" yaml:"queryId"`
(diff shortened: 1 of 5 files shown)


api/policy.go
```diff
@@ -56,45 +57,61 @@ This would prevent someone from toggling something to disabled or 0 respectively
...
+	LastUpdateUser string   `json:"lastUpdateUser"`
+}
+
+func (p *Policy) HasTag(t string) bool {
+	for _, v := range p.Tags {
+		if t == v {
+			return true
+		}
+	}
+	return false
```
dmurray-lacework, line 102: Could this be condensed```suggestion
			return  t == v
	}
```

Comment (dmurray-lacework) [inline api/policy.go:102]: could we use `ContainsStr` from go-sdk/internal/array/contains.go here?
```suggestion
return array.ContainsStr(p.Tags, t)
```

dmurray-lacework (approved):
Approved with one suggestion

api/policy.go
hazedav: Thanks @dmurray-lacework , admittedly I wasn't aware we had this in our extended library.

hazedav: Fixed.

PR #712, opened 2022-02-22
refactor(cli): Create cloud specific TF file

rmoles:

## Summary
Currently, the default output location for the Terraform generation is `~/lacework/main.tf`
We should create a file for each cloud to ensure we don't override TF from another cloud.

With this change, we can pass any filename to the output location
eg. 
```
~/lacework/aws.tf
~/lacework/azure.tf
~/lacework/gcp.tf
```

make test ✅ 
run dev version of CLI and confirmed output location ✅ 
make fmt ✅ 

## Issue
https://lacework.atlassian.net/browse/ALLY-890

### DIFF
--- a/cli/cmd/generate.go
+++ b/cli/cmd/generate.go
@@ -196,7 +196,7 @@ func determineOutputDirPath(location string) (string, error) {
 }
 
 // writeHclOutputPreCheck Prompt for confirmation if main.tf already exists; return true to continue
-func writeHclOutputPreCheck(outputLocation string, cloud string) (bool, error) {
+func writeHclOutputPreCheck(outputLocation string, filename string) (bool, error) {
 	// If noninteractive, continue
 	if !cli.InteractiveMode() {
 		return true, nil
@@ -207,12 +207,11 @@ func writeHclOutputPreCheck(outputLocation string, cloud string) (bool, error) {
 		return false, err
 	}
 
-	fileName := cloud
-	if fileName == "" {
-		fileName = "main"
+	if filename == "" {
+		filename = "main"
 	}
 
-	hclPath := filepath.FromSlash(fmt.Sprintf("%s/%s.tf", outputDir, fileName))
+	hclPath := filepath.FromSlash(fmt.Sprintf("%s/%s.tf", outputDir, filename))
 
 	// If the file doesn't exist, carry on
 	if _, err := os.Stat(hclPath); os.IsNotExist(err) {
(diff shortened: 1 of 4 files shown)


cli/cmd/generate.go
```diff
@@ -196,7 +196,7 @@ func determineOutputDirPath(location string) (string, error) {
 }
 
 // writeHclOutputPreCheck Prompt for confirmation if main.tf already exists; return true to continue
-func writeHclOutputPreCheck(outputLocation string) (bool, error) {
+func writeHclOutputPreCheck(outputLocation string, cloud string) (bool, error) {
```
dmurray-lacework, line 199: nit: do we what to call this cloud? I believe on the call yesterday we discussed some examples of TF that wasn't creating cloud infra.

cli/cmd/generate.go
```diff
@@ -207,7 +207,12 @@ func writeHclOutputPreCheck(outputLocation string) (bool, error) {
 		return false, err
 	}
 
-	hclPath := filepath.FromSlash(fmt.Sprintf("%s/main.tf", outputDir))
+	fileName := cloud
```
dmurray-lacework, line 210: Any need for reassignment here. Maybe filename could be used as the arg name instead of cloud.

cli/cmd/generate.go
```diff
@@ -245,8 +250,13 @@ func writeHclOutput(hcl string, location string) (string, error) {
 		}
 	}
 
+	fileName := cloud
```
dmurray-lacework, line 253: Same as comment above

cli/cmd/generate_aws.go
```diff
@@ -117,7 +117,7 @@ This command can also be run in noninteractive mode. See help output for more de
 			}
 
 			// Write-out generated code to location specified
-			dirname, location, err := writeGeneratedCodeToLocation(cmd, hcl)
+			dirname, location, err := writeGeneratedCodeToLocation(cmd, hcl, "aws")
```
dmurray-lacework, line 120: Wondering if it would be a good pattern to have the file output name declared as a const at the top of each generate_something file? No strong feelings on this, just a thought.

cli/cmd/generate.go
rmoles: fair point. Will update

cli/cmd/generate_aws.go
rmoles: I had considered that. Didn't feel like making it a CONST gave us much so decided to go with this approach. I have no strong feelings on this either, so happy enough to change if others feel strongly about it.

PR #714, opened 2022-02-28
fix(cli): delete listed twice on help

hazedav:
## Summary
Policies (LQL): Delete registered twice resulting in strange help output

```
lacework policy --help
Available Commands:
  create      Create a policy
  delete      Delete a policy
  delete      Delete a policy
  list        List policies
  list-tags   List policy tags
  show        Show policy
  update      Update a policy
```

Manually

## Issue
https://lacework.atlassian.net/browse/ALLY-900

### DIFF
--- a/cli/cmd/policy.go
+++ b/cli/cmd/policy.go
@@ -132,7 +132,6 @@ func init() {
 	policyCmd.AddCommand(policyListCmd)
 	policyCmd.AddCommand(policyListTagsCmd)
 	policyCmd.AddCommand(policyShowCmd)
-	policyCmd.AddCommand(policyDeleteCmd)
 
 	// policy list specific flags
 	policyListCmd.Flags().StringVar(


dmurray-lacework (approved):

cli/cmd/policy.go
```diff
@@ -132,7 +132,6 @@ func init() {
 	policyCmd.AddCommand(policyListCmd)
 	policyCmd.AddCommand(policyListTagsCmd)
 	policyCmd.AddCommand(policyShowCmd)
-	policyCmd.AddCommand(policyDeleteCmd)
```
hazedav, line 135: This is already registered in `policy_delete.go`.

afiune:
@hazedav Could we add a test, please? (like this https://github.com/lacework/go-sdk/blob/main/integration/help_test.go#L132-L166)

hazedav:
> @hazedav Could we add a test, please? (like this https://github.com/lacework/go-sdk/blob/main/integration/help_test.go#L132-L166)

Is this a standard test pattern that every CLI command asserts its canonical help output?  I see exactly 1 sub command doing this...

PR #716, opened 2022-03-01
feat(cli): generate & execute gcp terraform code

rmoles:
## Summary

Enable the automated generation & execution of Terraform code for integrating Lacework with GCP

## How did you test this change
generated dev version of CLI and run manually ✅ 
Unit tests (pending)
Integration tests (pending)

## Issue
https://lacework.atlassian.net/browse/ALLY-393


ipcrm (commented):
@rmoles looking good!

I'll take a further look at the code itself but I hit a problem pretty early on.

When I ran through the config I just said yes to integrating both config & auditLog and gave it no other information because it said they were optional: 

```
╷
│ Error: project: required field is not set
│
│   with module.gcp_project_audit_log.data.google_project.selected,
│   on .terraform/modules/gcp_project_audit_log/main.tf line 54, in data "google_project" "selected":
│   54: data "google_project" "selected" {
│
╵
```

FWIW; here was my cached inputs:

```json
{
  "data": {
    "AuditLog": true,
    "Config": true,
    "ServiceAccountCredentials": "",
    "OrganizationIntegration": false,
    "GcpOrganizationId": "",
    "GcpProjectId": "",
    "ExistingServiceAccount": null,
    "ConfigIntegrationName": "",
    "AuditLogLabels": null,
    "BucketLabels": null,
    "PubSubSubscriptionLabels": null,
    "PubSubTopicLabels": null,
    "BucketRegion": "",
    "BucketLocation": "",
    "BucketName": "",
    "ExistingLogBucketName": "",
    "ExistingLogSinkName": "",
    "EnableForceDestroyBucket": false,
    "EnableUBLA": false,
    "LogBucketLifecycleRuleAge": -1,
    "LogBucketRetentionDays": 0,
    "AuditLogIntegrationName": "",
    "LaceworkProfile": ""
  },
  "expires_at": "2022-03-02T09:12:42.49619-05:00"
}
```

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+			//// set to nil if LogBucketLifecycleRuleAge value is -9999
+			//// -9999 is a dummy default value used to verify if the user has provided a value
+			//bucketLifecycleRuleAge, err := cmd.Flags().GetInt("bucket_lifecycle_rule_age")
+			//if err != nil {
+			//	return errors.Wrap(err, "failed to load command flags")
+			//}
+			//
+			//if bucketLifecycleRuleAge == 99999 && *GenerateGcpCommandState.LogBucketLifecycleRuleAge == 99999 {
+			//	nilSetter(GenerateGcpCommandState.LogBucketLifecycleRuleAge)
+			//}
```
ipcrm, line 196: nit: maybe some leftover commented code

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+
+			// Validate gcp sa credentials file, if passed
+			gcpSaCredentials, err := cmd.Flags().GetString("service_account_credentials")
+			if err != nil {
+				return errors.Wrap(err, "failed to load command flags")
+			}
+			if err := validateServiceAccountCredentialsFile(gcpSaCredentials); gcpSaCredentials != "" && err != nil {
```
ipcrm, line 203: could `validateServiceAccountCredentialsFile` be updated to just return nil if `credFile` is ""?

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+				return err
+			}
+
+			// Validate gcp region, if passed
+			region, err := cmd.Flags().GetString("bucket_region")
+			if err != nil {
+				return errors.Wrap(err, "failed to load command flags")
+			}
+			if err := validateGcpRegion(region); region != "" && err != nil {
```
ipcrm, line 212: could the `region != ""` check be moved into `validateGcpRegion`?

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+		&GenerateGcpExistingServiceAccountDetails.PrivateKey,
+		"existing_service_account_private_key",
+		"",
+		"specify existing service account key (base64 encoded)")
+	generateGcpTfCommand.PersistentFlags().StringVar(
+		&GenerateGcpCommandState.ConfigIntegrationName,
+		"config_integration_name",
+		"",
+		"specify a custom config integration name")
+	// TODO: Implement AuditLogLabels, BucketLabels, PubSubSubscriptionLabels & PubSubTopicLabels
```
ipcrm, line 314: @rmoles still planning on this in this PR or should we raise an additional issue?

djmctavish: Sounds reasonable I am not sure the initial implementation would need all the 'i's dotted and the 't's crossed

cli/cmd/generate.go
```diff
@@ -30,6 +30,30 @@ func (a *AwsGenerateCommandExtraState) writeCache() {
 	}
 }
 
+type GcpGenerateCommandExtraState struct {
```
djmctavish, line 33: Would it be better to have the extra state definition and associated function in the same file as the generate args definition, eg in the lwgenerate/gcp.go file rather than the 'generic' generate.go file

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+		"specify a custom audit log integration name")
+
+}
+
+func validateServiceAccountCredentialsFile(credFile string) error {
+	if file.FileExists(credFile) {
+		jsonFile, err := os.Open(credFile)
+		// if we os.Open returns an error then handle it
+		if err != nil {
+			return errors.Wrap(err, "Issue opening GCP credentials file")
```
afiune, line 373: Nit: We usually start error messages with lowercases.

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+		}
+		defer jsonFile.Close()
+
+		byteValue, _ := ioutil.ReadAll(jsonFile)
```
afiune, line 377: Why are we ignoring this error?

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+			return errors.New("Invalid GCP Service Account credentials file. " +
+				"The private_key and client_email fields MUST be present. private_key must be base64 encoded")
+		}
+
+	}
+	return nil
+}
+
+func validateSaCredFileContent(credFileContent map[string]interface{}) bool {
+	if credFileContent["private_key"] != nil && credFileContent["client_email"] != nil {
```
afiune, line 395: Maybe unit test this function? I have a feeling that if the file does not have these fields and we try to access them this wey, we will panic.

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+		err := validateStringIsBase64(credFileContent["private_key"].(string))
```
afiune, line 396: When casting, always try to "try" with `str, ok :=` - if not, we will panic

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+}
+
+// create survey.Validator for string is base64
+func validateStringIsBase64(val interface{}) error {
+	switch value := val.(type) {
+	case string:
+		// if value isn't base64, return error
+		_, err := base64.StdEncoding.DecodeString(value)
+		if err != nil {
+			return errors.Wrap(err, "provided private key is not base64 encoded")
```
afiune, line 411: Here an example of an error message starting with lowercases, lets update all errors to follow this pattern

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+		},
+		{
+			Prompt:   &survey.Input{Message: QuestionExistingServiceAccountPrivateKey, Default: config.ExistingServiceAccount.PrivateKey},
+			Response: &config.ExistingServiceAccount.PrivateKey,
+			Opts:     []survey.AskOpt{survey.WithValidator(survey.Required), survey.WithValidator(validateStringIsBase64)},
+		}}); err != nil {
+		return err
+	}
+
+	return nil
```
afiune, line 575: We do not need the whole if statement above, you can just execute the function and return the err variable. It will be nil or not, no need to check for nil.

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+		{
+			Prompt:   &survey.Input{Message: QuestionGcpAuditLogIntegrationName, Default: config.AuditLogIntegrationName},
+			Checks:   []*bool{&config.AuditLog},
+			Required: true,
+			Response: &config.AuditLogIntegrationName,
+		}}); err != nil {
+		return err
+	}
+
+	return nil
```
afiune, line 595: Same, above the if statement. Just return err.

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+// entry point for launching a survey to build out the required generation parameters
+func promptGcpGenerate(
+	config *gcp.GenerateGcpTfConfigurationArgs,
+	existingServiceAccount *gcp.ExistingServiceAccountDetails,
+	extraState *GcpGenerateCommandExtraState,
+) error {
+	// Cache for later use if generation is abandon and in interactive mode
+	if cli.InteractiveMode() {
+		defer writeGcpGenerationArgsCache(config)
+		defer extraState.writeCache()
```
afiune, line 706: Why these two state structures have different cache functions, this one in embedded and the other one above not?

afiune (commented):
Almost there, a few nitpicks! Great work, Ross.

cli/cmd/generate_gcp.go
ipcrm: It's because there is a need to do further validation on the generation args. Inside `writeGcpGenerationArgsCache` there's a check/validation that happens to ensure all required structs are populated enough to actually cache (so we don't end up in weird states when resuming).  

(https://github.com/lacework/go-sdk/blob/c11b799441b5e641984e12781f292e9f8281640d/cli/cmd/generate_gcp.go#L688)

cli/cmd/generate_gcp.go
rmoles: I think this was actually already covered in the function, I just forgot to update this.

cli/cmd/generate_gcp.go
rmoles: good point

cli/cmd/generate_gcp.go
rmoles: About to push some unit tests. At the moment it seems like it handles trying to access the key without panicing

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,780 @@
...
+	default:
+		// if the value passed is not a string
+		return errors.New("value must be a string")
+	}
+
+	return nil
+}
+
+// survey.Validator for gcp region
+func validateGcpRegion(val interface{}) error {
```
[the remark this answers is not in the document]
rmoles: I haven't had any issues with this so far

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,774 @@
...
+	generateGcpTfCommand.PersistentFlags().IntVar(
+		&GenerateGcpCommandState.LogBucketRetentionDays,
+		"bucket_retention_days",
+		0,
+		"specify the bucket retention days")
+	generateGcpTfCommand.PersistentFlags().StringVar(
+		&GenerateGcpCommandState.AuditLogIntegrationName,
+		"audit_log_integration_name",
+		"",
+		"specify a custom audit log integration name")
```
djmctavish, line 353: I think that we need a flag to be able to run the terraform in non-interactive mode like "--terraform-apply" etc

rmoles: good catch. Will add this in

afiune:
@rmoles we had some issues with the CI pipelines, everything should be back online now.

I am re-running your pipeline and starting review of the new changes. 💯

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,787 @@
...
+	"github.com/AlecAivazis/survey/v2"
+	"github.com/pkg/errors"
+
+	"github.com/lacework/go-sdk/internal/file"
+	"github.com/lacework/go-sdk/lwgenerate/gcp"
+)
+
+var (
+	// Define question text here to be reused in testing
+	QuestionGcpEnableConfig            = "Enable Config Integration?"
```
afiune, line 24: Based on our Slack message https://lacework.slack.com/archives/CUTNH5QKY/p1646692456149189

@Karmali how does this sound? 🤔 `Enable compliance integration?` - Does that sound accurate?

cc @marcosgm 
```suggestion
	QuestionGcpEnableConfig            = "Enable compliance integration?"
```

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,787 @@
...
+	QuestionGcpEnableAuditLog          = "Enable AuditLog Integration?"
```
afiune, line 25: Based on Google's doc, it is Audit Logs - https://cloud.google.com/logging/docs/audit

```suggestion
	QuestionGcpEnableAuditLog          = "Enable Audit Logs integration?"
```

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,787 @@
...
+	QuestionGcpOrganizationIntegration = "Organization Integration?"
```
afiune, line 26:
```suggestion
	QuestionGcpOrganizationIntegration = "Organization integration?"
```

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,787 @@
...
+	QuestionGcpOrganizationID          = "Specify the GCP Organization ID:"
+	QuestionGcpProjectID               = "Specify the Project ID to be used to provision Lacework resources:"
```
afiune, line 28:
```suggestion
	QuestionGcpProjectID               = "Specify the project ID to be used to provision Lacework resources:"
```

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,787 @@
...
+	QuestionGcpServiceAccountCredsPath = "Specify Service Account credentials JSON path: (optional)"
```
afiune, line 29:
```suggestion
	QuestionGcpServiceAccountCredsPath = "Specify service account credentials JSON path: (optional)"
```

afiune: I think you get the pattern. We need to talk with product and docs team to check that these end-user messages are correct.

I am going to update the AWS generate command to AVOID using capitalized words where they are not needed.

afiune:
Real failure from your code:

Screen Shot 2022-03-08 at 2 52 16 PM

cli/cmd/generate_gcp.go
```diff
@@ -0,0 +1,787 @@
...
+* Prompt for the required information to setup the integration
+* Generate new Terraform code using the inputs
+* Optionally, run the generated Terraform code:
+  * If Terraform is already installed, the version will be confirmed suitable for use
+	* If Terraform is not installed, or the version installed is not suitable, a new version will be installed into a temporary location
+	* Once Terraform is detected or installed, Terraform plan will be executed
+	* The command will prompt with the outcome of the plan and allow to view more details or continue with Terraform apply
+	* If confirmed, Terraform apply will be run, completing the setup of the cloud account
+
+This command can also be run in noninteractive mode. See help output for more details on supplying required values for generation.
```
afiune, line 85: We need to cut these lines at a maximum of 80-90 chars per line, just so that it is easy to display in any terminal, for example:
```
$ lacework  configure -h
Configure settings that the Lacework CLI uses to interact with the Lacework
platform. These include your Lacework account, API access key and secret.

To create a set of API keys, log in to your Lacework account via WebUI and
navigate to Settings > API Keys and click + Create New. Enter a name for
the key and an optional description, then click Save. To get the secret key,
download the generated API key file.

Use the flag --json_file to preload the downloaded API key file.

If this command is run with no flags, the Lacework CLI will store all
settings under the default profile. The information in the default profile
is used any time you run a Lacework CLI command that doesn't explicitly
specify a profile to use.

You can configure multiple profiles by using the --profile flag. If a
config file does not exist (the default location is ~/.lacework.toml),
the Lacework CLI will create it for you.

Usage:
  lacework configure [flags]
  lacework configure [command]

Available Commands:
  list           List all configured profiles at ~/.lacework.toml
  show           Show current configuration data
  switch-profile Switch between configured profiles

Flags:
  -h, --help               help for configure
  -j, --json_file string   loads the API key JSON file downloaded from the WebUI

Global Flags:
      --api_token string    access token (replaces the use of api_key and api_secret)
      --organization        access organization level data sets (org admins only)
      --subaccount string   sub-account name inside your organization (org admins only)

Use "lacework configure [command] --help" for more information about a command.
```

NOTE: AWS generate doesn't follow this rule, I am about to fix that.

afiune (commented):
I am going to start pulling this code down and testing it locally 💯 

tenor-84251231

cli/cmd/generate_gcp.go
marcosgm: lowercase is good, agreed

cli/cmd/generate_gcp.go
rmoles: I'm with @marcosgm on this. Considering we have this integration names config in the UI & docs and the module itself is also called config, I'd prefer to keep it as is.

cli/cmd/generate_gcp.go
rmoles: This is another place where I have named this to be in line with our TF module name.

cli/cmd/generate_gcp.go
rmoles: I can update the help text.
I think the command line flags help text is still going to push us over the line length though.

rmoles: Here is an updated copy of the help text with this line length roughly enforced.
I'm curious about how we can to police this going forward 🤔 
```
Use this command to generate Terraform code for deploying Lacework into an
GCP environment.

By default, this command will function interactively, prompting for the required
information to setup the new cloud account. In interactive mode, this command will:

* Prompt for the required information to setup the integration
* Generate new Terraform code using the inputs
* Optionally, run the generated Terraform code:
  * If Terraform is already installed, the version will be confirmed suitable for use
	* If Terraform is not installed, or the version installed is not suitable,
          a new version will be installed into a temporary location
	* Once Terraform is detected or installed, Terraform plan will be executed
	* The command will prompt with the outcome of the plan and allow to view more
          details or continue with Terraform apply
	* If confirmed, Terraform apply will be run, completing the setup of the
          cloud account

This command can also be run in noninteractive mode.
See help output for more details on supplying required values for generation.

Usage:
  lacework cloud-account iac-generate gcp [flags]

Flags:
      --apply                                         run terraform apply without executing plan or prompting
      --audit_log                                     enable audit log integration
      --audit_log_integration_name string             specify a custom audit log integration name
      --bucket_lifecycle_rule_age int                 specify the lifecycle rule age (default -1)
      --bucket_location string                        specify gcp bucket location
      --bucket_name string                            specify new bucket name
      --bucket_region string                          specify gcp bucket region
      --bucket_retention_days int                     specify the bucket retention days
      --config                                        enable config integration
      --config_integration_name string                specify a custom config integration name
      --enable_force_destroy_bucket                   enable force bucket destroy
      --enable_ubla                                   enable universal bucket level access(ubla)
      --existing_bucket_name string                   specify existing bucket name
      --existing_service_account_name string          specify existing service account name
      --existing_service_account_private_key string   specify existing service account key (base64 encoded)
      --existing_sink_name string                     specify existing sink name
  -h, --help                                          help for gcp
      --organization_id string                        specify the organization id (only set if organization_integration is set)
      --organization_integration                      enable organization integration
      --output string                                 location to write generated content
      --project_id string                             specify the project id to be used to provision lacework resources (required)
      --service_account_credentials string            specify a Service Account credentials JSON path (leave blank to make use of google credential ENV vars)

Global Flags:
      --api_token string    access token (replaces the use of api_key and api_secret)
      --organization        access organization level data sets (org admins only)
      --subaccount string   sub-account name inside your organization (org admins only)
```

As mentioned, the CLI flags push us over the line length limit at the moment

lwgenerate/gcp/gcp.go
```diff
@@ -22,6 +22,25 @@ func NewExistingServiceAccountDetails(name string, privateKey string) *ExistingS
...
+	if e.Name == "" && e.PrivateKey == "" {
+		return false
+	}
+
+	// If all values are populated, return false
+	if e.Name != "" && e.PrivateKey != "" {
+		return false
+	}
+
+	return true
```
dmurray-lacework, line 41: nit: you could probably condense this. Though it's pretty easy to read as it is.

dmurray-lacework (approved):

lwgenerate/gcp/gcp.go
rmoles: Going to leave this as-is for readability.

integration/test_resources/help/cloud-account_iac-generate_gcp
```diff
@@ -0,0 +1,55 @@
+Use this command to generate Terraform code for deploying Lacework into an GCP environment.
+
+By default, this command will function interactively, prompting for the required information to setup the new cloud account. In interactive mode, this command will:
```
JoshuaKim2021, line 3: How about: this command interactively prompts for the required information to...

integration/test_resources/help/cloud-account_iac-generate_gcp
```diff
@@ -0,0 +1,55 @@
...
+
+* Prompt for the required information to setup the integration
+* Generate new Terraform code using the inputs
+* Optionally, run the generated Terraform code:
+  * If Terraform is already installed, the version will be confirmed suitable for use
```
JoshuaKim2021, line 8: how about: the version is verified as compatible for use

integration/test_resources/help/cloud-account_iac-generate_gcp
```diff
@@ -0,0 +1,55 @@
...
+	* If Terraform is not installed, or the version installed is not suitable, a new version will be installed into a temporary location
```
JoshuaKim2021, line 9: should we use "compatible" instead of "suitable"?

integration/test_resources/help/cloud-account_iac-generate_gcp
```diff
@@ -0,0 +1,55 @@
...
+	* Once Terraform is detected or installed, Terraform plan will be executed
+	* The command will prompt with the outcome of the plan and allow to view more details or continue with Terraform apply
+	* If confirmed, Terraform apply will be run, completing the setup of the cloud account
+
+This command can also be run in noninteractive mode. See help output for more details on supplying required values for generation.
```
JoshuaKim2021, line 14: how about: See help output for more details on the parameter value required for Terraform code generation.

rmoles: @JoshuaKim2021
what do you think about changing `the parameter value required` in your suggestion to `the parameter value(s) required`

JoshuaKim2021: oh right. I agree with that.

cli/cmd/generate_gcp.go
afiune: I think @marcosgm said to keep it as `configuration` so:
```
Enable configuration integration?
```

afiune (changes requested):
When running this command in conjunction with the AWS generation we duplicate the required provider configuration, that is, we don't check if the folder already has this block defined:
```
terraform {
  required_providers {
    lacework = {
      source  = "lacework/lacework"
      version = "~> 0.12.2"
    }
  }
}
```
When running the generated code Terraform fails with:

Screen Shot 2022-03-09 at 1 50 40 PM

cli/cmd/generate_gcp.go
rmoles: @marcosgm just want to double-check. Are we okay to use config here? This will match what we have in AWS and Azure and in all of our public-facing docs for this integration type.

cli/cmd/generate_gcp.go
rmoles: I'm going to leave this as is to keep us consistent with our Docs and module names.

JoshuaKim2021: I agree with the Audit Log being two words.

cli/cmd/generate_gcp.go
JoshuaKim2021: The documentation has "Config" instead of "Compliance" in the integration doc articles. That terminology is in a lot of places in the external documentation actually. @marcosgm can you raise a JIRA ticket with the doc_impact field as doc-review so that the doc team can evaluate/modify those instances.

@marcosgm Are we thinking that the term "Config" is bad even when we are talking about GCP and Azure? I realize it is problematic with AWS because there is an AWS Config term.

rmoles:
> When running this command in conjunction with the AWS generation we duplicate the required provider configuration, that is, we don't check if the folder already has this block defined:
> 
> ```
> terraform {
>   required_providers {
>     lacework = {
>       source  = "lacework/lacework"
>       version = "~> 0.12.2"
>     }
>   }
> }
> ```
> 
> When running the generated code Terraform fails with:
> 
> Screen Shot 2022-03-09 at 1 50 40 PM

Good catch! IMO this is a strong argument to go with the folder approach rather than a flat lacework dir. AFAICT there is no simple way for us to check whether the lacework provider is already present.

Having spoken with @dmurray-lacework about this we came up with the following.
- Possible fixes
- Have all generation use 1 provider file
- Parse all tf before new generation to check for provider block
- Revert back to separate folder approach
- Run Tf exec providers

The `lacework provider` command, gets into a chicken and egg scenario, as we would need to call this command at the lwgenerate level and we don't check for the TF installation at this level.

Personally, I'm against the 1 provider file approach. It departs from our current flat-file approach, which is very useful for those with little TF experience to see everything in one place. This would also mean a re-work of each implementation to extract the lacework provider and write it to a separate file.

I'd prefer not to get into the scenario of parsing all files in a dir to check if we already have a provider in place...

Somewhat selfishly I am in favor of the move back to the folder approach. IMO it's the path of least resistance. We can always revisit this implementation in the future.

Let me know your thoughts on this @afiune, @ipcrm

cli/cmd/generate_gcp.go
JoshuaKim2021: @marcosgm Are we thinking that the term "Config" is bad even when we are talking about GCP and Azure? I realize it is problematic with AWS because there is an AWS Config term.  But I'm not sure there is such a concern with GCP/Azure. But then we run into inconsistent usage of "Config" vs "Compliance" for different CSPs...

afiune:
@rmoles I'm fine with that folder approach. Does that mean you are going back to use the file name `main.tf`?

cli/cmd/generate_aws.go
```diff
@@ -136,7 +138,7 @@ This command can also be run in noninteractive mode. See help output for more de
 			locationDir := filepath.Dir(location)
 			if GenerateAwsCommandExtraState.TerraformApply {
 				// Execution pre-run check
-				err := executionPreRunChecks(dirname, locationDir)
+				err := executionPreRunChecks(dirname, locationDir, "aws")
```
afiune, line 141: Are we switching back to using the `main.tf`?

cli/cmd/generate.go
afiune: I would agree with this suggestion 👆🏽

cli/cmd/generate.go
```diff
@@ -192,26 +162,22 @@ func determineOutputDirPath(location string) (string, error) {
...
 	if err != nil {
 		return false, err
 	}
 
-	if filename == "" {
-		filename = "main"
-	}
-
-	hclPath := filepath.FromSlash(fmt.Sprintf("%s/%s.tf", outputDir, filename))
+	hclPath := filepath.FromSlash(fmt.Sprintf("%s/main.tf", outputDir))
```
afiune, line 180: never mind - we are back to call it `main.tf`

afiune (approved):
This code looks good, I don't want to delayed any longer, having said that, it is late and I didn't have the chance to test it properly. I will do so early in the morning, though if I find something we can tackle it afterwards in new PRs.

Great work, Ross!

tenor-7037827

afiune:
Check windows!! 🪟

rmoles:
check the windows plz

PR #717, opened 2022-03-01
refactor(cli): further deprecate evaluatorId

hazedav:
## Summary
Deprecate Evaluator ID from the Go-SDK and Lacework CLI

Existing and updated integration testing

## Issue
https://lacework.atlassian.net/browse/ALLY-905

### DIFF
--- a/cli/cmd/lql.go
+++ b/cli/cmd/lql.go
@@ -314,7 +314,6 @@ It looks like you attempted to submit a query in JSON format.
 Please validate that the JSON is formatted properly and adheres to the following schema:
 
 {
-    "evaluatorId": "Cloudtrail",
     "queryId": "MyLQL",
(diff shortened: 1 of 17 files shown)


dmurray-lacework (approved):

afiune:
Am I assuming that we will have a separate PR to remove the duplication of Query IDs? We agreed
we should be using anonymous queries from now on.

hazedav:
> Am I assuming that we will have a separate PR to remove the duplication of Query IDs? We agreed
we should be using anonymous queries from now on.

This really has nothing to do w/ the CLI as we just pass queryText.  The main thing we'll handle in the CLI are any areas where we communicate what a query should look like...

afiune (approved):
tenor-19361462

PR #719, opened 2022-03-03
feat(cli): sort query and policy list output

hazedav:
## Summary
As a Lacework user, I would like the `lacework query ls`, `lacework query sources`, `lacework policy ls`, and `lacework policy list-tags` outputs to be sorted such that I can easily find items that I’m looking for

Manually tested
Integration tested

## Issue
https://lacework.atlassian.net/browse/ALLY-910

### DIFF
--- a/cli/cmd/lql_list.go
+++ b/cli/cmd/lql_list.go
@@ -19,6 +19,8 @@
 package cmd
 
 import (
+	"sort"
+
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 
@@ -50,6 +52,12 @@ func queryTable(queryData []api.Query) (out [][]string) {
 			query.LastUpdateUser,
 		})
 	}
+
+	// order by ID
+	sort.Slice(out, func(i, j int) bool {
+		return out[i][0] < out[j][0]
+	})
+
 	return
 }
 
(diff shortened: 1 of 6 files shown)


afiune (approved):
What about policies?

hazedav:
> What about policies?

The ticket was scoped down to just queries.  We can certainly address policies as well; but:

1.  since they mostly contain integers the sorting is slightly more complex.
2.  it wasn't clear if we want to sort policies by severity or by id

So we need some guidance as to whether we want sort by severity, id, severity-then-id, or `id-then-severity` and for any id based sorting whether we want `lwcustom-34` to come before `lwcustom-4` in the list...

David

afiune:
@hazedav wasn't involved in the scoping but, it makes sense! 🙌🏽

hazedav:
> @hazedav wasn't involved in the scoping but, it makes sense! 🙌🏽

@andrewelizondo / @afiune do you want to weigh in on how you want policies sorted or you just want me to do it by ID and handle the integers?

andrewelizondo:
Since we have a different mechanism to sort/organize policies it should probably be done separately, for datasets we really only have namespace which makes this the best approach currently. LGTM 👍

hazedav:
@afiune I went ahead and added `policy ls` sort...

andrewelizondo (approved):

hazedav:
@afiune gave me a verbal OK on this.

PR #734, opened 2022-03-16
test: fix policy severity filter test

dmurray-lacework:
The filter output is expecting to only see 'critical'  severity. It expects result not to contain 'lacework-global-1' however it is matching 'lacework-global-1**' eg 'lacework-global-189'

Fixed by adding a space  to the expected output so we do not match lacework-global-189 etc


rmoles:
Check windows!!

dmurray-lacework:
check the windows

rmoles (approved):

PR #742, opened 2022-03-22
feat(api): Implement AwsEksAudit CloudAccount interface

rmoles:

## Summary

In order to facilitate the creation of the EKS Audit Log Terraform Module, we need to add support for the AwsEksAudit cloud account integration

make fmt ✅ 
make test ✅ 

## Issue

https://lacework.atlassian.net/browse/ALLY-945

### DIFF
--- a/api/_examples/cloud-accouts/main.go
+++ b/api/_examples/cloud-accouts/main.go
@@ -33,6 +33,8 @@ func main() {
 			//case api.GcpCfgCloudAccount:
 			//case api.GcpAtSesCloudAccount:
 			support = "Supported"
+		case api.AwsEksAuditCloudAccount.String():
+			support = "Supported"
 		}
 
 		// Output: INTEGRATION-GUID:INTEGRATION-TYPE:[Supported|Unsupported]
@@ -63,7 +65,7 @@ func main() {
                               }
                             }`))
 
-	myCloudAccount := api.NewCloudAccount(
+	awsCtSqsCloudAccount := api.NewCloudAccount(
 		"cloud-from-golang",
 		api.AwsCtSqsCloudAccount,
 		awsCtSqsData,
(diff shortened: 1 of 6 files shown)


dmurray-lacework:
@rmoles could you add an example usage to examples/ folder

api/_examples/cloud-accouts/main.go
```diff
@@ -33,6 +33,8 @@ func main() {
 			//case api.GcpCfgCloudAccount:
 			//case api.GcpAtSesCloudAccount:
 			support = "Supported"
+		case api.AwsEksAuditCloudAccount.String():
```
dmurray-lacework, line 36: While you're in this file, could you fix the name

api/cloud_accounts_aws_eks_audit_test.go
```diff
@@ -0,0 +1,212 @@
...
+      },
+      "lw_account_3": {
+        "aws_accounts": [
+          "553453453",
+          "934534535"
+        ]
+      }
+    }
+  }`)
+	awsCtSqsData := api.AwsCtSqsData{
```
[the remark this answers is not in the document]
rmoles: Good catch. Struggling with what to do with this test, as it is somewhat unique to `AwsCtSqs`, as it is testing the encoding and decoding of a template file, which AFAICT is not applicable for `AwsEksAudit`

api/_examples/cloud-accouts/main.go
```diff
@@ -78,11 +80,33 @@ func main() {
 		log.Fatal(err)
 	}
 
-	response, err := orgLwClient.V2.CloudAccounts.Create(myCloudAccount)
+	awsCtSqsResponse, err := orgLwClient.V2.CloudAccounts.Create(awsCtSqsCloudAccount)
```
dmurray-lacework, line 83: Can we move these to 2 subfolders inside cloudaccounts. I think it's better if we separate the examples.

Eg.
```
 _examples/cloud_accounts/eks_audit/main.go
_examples/cloud_accounts/aws_cloudtrail/main.go
```

rmoles: Yeah makes sense. Was thinking that they should probably be separate, but there wasn't any precedent in this area, so left it as a flat file.

dmurray-lacework (changes requested):
The examples folder aws_eks_ct_sqs needs renamed to aws_ct_sqs

dmurray-lacework (approved):
Nice job

PR #743, opened 2022-03-22
feat(lql): more suitable default for lql run

hazedav:
Use "Last 24 Hours" instead of "Today"

Change is covered by existing integration

## Issue
https://lacework.atlassian.net/browse/ALLY-943


afiune (approved):
tenor-222004137

PR #748, opened 2022-03-23
refactor(cli): inefficient policy sort

hazedav:
## Summary
During policy list operation we were sorting the policy list for each entry which is very inefficient.
Also abstract sorting in case future methods need to take advantage of the sorting algorithm

Existing integration testing

## Issue
https://lacework.atlassian.net/browse/ALLY-950


afiune (commented):
The only tests I found for sorting where https://github.com/lacework/go-sdk/blob/60cefdc9d1aa8f49e8151830010f31a82a97df4c/integration/policy_test.go#L315-L322

Now that we have refactored and abstracted the function, can we add table driven unit tests like these https://github.com/lacework/go-sdk/blob/60cefdc9d1aa8f49e8151830010f31a82a97df4c/cli/cmd/compliance_gcp_test.go#L35-L122 

Best practice: https://dave.cheney.net/2019/05/07/prefer-table-driven-tests

hazedav:
@afiune , we can transition from integration to unit testing here.

hazedav:
We're going to have a chicken and egg problem with respect to `lacework-content-all` changes, so I added a pipeline locked to this branch:
https://g.codefresh.io/build/6241d77e145328fde5c5b0d2

hazedav:
@afiune / @dmurray-lacework , anything else needed here?

dmurray-lacework (approved):

PR #761, opened 2022-04-11
feat(cli): default template when creating LQL queries

afiune:
## Summary

As a Lacework CLI user,
I want to have a boilerplate template when creating new LQL queries,
So I don't have to type the entire query from scratch.

I run the command `lacework query create` without inputs and my default editor opened with:

Screen Shot 2022-04-11 at 10 26 59 PM

## Issue

N/A - Needed for my Hackathon Project.

### DIFF
--- a/cli/cmd/lql.go
+++ b/cli/cmd/lql.go
@@ -271,7 +271,7 @@ queryText: |-
           --- Add query filter(s), if any. If not, remove this block.
       }
       return {
-          --- List fields to return from the selected source. Use 'lacework query show-source <datasource>'
+          --- List fields to return from the selected source. Use 'lacework query describe <datasource>'.
       }
   }`
 		prompt.HideDefault = true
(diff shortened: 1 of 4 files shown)


afiune:
I detected a bug in the `lacework query show` command, when there are special characters like `%` the CLI was outputting:
 
Screen Shot 2022-04-11 at 10 20 53 PM

**(detected it via the new test 💯🙌🏽)**

Now it shows the correct query.

tenor-11877781

cli/cmd/lql.go
```diff
@@ -257,10 +257,28 @@ func inputQueryFromURL(url string) (query string, err error) {
...
 		Message:  fmt.Sprintf("Type a query to %s", action),
-		FileName: "query*.lql",
+		FileName: "query*.yaml",
+	}
+
+	if action == "create" {
+		prompt.Default = `queryId: YourQueryID
+queryText: |-
+  {
+      source {
```
hazedav, line 267: one observation is the use of alias `sources` and non-alias `show-source` vs. `describe`.  Not sure if it matters that much as both get you to the same outcome...

hazedav (approved):

cli/cmd/lql.go
afiune: updated! 💯

PR #765, opened 2022-04-12
feat(api): AlertProfiles v2 service

dmurray-lacework:
## Summary

New Alert Profiles service added to go-sdk .

Run `_examples/alert_profiles/main.go`
Unit Test `api/alert_profiles_test.go`

## Issue

https://lacework.atlassian.net/browse/ALLY-979

### DIFF
--- a/api/alert_profiles.go
+++ b/api/alert_profiles.go
@@ -1,0 +1,154 @@
+//
+// Author:: Darren Murray (<darren.murray@lacework.net>)
+// Copyright:: Copyright 2022, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package api
+
+import (
+	"fmt"
+
+	"github.com/pkg/errors"
+)
+
+// AlertProfilesService is the service that interacts with
+// the AlertProfiles schema from the Lacework APIv2 Server
+type AlertProfilesService struct {
+	client    *Client
+	Templates AlertTemplatesService
+}
+
+// NewAlertProfile returns an instance of the AlertProfileConfig struct
+//
+// Basic usage: Initialize a new AlertProfileConfig struct, then
+//              use the new instance to do CRUD operations
+//
+//   client, err := api.NewClient("account")
+//   if err != nil {
+//     return err
+//   }
+//
+//   alertProfile := api.NewAlertProfile(
+//		"Foo",
+//		api.AlertProfileConfig{
+// 		...
+//     },
+//   )
+//
+//   client.V2.AlertProfiles.Create(AlertProfile)
+//
+func NewAlertProfile(id string, extends string, alerts []AlertProfileAlert) AlertProfileConfig {
+	profile := AlertProfileConfig{
+		Guid:    id,
+		Extends: extends,
+		Alerts:  alerts,
+	}
+	return profile
+}
+
+// List returns a list of Alert Profiles
+func (svc *AlertProfilesService) List() (response AlertProfilesResponse, err error) {
+	err = svc.client.RequestDecoder("GET", apiV2AlertProfiles, nil, &response)
+	return
+}
+
+// Create creates a single Alert Profile
+func (svc *AlertProfilesService) Create(profile AlertProfileConfig) (
+	response AlertProfileResponse,
+	err error,
+) {
+	err = svc.client.RequestEncoderDecoder("POST", apiV2AlertProfiles, profile, &response)
+	return
+}
+
+// Delete deletes a Alert Profile that matches the provided guid
+func (svc *AlertProfilesService) Delete(guid string) error {
+	if guid == "" {
+		return errors.New("specify an intgGuid")
+	}
+
+	return svc.client.RequestDecoder(
+		"DELETE",
+		fmt.Sprintf(apiV2AlertProfileFromGUID, guid),
+		nil,
+		nil,
+	)
+}
+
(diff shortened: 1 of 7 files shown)


api/alert_profiles.go
```diff
@@ -0,0 +1,139 @@
...
+	"fmt"
+
+	"github.com/pkg/errors"
+)
+
+// AlertProfilesService is the service that interacts with
+// the AlertProfiles schema from the Lacework APIv2 Server
+type AlertProfilesService struct {
+	client    *Client
+	Templates AlertTemplatesService
```
dmurray-lacework, line 31: AlertTemplates does not have a GET so cannot implement with our V2Service interface.

The api path is nested under AlertProfiles.
/api/v2/AlertProfiles/{alertProfileId}/AlertTemplates

api/alert_profiles.go
```diff
@@ -0,0 +1,154 @@
...
+type AlertProfileField struct {
+	Name string `json:"name"`
+}
+
+type AlertProfileDescriptionKeys struct {
+	Name string `json:"name"`
+	Spec string `json:"spec"`
+}
+
+type AlertProfileAlert struct {
```
hazedav, line 137: Isn't an `AlertProfileAlert` an `AlertTemplate` ?

api/alert_templates.go
```diff
@@ -0,0 +1,57 @@
...
+func (svc *AlertTemplatesService) Create(alertProfileID string, template AlertProfileAlert) (
+	response AlertProfileResponse,
+	err error,
+) {
+	apiPath := fmt.Sprintf(apiV2AlertTemplates, alertProfileID)
+	err = svc.client.RequestEncoderDecoder("POST", apiPath, template, &response)
+	return
+}
+
+func (svc *AlertTemplatesService) Update(alertProfileID string, alertTemplateID string, template AlertProfileUpdate) (
```
hazedav, line 36: It appears the API lets you update an alert template in multiple ways?

You can either PATCH the `alerts` object via `v2/AlertProfiles/{alertProfileId}` or you can patch the actual alert template via `v2/AlertProfiles/{alertProfileId}/AlertTemplates/{alertTemplateName}`.

Given that the `AlertProfileService.Update` takes `AlertProfileUpdate`, shouldn't this be taking `AlertTemplateUpdate` ?

api/alert_profiles.go
dmurray-lacework: Good point, I'm leaning towards AlertTemplate

api/alert_templates.go
dmurray-lacework: 👍  Yeh it should be AlertTemplateUpdate. Will fix

api/alert_profiles.go
```diff
@@ -0,0 +1,154 @@
...
+// Basic usage: Initialize a new AlertProfileConfig struct, then
+//              use the new instance to do CRUD operations
+//
+//   client, err := api.NewClient("account")
+//   if err != nil {
+//     return err
+//   }
+//
+//   alertProfile := api.NewAlertProfile(
+//		"Foo",
```
afiune, line 45: There is one string missing in this example, the `extends` string.

api/alert_profiles.go
afiune: yeah, this should be AlertTemplate 👍🏽

api/alert_profiles.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+type AlertProfileAlert struct {
+	Name        string `json:"name"`
+	EventName   string `json:"eventName"`
+	Description string `json:"description"`
+	Subject     string `json:"subject"`
+}
+
+type AlertProfileUpdate struct {
```
afiune, line 144: I can't believe that we can't update the Alert Profile itself, this is only updating the alert templates 😕

api/alert_templates.go
afiune: I am so confused as well @hazedav 😖

api/alert_templates.go
```diff
@@ -0,0 +1,57 @@
...
+	apiPath := fmt.Sprintf(apiV2AlertTemplatesFromGUID, alertProfileID, alertTemplateID)
+	err = svc.client.RequestEncoderDecoder("POST", apiPath, nil, nil)
+	return
+}
+
+type AlertTemplateUpdate struct {
+	EventName   string `json:"eventName"`
+	Description string `json:"description"`
+	Subject     string `json:"subject"`
+}
```
afiune, line 57: Should this have `omitempty`? or they all MUST be provided?

api/v2.go
```diff
@@ -53,6 +54,7 @@ func NewV2Endpoints(c *Client) *V2Endpoints {
 	v2 := &V2Endpoints{c,
 		&UserProfileService{c},
 		&AlertChannelsService{c},
+		&AlertProfilesService{c, AlertTemplatesService{c}},
```
afiune, line 57: shouldn't we follow this pattern for nested services? https://github.com/lacework/go-sdk/blob/main/api/v2_vulnerabilities.go#L23-L36

afiune (commented):
Almost done, great comments about Alert Templates.

api/alert_templates.go
dmurray-lacework: 👍

api/v2.go
dmurray-lacework: Is this not slightly different in the vulnerability services are at the same level. Templates is nested under profiles.

V2.Vulnerabilities.Hosts.Func() & V2.Vulnerabilities.Containers.Func()

vs 

V2.Vulnerabilities.AlertProfiles.Func()
V2.Vulnerabilities.AlertProfiles.Templates.Func(alertProfileID)

dmurray-lacework: I'll refactor as you've suggested and see.

V2.Vulnerabilities.Alert.Profiles.Func()
V2.Vulnerabilities.Alert.Templates.Func()

Not sure if it matters but we also V2.Vulnerabilities.AlertRules and V2.Vulnerabilities.AlertChannels then not nested under "Alert"

dmurray-lacework: Seeing it, I do prefer the suggested way 👍

api/alert_profiles.go
dmurray-lacework: Yes, the api only allows updating the alert templates
https://your-account.lacework.net/api/v2/docs#tag/AlertProfiles/paths/~1api~1v2~1AlertProfiles~1{id}/patch

hazedav (approved):
LGTM

PR #767, opened 2022-04-12
refactor: enforce query fail on count

afiune:
## Summary
Refactor how `--fail_on_count` bubbles up the error code and message by following
the best practice of policy enforcement. 

Ran the command `lacework query run LW_Global_AWS_CTA_IAMAccessKeyChange --fail_on_count '!=0'` and got:
```
[ ... ]
ERROR ENFORCE POLICY: query matched fail_on_count expression. [count:22] [expr:!=0] (exit code: 9)
```

## Issue

https://lacework.atlassian.net/browse/ALLY-485

### DIFF
--- a/cli/cmd/errors_lql.go
+++ b/cli/cmd/errors_lql.go
@@ -1,0 +1,101 @@
+//
+// Author:: Salim Afiune Maya (<afiune@lacework.net>)
+// Copyright:: Copyright 2022, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/lacework/go-sdk/internal/failon"
+)
+
+type queryPolicyError struct {
+	ExitCode    int
+	Message     string
+	Err         error
+	FailonCount string
+	Count       int
+}
+
+func NewQueryPolicyError(failonCount string, count int) *queryPolicyError {
+	return &queryPolicyError{
+		FailonCount: failonCount,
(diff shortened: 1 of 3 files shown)


cli/cmd/errors_lql.go
```diff
@@ -0,0 +1,101 @@
...
+	co := failon.CountOperation{}
+	if err := co.Parse(e.FailonCount); err != nil {
+		e.ExitCode = 123
+		e.Message = err.Error()
+		return false
+	}
+
+	isFail, err := co.IsFail(e.Count)
+	if err != nil {
+		e.ExitCode = 123
```
afiune, line 85: I want to use a different error code to detect if we end up in this place, even though we should not get here because we are parsing and "failing" inside the command PreRunE func. But just in case.

cli/cmd/lql.go
```diff
@@ -415,14 +415,39 @@ func runQuery(cmd *cobra.Command, args []string) error {
...
+	if err != nil {
+		return errors.Wrap(err, "unable to run query")
+	}
+
+	// output
+	if err = cli.OutputJSON(response.Data); err != nil {
+		return err
+	}
+
+	// fail_on_count post
```
afiune, line 435: this could be removed

cli/cmd/lql.go
```diff
@@ -120,6 +120,19 @@ Start and End times are required to run a query:
     A. CLI flags take precedence over JSON specifications  
     B. JSON specifications take precedence over ParamInfo specifications  `,
 		Args: cobra.MaximumNArgs(1),
+		PreRunE: func(_ *cobra.Command, _ []string) error {
```
hazedav, line 123: I like this

cli/cmd/errors_lql.go
```diff
@@ -0,0 +1,101 @@
...
+
+package cmd
+
+import (
+	"fmt"
+
+	"github.com/lacework/go-sdk/internal/failon"
+)
+
+type queryPolicyError struct {
```
hazedav, line 27: nit.  is there a better term other than `queryPolicy`.  Given the association of queries with LPP (platform policies) this is slightly misleading.

hazedav (approved):
Approved w/ comments

cli/cmd/errors_lql.go
afiune: the important part is the `<type>Error` - Any name you want this to be called? I was thinking to remove the internal package you created `failon` and use the name of the struct defined there `CountOperationError`. Thoughts?

afiune:
@hazedav lets refactor both "policy enforcement" workflows.

PR #770, opened 2022-04-14
fix(api): add newResponse for function SearchAllPages

aircraft-cerier:
This is necessary as using the current response to iterate over pages
corrupts data that are slices, such as the tags within imageInfo. This is because
the memory address from the previous response will still contain data from the previous request.
By creating a new response using the nextPage url we get a new spot in memory to put the next
requests data.

### DIFF
--- a/api/v2_vulnerabilities.go
+++ b/api/v2_vulnerabilities.go
@@ -220,8 +220,12 @@ func (svc *v2HostVulnerabilityService) SearchAllPages(filters SearchFilter) (
 	for {
 		all = append(all, response.Data...)
 
-		pageOk, err = svc.client.NextPage(&response)
+		newResponse := VulnerabilitiesContainersResponse{
+			Paging: response.Paging,
+		}
+		pageOk, err = svc.client.NextPage(&newResponse)
 		if err == nil && pageOk {
+			response = newResponse
 			continue
 		}
 		break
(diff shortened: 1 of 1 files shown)


afiune:
Make it so!

afiune:
@aircraft-cerier Thank you so much for your first contribution to this project! 🎉 I think this update makes sense and it will help us fix the bug you are experiencing, but for us to merge this change we need two things:
1. We require all commits to be signed with a GPG key https://github.com/lacework/go-sdk/blob/main/DEVELOPER_GUIDELINES.md#signed-commits
2. We had some pipeline issues that we have fixed in the `main` branch so please, rebase your PR so that the pipeline goes green

Please let us know if you need any help with these two requests. We look forward to merging this pull request! 💯

aircraft-cerier:
@afiune I think i fixed your 2 requests, let me know if i'm missing something

rmoles:
Make it so!

rmoles (approved):
image
Thanks for the contribution @aircraft-cerier!

dmurray-lacework:
Hi @aircraft-cerier I see one of your commits is still missing the "verified" icon. Can you do a rebase, then push with a signed commit.

rmoles:
Make it so!

rmoles:
Make it so!

dmurray-lacework (approved):

PR #783, opened 2022-04-28
fix: Make azure active directory integration default to false RAIN-30986

djmctavish:
## Summary
The flag for setting the Azure Active Directory integration in the cli was defaulted to be true, meaning that the absence of the flag meant that an active directory integration was included in the generated terraform, this needed to be set to false as default and only included in the terraform if the user explicitly required it

Tested the change manually and checked the results.
Re-ran all the unit-tests
Re-ran all the Integration tests

## Issue
https://lacework.atlassian.net/browse/RAIN-30986

### DIFF
--- a/cli/cmd/generate_azure.go
+++ b/cli/cmd/generate_azure.go
@@ -354,7 +354,7 @@ func initGenerateAzureTfCommandFlags() {
 	generateAzureTfCommand.PersistentFlags().BoolVar(
 		&GenerateAzureCommandState.CreateAdIntegration,
 		"ad_create",
-		true,
+		false,
 		"create new active directory integration")
 
 	generateAzureTfCommand.PersistentFlags().BoolVar(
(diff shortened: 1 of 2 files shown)


cli/cmd/generate_azure.go
```diff
@@ -354,7 +354,7 @@ func initGenerateAzureTfCommandFlags() {
 	generateAzureTfCommand.PersistentFlags().BoolVar(
 		&GenerateAzureCommandState.CreateAdIntegration,
 		"ad_create",
-		true,
+		false,
```
rmoles, line 357: Does the config/activity log integration create the ad application if this is set to false and no alternative is supplied by the user?

djmctavish: No they get an error if this is not set and they do not give details of an existing active directory integration

rmoles: Based on our conversation via slack, I personally prefer this defaulted to `true`

djmctavish: Giving it some thought I think that you are right Ross, I will leave it as it is currently

rmoles:
closing as it was decided that this change is not required.

PR #784, opened 2022-05-03
feat: Add support for APIv2 CloudAccounts List

rmoles:

## Summary

Currently, the `lacework integrations list` command uses the v1 integration API which does not show new cloud account integrations such as the `AwsEksAudit` CA integration.

We should add `lacework cloud-accounts list` as a supported command to ensure users can display all of their cloud account integrations via the CLI. We should aim to add full CRUD for cloud-accounts, but this is a first step to allow the users to at least see their AWS EKS integrations. Ticket to cover full CRUD https://lacework.atlassian.net/browse/ALLY-999

image

## Issue

https://lacework.atlassian.net/browse/ALLY-956

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -102,6 +102,7 @@ const (
 	apiV2AlertRuleFromGUID = "v2/AlertRules/%s"
 
 	apiV2CloudAccounts        = "v2/CloudAccounts"
+	apiV2CloudAccountsByType  = "v2/CloudAccounts/%s"
 	apiV2CloudAccountFromGUID = "v2/CloudAccounts/%s"
 
 	apiV2AgentAccessTokens       = "v2/AgentAccessTokens"
--- a/api/api.go
+++ b/api/api.go
@@ -101,9 +101,8 @@ const (
 	apiV2AlertRules        = "v2/AlertRules"
 	apiV2AlertRuleFromGUID = "v2/AlertRules/%s"
 
-	apiV2CloudAccounts        = "v2/CloudAccounts"
-	apiV2CloudAccountsByType  = "v2/CloudAccounts/%s"
-	apiV2CloudAccountFromGUID = "v2/CloudAccounts/%s"
+	apiV2CloudAccounts          = "v2/CloudAccounts"
+	apiV2CloudAccountsWithParam = "v2/CloudAccounts/%s"
 
 	apiV2AgentAccessTokens       = "v2/AgentAccessTokens"
 	apiV2AgentAccessTokensSearch = "v2/AgentAccessTokens/search"
--- a/api/cloud_accounts.go
+++ b/api/cloud_accounts.go
@@ -84,15 +84,29 @@ type cloudAccountType int
 const (
 	// type that defines a non-existing Cloud Account integration
 	NoneCloudAccount cloudAccountType = iota
+	AwsCfgCloudAccount
 	AwsCtSqsCloudAccount
 	AwsEksAuditCloudAccount
+	AwsUsGovCfgCloudAccount
+	AwsUsGovCtSqsCloudAccount
+	AzureAlSeqCloudAccount
+	AzureCfgCloudAccount
+	GcpAtSesCloudAccount
+	GcpCfgCloudAccount
 )
 
 // CloudAccountTypes is the list of available Cloud Account integration types
 var CloudAccountTypes = map[cloudAccountType]string{
-	NoneCloudAccount:        "None",
-	AwsCtSqsCloudAccount:    "AwsCtSqs",
-	AwsEksAuditCloudAccount: "AwsEksAudit",
+	NoneCloudAccount:          "None",
+	AwsCfgCloudAccount:        "AwsCfg",
+	AwsCtSqsCloudAccount:      "AwsCtSqs",
+	AwsEksAuditCloudAccount:   "AwsEksAudit",
+	AwsUsGovCfgCloudAccount:   "AwsUsGovCfg",
+	AwsUsGovCtSqsCloudAccount: "AwsUsGovCtSqs",
+	AzureAlSeqCloudAccount:    "AzureAlSeq",
+	AzureCfgCloudAccount:      "AzureCfg",
+	GcpAtSesCloudAccount:      "GcpAtSes",
+	GcpCfgCloudAccount:        "GcpCfg",
 }
 
 // String returns the string representation of a Cloud Account integration type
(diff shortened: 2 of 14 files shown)


api/cloud_accounts.go
```diff
@@ -211,3 +218,22 @@ func (svc *CloudAccountsService) update(guid string, data interface{}, response
 	apiPath := fmt.Sprintf(apiV2CloudAccountFromGUID, guid)
 	return svc.client.RequestEncoderDecoder("PATCH", apiPath, data, response)
 }
+
+func (svc *CloudAccountsService) listByType(caType cloudAccountType, response interface{}) error {
```
hazedav, line 222: what's the benefit of `listByType` vs. just having this incorporated into `ListByType` ?

api/cloud_accounts.go
```diff
@@ -117,6 +117,13 @@ func (svc *CloudAccountsService) List() (response CloudAccountsResponse, err err
 	return
 }
 
+// ListByType lists the cloud accounts from the provided type that are available
```
hazedav, line 120: Do we need an api test for this in the `api_test` package?  There is a `cloud_accounts_test.go` established in the api folder...

api/cloud_accounts.go
```diff
@@ -211,3 +218,22 @@ func (svc *CloudAccountsService) update(guid string, data interface{}, response
 	apiPath := fmt.Sprintf(apiV2CloudAccountFromGUID, guid)
 	return svc.client.RequestEncoderDecoder("PATCH", apiPath, data, response)
 }
+
+func (svc *CloudAccountsService) listByType(caType cloudAccountType, response interface{}) error {
+	apiPath := fmt.Sprintf(apiV2CloudAccountsByType, caType.String())
+	return svc.client.RequestDecoder("GET", apiPath, nil, &response)
+}
+
+func (c v2CommonIntegrationData) Status() string {
```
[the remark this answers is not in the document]
rmoles: Guess it makes sense seeing as we have `func (c v2CommonIntegrationData) ID() string {` located near the struct

cli/cmd/cloud_account.go
```diff
@@ -12,9 +14,75 @@ var (
...
+		Use:   "list",
+		Short: "List all available cloud account integrations",
+		Args:  cobra.NoArgs,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			var (
+				cloudAccounts api.CloudAccountsResponse
+				err           error
+			)
+
+			if cloudAccountType != "" {
```
hazedav, line 32: At this point, how is `cloudAccountType` anything other than `""1

cli/cmd/cloud_account.go
```diff
@@ -12,9 +14,75 @@ var (
...
+			if err != nil {
+				return errors.Wrap(err, "unable to get cloud accounts")
+			}
+
+			if cli.JSONOutput() {
+				return cli.OutputJSON(cloudAccounts.Data)
+			}
+
+			if len(cloudAccounts.Data) == 0 {
+				cli.OutputHuman("There was no cloud account found.\n")
```
hazedav, line 50: This command seems to be having a `plural vs. singular` crisis.  I notice some of this was here before you added anything, but:

1. The API is `cloud_accounts`
2. The file is name `cloud_account`.
3. The command is `cloud-account` but aliased to `cloud-accounts`
4. In some message we say `unable to get cloud accounts`, but in other messaging `there was no cloud account`, yet we're listing multiple `accounts`...

I would actually be in favor of this command being `cloud-accounts` primary and aliased to `cloud-account`, and all of the messaging being plural in nature.  For instance `No cloud accounts found.`

cli/cmd/cloud_account.go
rmoles: I believe this is only called when the command is run

rmoles: And this then sets the value of the var
```
	// add type flag to cloud accounts list command
	cloudAccountListCmd.Flags().StringVarP(&cloudAccountType,
		"type", "t", "", "list all cloud accounts of a specific type",
	)
```

hazedav: I missed this flag.  TY.

cli/cmd/cloud_account.go
```diff
@@ -12,9 +14,75 @@ var (
...
+		Args:  cobra.NoArgs,
+		RunE: func(_ *cobra.Command, _ []string) error {
+			var (
+				cloudAccounts api.CloudAccountsResponse
+				err           error
+			)
+
+			if cloudAccountType != "" {
+				caType, found := api.FindCloudAccountType(cloudAccountType)
+				if !found {
```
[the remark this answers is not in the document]
hazedav: You confirmed this is arg validation and might make sense for `PreRunE`.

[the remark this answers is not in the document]
rmoles: You're right. Let me take a look at this.

api/cloud_accounts.go
hazedav: No worries.  Just didn't seem to be adding much value (i.e. not consumed by anything else in the API package).

cli/cmd/cloud_account.go
hazedav: Actually, looking at the list of commands (policy / query / integration / etc.) it seems like most things are in fact singular.  Perhaps just clean up that one `OutputHuman` to be `No cloud accounts found.`...

hazedav (approved):
Approved w/ comments.

cli/cmd/cloud_account.go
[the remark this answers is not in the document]
rmoles: Actually... Looking at this further. I'm not sure whether this brings us much value. It would mean calling `api.FindCloudAccountType` in two places.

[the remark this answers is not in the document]
hazedav: All good!

cli/cmd/cloud_account.go
```diff
@@ -12,9 +14,75 @@ var (
...
+			}
+			if err != nil {
+				return errors.Wrap(err, "unable to get cloud accounts")
+			}
+
+			if cli.JSONOutput() {
+				return cli.OutputJSON(cloudAccounts.Data)
+			}
+
+			if len(cloudAccounts.Data) == 0 {
```
dmurray-lacework, line 49: You need to move this length check above the json output. Otherwise when there are no cloud accounts and the --json flag is set the output will be `null`

api/api.go
```diff
@@ -102,6 +102,7 @@ const (
 	apiV2AlertRuleFromGUID = "v2/AlertRules/%s"
 
 	apiV2CloudAccounts        = "v2/CloudAccounts"
+	apiV2CloudAccountsByType  = "v2/CloudAccounts/%s"
 	apiV2CloudAccountFromGUID = "v2/CloudAccounts/%s"
```
dmurray-lacework, line 106: Wonder should there be a distinction between these 2 consts. They're both the same. Thoughts on combining them in to 1 const? `apiV2CloudAccountsWithParam` or something to that effect.

rmoles: I think that's reasonable. Let me update

api/cloud_accounts.go
dmurray-lacework: This is pretty similar to the existing get func, since the api path consts are both "v2/CloudAccounts/%s"

If this comment is implemented, the existing get func could be reused https://github.com/lacework/go-sdk/pull/784/files#r865763917

dmurray-lacework (approved):
Looks good @rmoles a couple of comments that can be addressed.

api/cloud_accounts.go
rmoles: re-worked the get function and removed the `listByType` function. We now use `get` within `ListByType `

PR #794, opened 2022-05-23
fix(cli): policy-sync should not touch policy state

hazedav:
## Summary
As a user sync’ing policies I would prefer that they don’t sync the state portion of the policy definition such that policies that I have enabled or disabled remain enabled/disabled as such.

Manually.  We do not have CDK integration testing yet.

## Issue
https://lacework.atlassian.net/browse/ALLY-1008


hazedav:
> can we add a test?

Certainly possible, but not straightforward since we haven't built out any integration test infrastructure to install and run CDK operations on live environments...

hazedav:
@afiune , will create a ticket to introduce automated integration testing for Lacework Content Library

PR #796, opened 2022-05-23
fix(cli): Add MachineDetail tags 

hazedav:
## Summary
The API has a structure defined for machine details tags.  The problem is the platform does not return a deterministic set of tags.  At a minimum the structure needs to be extended to output the set of tags known to GKE.

Manual
Existing automation

## Issue
https://lacework.atlassian.net/browse/ALLY-1009

### DIFF
--- a/api/entities_machine_details.go
+++ b/api/entities_machine_details.go
@@ -89,20 +89,42 @@ type MachineDetailEntity struct {
 	Os            string    `json:"os"`
 	OsVersion     string    `json:"osVersion"`
 	Tags          struct {
-		Account        string `json:"Account"`
-		AmiID          string `json:"AmiId"`
-		ExternalIP     string `json:"ExternalIp"`
-		Hostname       string `json:"Hostname"`
-		Name           string `json:"Name"`
-		InstanceID     string `json:"InstanceId"`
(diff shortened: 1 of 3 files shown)


api/entities_machine_details.go
```diff
@@ -89,20 +89,42 @@ type MachineDetailEntity struct {
...
-		InternalIP     string `json:"InternalIp"`
-		LwTokenShort   string `json:"LwTokenShort"`
-		SubnetID       string `json:"SubnetId"`
-		VMInstanceType string `json:"VmInstanceType"`
-		VMProvider     string `json:"VmProvider"`
-		VpcID          string `json:"VpcId"`
-		Zone           string `json:"Zone"`
-		Arch           string `json:"arch"`
-		Os             string `json:"os"`
+		// Shared Tags
```
hazedav, line 92: It's worth discussing whether we should take a different approach here.  This fix is a bit of a moving target since there are plethora of supported instance types for the agent.  I personally don't have the time to enumerate them all.  Even if we capture all the known MachineDetail tags to date how frequently will this change and will we be notified of such changes?

Should we switch to a `map[string]interface{}` ?

afiune (approved):
tenor-122918710

PR #804, opened 2022-06-02
refactor(cli): query run progress message

hazedav:
Unit testing

## Issue
https://lacework.atlassian.net/browse/ALLY-1015

### DIFF
--- a/api/lql_execute.go
+++ b/api/lql_execute.go
@@ -54,9 +54,12 @@ type ExecuteQueryByIDRequest struct {
 	Arguments []ExecuteQueryArgument `json:"arguments"`
 }
 
-func validateQueryArguments(args []ExecuteQueryArgument) (err error) {
-	var hasStart, hasEnd bool
-	var start, end time.Time
+func validateQueryArguments(args []ExecuteQueryArgument) error {
+	var (
+		hasStart, hasEnd bool
+		start, end       time.Time
+		err              error
+	)
 
 	for _, arg := range args {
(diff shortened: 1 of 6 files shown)


api/lql_execute.go
```diff
@@ -52,15 +59,15 @@ func validateQueryArguments(args []ExecuteQueryArgument) (err error) {
 	var start, end time.Time
 
 	for _, arg := range args {
-		if arg.Name == "StartTimeRange" {
+		if arg.Name == QueryStartTimeRange {
 			hasStart = true
 			start, err = validateQueryTimeString(arg.Value)
 		}
 		if err != nil {
 			return errors.Wrap(err, "invalid StartTimeRange argument")
```
afiune, line 67: You are opening a new issue, this function has named returns `(err error)` but you override all `return XYZ` statements which is an anti-pattern and we should avoid. 

The best thing to do here is to remove the named return and initialize it at the beginning of the function.

afiune (changes requested):
Refactors looks O.k. - But now I have a comment from a different function.

api/lql_execute.go
hazedav: Opening a new issue?  I have no problem fixing this, but what you're referring to has been in place for 10 mos.

hazedav: Fixed.

hazedav:
> Refactors looks O.k. - But now I have a comment from a different function.

No worries.  Fixed.

afiune (approved):
tenor-222350133

PR #811, opened 2022-06-07
fix: expiration time format

dmurray-lacework:
fix time format for expiry time

relates to https://github.com/lacework/terraform-provider-lacework/pull/316


afiune (approved):
tenor-30215096

PR #813, opened 2022-06-08
fix(cli): compliance gcp enable command

dmurray-lacework:
fix `lacework comp gcp enable <report_type>`

The CLI command `lacework compliance gcp enable` had been incorrectly set to send `disable` in the request body. This only affected the Gcp command. Aws and Azure were correct.

This fix set the correct request body.


afiune:
What is the fix?

PR #822, opened 2022-06-15
style(cli): use switch instead of if statements

afiune:
## Summary

I wanted to make a style change to switch the use of if statements in favor of a switch statement, doing so
I saw one leftover `fmt.Println()` which I am removing. 

Nice feature @hazedav !!

tenor-116042133

Ran a query with a matching expression and without it.

## Issue
https://lacework.atlassian.net/browse/ALLY-485

### DIFF
--- a/cli/cmd/lql.go
+++ b/cli/cmd/lql.go
@@ -470,7 +470,6 @@ func runQuery(cmd *cobra.Command, args []string) error {
 			queryCmdState.FailOnCount,
 			len(response.Data),
 		)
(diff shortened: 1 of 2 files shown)


cli/cmd/lql.go
```diff
@@ -470,7 +470,6 @@ func runQuery(cmd *cobra.Command, args []string) error {
 			queryCmdState.FailOnCount,
 			len(response.Data),
 		)
-		fmt.Println(queryFailonError.ExitCode)
```
hazedav, line 473: FML

PR #825, opened 2022-06-16
feat(cli): Make the html table scrollable

jjlace:
Signed-off-by: John John <john.john@lacework.net>

## Summary

Generated html evaluation was not horizontally scrollable.
So the vuln table was getting clipped and data wasn't readable

* Built the CLI locally, and generated an html eval result
* Attached image shows the horizontally scrolled table
image

## Issue

https://lacework.atlassian.net/browse/RAIN-31726


afiune (approved):
tenor-60996840

PR #826, opened 2022-06-17
feat(api): add policy exceptions service

dmurray-lacework:

## Summary

Add new policy exceptions service to interact with `v2/Exceptions` schema

Manually run `examples/policy_exceptions/main.go`
Add unit tests

## Issue

https://lacework.atlassian.net/browse/ALLY-1046

### DIFF
--- a/api/policy_exceptions.go
+++ b/api/policy_exceptions.go
@@ -1,0 +1,109 @@
+//
+// Author:: Darren Murray (<darren.murray@lacework.net>)
+// Copyright:: Copyright 2022, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


afiune (commented):
Code looks good, I just have one suggestion, wouldn't it make more sense
to nest this service inside the `Policy` service? This is just UX/aesthetics

Then we could do this instead `lacework.Policy.Exception.Create()`

api/policy_exceptions.go
```diff
@@ -0,0 +1,109 @@
...
+
+type PolicyExceptionsResponse struct {
+	Data []PolicyException `json:"data"`
+}
+type PolicyException struct {
+	ExceptionID    string                      `json:"exceptionId,omitempty"`
+	Description    string                      `json:"description"`
+	Constraints    []PolicyExceptionConstraint `json:"constraints"`
+	LastUpdateTime string                      `json:"lastUpdateTime,omitempty"`
+	LastUpdateUser string                      `json:"lastUpdateUser,omitempty"`
```
afiune, line 103: can these fields be `time.Time`?

afiune (approved):
One comment but out of that, looks good.

api/policy_exceptions.go
dmurray-lacework: LastUpdateTime, probably yeh. Though , from looking at our other lastupdatedtime fields we often use string. LastUpdateTime is only returned in the response, in terraform it'll be a computed value. A user will never have to construct a date for this field in a request. If we want to have this field also as a date type, I think having a method on PolicyException struct might be a good idea instead

PR #830, opened 2022-06-23
fix(api): allow user defined filter for ListImage and ListAllImages

aircraft-cerier:
This allows the user to provide their own filter, specifically a different timeframe other than 7 days for `ListImage` and `ListAllImages`


afiune:
Make it so

afiune:
@aircraft-cerier Thank you in advance for your time!

I don't think this is a bug-fix but more a feature request to have a function to allow the user to provide their own filter, how about leaving these functions as they are (so that they are backward compatible) and instead, adding a new function like:
```go
func (svc *EntitiesService) ListImagesWithFilters(filters SearchFilter) (response ImagesEntityResponse, err error) {
	err = svc.Search(&response, filters)
	return
}
```
This will have the same outcome as the change you are proposing.

Additionally, your change has some CI failures that you might need to look at:
```
=== Errors
api/entities_images_test.go:121:19: not enough arguments in call to c.V2.Entities.ListAllImages
        have ()
        want (api.SearchFilter)
api/entities_images_test.go:156:19: not enough arguments in call to c.V2.Entities.ListAllImages
        have ()
        want (api.SearchFilter)
```

PR #835, opened 2022-06-27
feature(api): allow user defined filter for List and ListAll Images and MachineDetails

aircraft-cerier:
This allows the user to provide their own filter, specifically a different timeframe other than 7 days for `ListImage`, `ListAllImages`, `ListMachineDetails`, and `ListAllMachineDetails`


afiune:
Make it so

afiune (approved):

PR #854, opened 2022-07-08
fix(cli): install.ps1 avoid Windows agent collision

afiune:
## Summary

This change is fixing the problem where, if a Windows workstation has
our agent installed and the user tries to install the Lacework CLI, the
installation script fails with:
```
Cannot remove item C:/Program Data/Lacework/Logs/LWData_0.log because it
is being used.
```

The solution is updating the installation script to avoid removing the
entire directory and instead, forcing the installation of the Lacework
CLI.

I used our Vagrant Windows 10 box and tested the script locally, also users can point
directly to this script with the following commands:
```
Set-ExecutionPolicy Bypass -Scope Process -Force;
iex ((New-Object System.Net.WebClient).DownloadString('https://raw.githubusercontent.com/lacework/go-sdk/0917a32a5380804883b4bdf6728addcb1e44341c/cli/install.ps1'))
```
### With This Fix
Screen Shot 2022-07-08 at 8 39 43 AM

### Without The Fix
Screen Shot 2022-07-08 at 8 40 15 AM

## Issue

https://lacework.atlassian.net/browse/ALLY-1082

### DIFF
--- a/cli/install.ps1
+++ b/cli/install.ps1
@@ -94,10 +94,9 @@ Function Assert-Shasum($archive) {
 
 Function Install-Lacework-CLI {
     $laceworkPath = Join-Path $env:ProgramData Lacework
-    if (Test-Path $laceworkPath) { Remove-Item $laceworkPath -Recurse -Force }
-    New-Item $laceworkPath -ItemType Directory | Out-Null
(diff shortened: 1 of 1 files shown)


dmurray-lacework (approved):

dmurray-lacework:
@afiune  we need to update the choco install script too

PR #858, opened 2022-07-14
docs: fix path to examples in readme

hal-y:
## Summary
This will fix the api readme file to point to the right examples directory.

verified the new path is the correct one

## Issue
haven't created one for this


afiune (approved):
tenor-16200671

PR #863, opened 2022-07-15
fix: vulnerabilty_exception fixable parameter omit when unset

dmurray-lacework:
Signed-off-by: Darren Murray [darren.murray@lacework.net](mailto:darren.murray@lacework.net)

Issue:
https://lacework.atlassian.net/browse/ALLY-1085
Fixes https://github.com/lacework/terraform-provider-lacework/issues/339

Description:
Allow fixable to be nil value

### DIFF
--- a/api/vulnerability_exceptions.go
+++ b/api/vulnerability_exceptions.go
@@ -446,7 +446,7 @@ type VulnerabilityExceptionPackage struct {
 	Version string
 }
 
-func (vc VulnerabilityExceptionCriteria) FixableEnabled()  *bool {
+func (vc VulnerabilityExceptionCriteria) FixableEnabled() *bool {
 	if vc.Fixable == nil {
 		return nil
 	}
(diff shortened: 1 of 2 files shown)


api/vulnerability_exceptions.go
```diff
@@ -446,11 +446,17 @@ type VulnerabilityExceptionPackage struct {
...
+	}
+
 	if len(vc.Fixable) > 0 {
-		return vc.Fixable[0] == 1
+		truePtr := vc.Fixable[0] == 1
+		return &truePtr
 	}
-	return false
+	falsePtr := false
+	return &falsePtr
```
afiune, line 459: Can we write a few unit tests to verify the three things/options we want this function to return.

1) When it is false, return false
1) When it is true, return true
1) When it is not present, return nil
1) When it is empty, return nil (but I don't know if this is possible or, what if the APIs return empty when we set it and then remove it)

afiune: Maybe to remove the `fixable == true` or `fixable == false` we need to use a `PUT`

afiune (approved):

PR #866, opened 2022-07-18
refactor: use v2/AgentInfo for agents list command

dmurray-lacework:

## Summary

Use `v2/AgentInfo` api instead of `Entities/MachineDetails` in agent list command. Add status field to output

Run `lacework agent list`

## Issue

https://lacework.atlassian.net/browse/ALLY-1071


api/agent_access_tokens.go
```diff
@@ -169,3 +169,71 @@ type AgentAccessTokenRequest struct {
...
+func (r *AgentInfoResponse) ResetPaging() {
+	r.Paging = V2Pagination{}
+}
+
+type AgentInfo struct {
+	AgentVersion string    `json:"agentVersion"`
+	CreatedTime  time.Time `json:"createdTime"`
+	Hostname     string    `json:"hostname"`
+	IpAddr       string    `json:"ipAddr"`
+	LastUpdate   string    `json:"lastUpdate"`
```
afiune, line 195: We need to change this to:
```suggestion
	LastUpdate   time.Time    `json:"lastUpdate"`
```
But we need to wait for this change to be released: https://github.com/lacework/services/pull/8095

Jira: https://lacework.atlassian.net/browse/RAIN-34237

cli/cmd/agent_list.go
```diff
@@ -175,15 +175,15 @@ func listAgents(_ *cobra.Command, _ []string) error {
 
 	cli.OutputHuman(
 		renderSimpleTable(
-			[]string{"MID", "Short Agent Token", "Hostname", "Name", "IP Address", "External IP", "OS Arch", "Last Checkin"},
-			machineDetailsToListAgentTable(machines),
+			[]string{"MID", "Short Agent Token", "Hostname", "Name", "Status", "IP Address", "External IP", "OS Arch", "Last Checkin"},
```
afiune, line 178: Add column created time and use the `lastUpdate` field for the Last Checkin column.

Order table like:
```suggestion
			[]string{"MID", "Short Agent Token", "Status", "Hostname", "Name", "IP Address", "External IP", "OS Arch", "Last Check-in", "Created Time"},
```

Choose if Created at or Created time

cli/cmd/agent_list.go
```diff
@@ -207,17 +210,31 @@ func machineExist(machines []api.AgentInfo, mid int) bool {
 func agentInfoToListAgentTable(machines []api.AgentInfo) [][]string {
 	out := [][]string{}
 	for _, m := range machines {
+
+		// @afiune testing how to truncate a long value
+		hostname := m.Hostname
+		if len(hostname) > 20 {
+			hostname = fmt.Sprintf("%s...", hostname[:20])
```
dmurray-lacework, line 217: Are we truncating for formatting purposes? After we've obscured information how does the user see the full value, the json response?

dmurray-lacework: If we are going to include fixed field limits we should have a single function that is shared for consistency. If it's possible to include it in the tablewriter that would be nice.

api/agent_access_tokens.go
```diff
@@ -169,3 +169,71 @@ type AgentAccessTokenRequest struct {
 	TokenAlias string                 `json:"tokenAlias,omitempty"`
 	Props      *AgentAccessTokenProps `json:"props,omitempty"`
 }
+
+func (svc *AgentAccessTokensService) Search(response interface{}, filters SearchFilter) error {
```
afiune, line 173: Why is this new function inside the `AgentAccessTokensService`??? This is really wrong!

Every API v2 schema should be its own service 😓 - I will have to refactor this too.

afiune (changes requested):

afiune (approved):
tenor-211043291

PR #868, opened 2022-07-20
feat(lql): add limit option 

hazedav:
## Summary
As a Lacework user, I want to be able to limit LQL query results

`lacework query run LW_Global_AWS_CTA_S3BucketPolicyChange --limit 10`

Manually
Automated Unit + Integration

## Issue
https://lacework.atlassian.net/browse/ALLY-1102


api/lql_execute.go
```diff
@@ -38,18 +38,34 @@ const (
 	QueryEndTimeRange   ExecuteQueryArgumentName = "EndTimeRange"
 )
 
+type ExecuteQueryOptions struct {
+	Limit *int `json:"limit,omitempty"`
+}
```
afiune, line 43: Are there any more query options?

cli/cmd/lql.go
```diff
@@ -153,34 +159,41 @@ func init() {
 	// run specific flags
 	setQuerySourceFlags(queryRunCmd)
 
-	// since time flag
-	queryRunCmd.Flags().StringVarP(
+	// limit flag
+	queryRunCmd.Flags().IntVar(
+		&queryCmdState.Limit,
+		"limit", 0,
```
afiune, line 165: Is this the only command we should add this new flag?

api/lql_execute.go
hazedav: This is the only one.

cli/cmd/lql.go
hazedav: Yes, my understanding of this flag (similar to the parameters) is that `limit` is only applicable to query execution

hazedav: I double checked the API docs and confirmed this.

api/lql_execute.go
```diff
@@ -38,18 +38,34 @@ const (
...
+func (eqo ExecuteQueryOptions) Validate() error {
+	if eqo.Limit == nil {
+		return nil
+	}
+	if *eqo.Limit >= 1 {
+		return nil
+	}
+	return errors.New("limit must be at least 1")
+}
+
```
afiune, line 54: Do we need to implement all of that? Shouldn't this be enforced by the APIs?

afiune: Testing passing a negative value does show:
```
[400] Incorrect configuration of field options.limit: must be greater than or equal to 1
```

afiune: Up to you to leave it, this is just a nitpick since you already implemented the logic/code.

afiune (approved):
## That was fast!
tenor-134928607

api/lql_execute.go
hazedav: I worked this up based on some early go-sdk feedback you had given about reducing API calls if we're able to validate client side.  The pro is the ability to reduce server-side calls, but the con is having to keep in-line with platform validation.

If you're good with it I might chose to remove this in favor of server-side validation...

David

Issue #874, opened 2022-07-29
feat(cli): CLI version tags for Docker image 


heywoodlh:
# Feature Request

The current [lacework/lacework-cli Docker image](https://hub.docker.com/r/lacework/lacework-cli) has tags related to the underlying operating system in the container -- which, in my opinion, does not provide much value to those consuming the CLI through containers.

It would be nice if one could pull/run the container tag based on the version of the Lacework CLI -- not the underlying container's operating system.

i.e. `docker run -it --rm lacework/lacework-cli:v0.39.0`

Note: My PR for the multi-arch Dockerfile [provides an example of using a Github Action to extract the version string from the container and then applies that as a tag](https://github.com/heywoodlh/actions/blob/4979f27591a0d578bd713519dff93d17a30ce8e8/.github/workflows/lacework-cli-buildx.yml#L37-L50).

dmurray-lacework:
Hey @heywoodlh thank you so much for raising this. The team has taken a quick look and we decided we need to rework our image tagging and release process. We'll create a ticket internally to make these changes. We'll keep you posted as soon as we start progress.

dmurray-lacework:
This work is now complete https://hub.docker.com/r/lacework/lacework-cli/tags

PR #875, opened 2022-07-29
feat(cli): multiarch Dockerfile

heywoodlh:
## Summary

The current [lacework/lacework-cli Docker image](https://hub.docker.com/r/lacework/lacework-cli) image is not cross-architecture (i.e. you cannot run the image natively on a non-x86_64 OS such as on an M1 Apple device, Raspberry Pi, etc.).

This Dockerfile provides a multi-stage build in which the Lacework CLI is compiled and then moved to the final stage. This allows the Docker image to build and subsequently run on any architecture that can build the Lacework CLI with Golang.

Here is an example Github Action for building the container using Docker's buildx plugin (for multi-architecture images): https://github.com/heywoodlh/actions/blob/master/.github/workflows/lacework-cli-buildx.yml

Which then pushes my multi-arch image to DockerHub: https://hub.docker.com/r/heywoodlh/lacework-cli

And you can see that the tagged image is multi-arch:

image

The image runs the same as before:

```
$ docker run -it --rm heywoodlh/lacework-cli version
lacework v0.39.1-dev (sha:unknown) (time:unknown)

A newer version of the Lacework CLI is available! The latest version is v0.39.0,
to update execute the following command:

curl https://raw.githubusercontent.com/lacework/go-sdk/main/cli/install.sh | bash
```

To test each architecture, if you have a local Docker system running and it is setup to run images on different architectures you can use the `--platform` flag:

ARM64:

```
$ docker run -it --rm --platform='linux/arm64' heywoodlh/lacework-cli:v0.39.1-dev  version
lacework v0.39.1-dev (sha:unknown) (time:unknown)
...
```

ARMv7 (32 bit):

```
$ docker run -it --rm --platform='linux/arm/v7' heywoodlh/lacework-cli:v0.39.1-dev version
lacework v0.39.1-dev (sha:unknown) (time:unknown)
...
```

### Bonus:

My Github Action also provides an example of pulling out the version string of the Lacework CLI and using it as a tag. Which would solve issue https://github.com/lacework/go-sdk/issues/874: https://github.com/heywoodlh/actions/blob/4979f27591a0d578bd713519dff93d17a30ce8e8/.github/workflows/lacework-cli-buildx.yml#L37-L50


rmoles:
make it so

afiune:
@heywoodlh thank you so much for your contribution, unfortunately we couldn't use your code since we did a refactor on the way we distribute our docker image. I still would appreciate your feedback about he changes we have made:

* https://github.com/lacework/go-sdk/pull/1074
* https://github.com/lacework/go-sdk/pull/1097

I look forward to collaborating with you in the near feature. Happy new year! 🎉

heywoodlh:
@afiune The refactor is much better -- thanks for the time and effort put in to implement cross-arch images and better tagging! Very welcome change.

PR #883, opened 2022-08-05
fix: alert profile filter

dmurray-lacework:
Fix alert profile filter used in `lacework alert-profile update`

This will return the correct list of profiles that can be updated. Non default profiles

### DIFF
--- a/cli/cmd/alert_profiles.go
+++ b/cli/cmd/alert_profiles.go
@@ -288,7 +288,7 @@ func promptSelectProfile() (string, error) {
 	if err != nil {
 		return "", err
 	}
-	var profileList = filterAlertProfiles(profileResponse)
+	var profileList = filterAlertProfilesByDefault(profileResponse)
 
 	questions := []*survey.Question{
 		{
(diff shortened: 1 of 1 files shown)


cli/cmd/alert_profiles.go
```diff
@@ -425,6 +425,21 @@ func promptAddAlertTemplate() api.AlertTemplate {
...
+	}
+
+	sort.Slice(profiles, func(i, j int) bool {
+		return profiles[i] < profiles[j]
+	})
+
+	return profiles
+}
+
 func filterAlertProfiles(response api.AlertProfilesResponse) []string {
```
afiune, line 443: Do we still need this function?

dmurray-lacework: yes

cli/cmd/alert_profiles.go
```diff
@@ -425,6 +425,21 @@ func promptAddAlertTemplate() api.AlertTemplate {
 	}
 }
 
+func filterAlertProfilesByDefault(response api.AlertProfilesResponse) []string {
```
rmoles, line 428: `ByNonDefault` makes a bit more sense to me as we are filtering out the Default Profiles
```suggestion
func filterAlertProfilesByNonDefault(response api.AlertProfilesResponse) []string {
```

rmoles (approved):
I would prefer a different name for the function.
However this is a nitpick and doesn't change the functionality.
Happy to merge if you'd prefer to leave as is.

PR #887, opened 2022-08-10
feat: DataExportRule service

dmurray-lacework:

## Summary

Add new v2 service DataExportRule

Run new examples `api/_examples/data-export-rules/main.go`

## Issue
https://lacework.atlassian.net/browse/ALLY-1150

### DIFF
--- a/api/data_export_rules.go
+++ b/api/data_export_rules.go
@@ -1,0 +1,113 @@
+//
+// Author:: Darren Murray (<darren.murray@lacework.net>)
+// Copyright:: Copyright 2022, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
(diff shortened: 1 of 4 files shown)


api/data_export_rules.go
```diff
@@ -0,0 +1,113 @@
...
+	Message string           `json:"message"`
+}
+
+type DataExportRuleResponse struct {
+	Data    DataExportRule `json:"data"`
+	Message string         `json:"message"`
+}
+
+type DataExportRule struct {
+	ID     string               `json:"mcGuid,omitempty"`
```
rmoles, line 44: What's the scenario that the ID would be empty?

rmoles: Or are we just allowing it at this level to handle it more gracefully and with a nice error message?

dmurray-lacework: A Post or Update request does not contain the id in the body

dmurray-lacework: For update it actually causes the request to be invalid, so we strip the id along with created/updated fields inside the update function

rmoles (approved):
Looks good to me 👍

PR #892, opened 2022-08-22
Ally 1161 cli subdomain capture

jon-stewart:
## Summary

The Ecosystem onboarding bundle invokes the Lacework CLI using `configure` in non-interactive mode with the global flags - not using `json_file`.  If the customer provided the onboarding service with the full URL (i.e. account.cluster.lacework.net) the onboarding script would pass that full URL to the CLI leading to an error.

This PR seeks to address this issue by treating the non-interactive `configure` account flag the same as if the information was provided by `json_file`.  As in we make use of the `lwdomain` package to validate and select the relevant domain information.

This PR also ensure that the api_key and api_secret validation see in interactive mode is made use of.

I have not deduplicated the validation code from the interactive prompt path.

- Unit test
- Manual test
- Invocation via onboard bundle [script](https://github.com/lacework/services/blob/main/cloud-account-registrar/onboarding/base_onboard_bundle.sh) 

## Issue

[ALLY-1161](https://lacework.atlassian.net/browse/ALLY-1161)

### DIFF
--- a/cli/cmd/configure.go
+++ b/cli/cmd/configure.go
@@ -223,10 +223,10 @@ func runConfigureSetup() error {
 			newProfile.Subaccount = subaccount
 		}
 		cli.OutputHuman("\n")
-	}
-
-	if err := newProfile.Verify(); err != nil {
-		return errors.Wrap(err, "unable to configure the command-line")
+	} else {
+		if err := newProfile.Verify(); err != nil {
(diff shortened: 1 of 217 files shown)


rmoles:
Probably worth noting that most customers will have a URL of `account.lacework.net` the `account.cluster.lacework.net` is somewhat unique to us internally. example customer foo - `foo.lacework.net`

PR #899, opened 2022-08-26
build(cli): sign Lacework CLI artifacts

afiune:
## Summary

Every day we have more Windows Lacework CLI users, most of them have expressed a security concern
since we were not signing our Windows artifacts which caused Real-time defender to alert and avoid installing
the Lacework CLI.

This PR is updating our CI pipeline that releases the Lacework CLI to sign the artifacts before releasing.

tenor-458396

We will test this new orchestration during our next release.

tenor-66486042

## Issue

https://lacework.atlassian.net/browse/ALLY-970


ipcrm (approved):
🤞 LGTM!

PR #906, opened 2022-09-01
refactor(chocolatey): address moderation requested changes

dmurray-lacework:
## Summary

Address changes requested by chocolatey moderation process

Completed
```
- The iconUrl should be switched from github raw to a CDN: https://docs.chocolatey.org/en-us/create/create-packages#package-icon-guidelines
- Any changes to environment variables should be undone in a chocolateyUninstall.ps1
- Checksum(s) should be hardcoded into the package instead of being downloaded at runtime.
```

## Issue

https://lacework.atlassian.net/browse/ALLY-1152

### DIFF
--- a/scripts/chocolatey/build/tools/chocolateyinstall.ps1
+++ b/scripts/chocolatey/build/tools/chocolateyinstall.ps1
@@ -11,118 +11,32 @@ This script installs the Lacework Command Line Interface.
 
 $ErrorActionPreference = "stop"
 
-Set-Variable GithubReleasesRootUrl -Option ReadOnly -value "https://github.com/lacework/go-sdk/releases" -Force
-Set-Variable PackageName -Option ReadOnly -value "lacework-cli-windows-amd64.exe.zip" -Force
+$version = '0.36.0'
+$packageName = 'lacework-cli'
+$toolsDir = "$( Split-Path -parent $MyInvocation.MyCommand.Definition )"
+$url = "https://github.com/lacework/go-sdk/releases/download/v0.36.0/lacework-cli-windows-386.exe.zip"
+$url64 = "https://github.com/lacework/go-sdk/releases/download/v0.36.0/lacework-cli-windows-amd64.exe.zip"
 
-Function Get-File($url, $dst)
-{
-    Write-Host "Downloading $url"
-    try
-    {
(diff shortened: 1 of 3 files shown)


scripts/chocolatey/build/tools/chocolateyinstall.ps1
```diff
@@ -11,118 +11,32 @@ This script installs the Lacework Command Line Interface.
...
-{
-    if ($PSVersionTable.PSEdition -eq 'Core')
-    {
-        [System.Security.Cryptography.SHA256]::Create()
-    }
-    else
-    {
-        New-Object -TypeName Security.Cryptography.SHA256Managed
-    }
+$version = '0.36.0'
```
afiune, line 14: should this version string have the `v`? that is, `v0.36.0` 🤔

scripts/chocolatey/build/tools/chocolateyinstall.ps1
```diff
@@ -11,118 +11,32 @@ This script installs the Lacework Command Line Interface.
...
+$packageName = 'lacework-cli'
+$toolsDir = "$( Split-Path -parent $MyInvocation.MyCommand.Definition )"
+$url = "https://github.com/lacework/go-sdk/releases/download/v0.36.0/lacework-cli-windows-386.exe.zip"
+$url64 = "https://github.com/lacework/go-sdk/releases/download/v0.36.0/lacework-cli-windows-amd64.exe.zip"
```
afiune, line 18: can we use the version above to build this URLs?

scripts/chocolatey/build/tools/chocolateyinstall.ps1
dmurray-lacework: that doesn't seem to be the convention with chocolatey.

scripts/chocolatey/build/tools/chocolateyinstall.ps1
dmurray-lacework: Yes, once we pass moderation we'll inject he var and do string replacement in CI. But for this version it's just hard coded

scripts/chocolatey/build/tools/chocolateyinstall.ps1
afiune: I don't know if that will confuse our users, but if there is no way around it, we will let it go

afiune (approved):

PR #934, opened 2022-10-06
feat(cli): find and fix AWS instances w/o agents

nschmeller:
feat(cli): find and fix AWS instances w/o agents

This commit is the first milestone for Project Capture The Flag. Project
CTF takes a user's cloud credentials, lists all of their hosts,
and installs the agent on hosts without agents. The purpose of
Project CTF is to allow customers to easily close gaps in their agent
deployments by using the CLI to find and install the agent on instances
that the customer would have trouble reaching otherwise.

This commit adds a new subcommand to the Lacework CLI called `aws-install`.
The user should invoke this command via `lacework agent aws-install`.
`aws-install` is currently a no-op and is used as a directory for further
subcommands.

This commit adds two subcommands to `aws-install` called `ec2ic` and `ec2-ssh`.
The user should invoke them as either `lacework agent ctf aws-install ec2ic` or
`lacework agent ctf aws-install ec2-ssh`.

The `ec2ic` subcommand uses EC2InstanceConnect to send SSH keys to the
user's EC2 instances. EC2InstanceConnect is only supported on official
channel AMIs that are Ubuntu 16.04 or later or Amazon Linux 2, so the
customer will not be able to use `ec2ic` for instances running other
AMIs. `ec2ic` adds flags to filter instances on tag, tag key, and region.

The `ec2-ssh` subcommand uses existing SSH authentication to install agents
on the user's EC2 instances. `ec2-ssh` assumes that all instances that it
finds will have SSH access via the authentication method passed via CLI
flags. `ec2-ssh` also adds flags to filter instances on tag, tag key, and
region.

`ec2ic` and `ec2-ssh` both have the limitation that they only work for
instances with public IP addresses that are open to the internet on port
22.

Fixes RAIN-37375

## Summary

This PR adds a new feature to 1) find all the AWS EC2 instances in a user's account without agents and 2) install agents on them. 

### Looking for feedback on...

I'd love feedback on:
* Name of the new command
* Nested structure (`aws-install` is a directory command and doesn't do anything)
* Anything else! Nits are welcome!

We now use this new feature in the CLI to install agents for the agent repo integration tests. Also added a unit test to verify the logic for finding instances and checking if there are agents on them. (This new unit test requires AWS credentials). 

## Issue

https://lacework.atlassian.net/browse/RAIN-37375


cli/cmd/agent.go
```diff
@@ -213,6 +295,53 @@ func init() {
...
+	agentCTFAWSEC2ICCmd.Flags().StringArrayVarP(&agentCmdState.CTFIncludeRegions,
+		"include_regions", "r", []string{}, "list of regions to filter on",
+	)
+
+	// 'agent ctf aws ssh' flags
+	agentCTFAWSSSHCmd.Flags().StringVar(&agentCmdState.CTFInfraTagKey,
+		"infra_tag_key", "", "only install agents on infra with this tag key set",
+	)
+	agentCTFAWSSSHCmd.Flags().StringArrayVar(&agentCmdState.CTFInfraTag,
+		"infra_tag", []string{}, "only install agents on infra with this tag",
```
afiune, line 321: nitpick: can we use a single string with the format `key:value`? (Just like we use it at https://github.com/lacework/go-sdk/blob/main/cli/cmd/agent_list.go#L72-L74)

cli/cmd/agent.go
```diff
@@ -213,6 +295,53 @@ func init() {
...
+	agentCTFAWSEC2ICCmd.Flags().BoolVar(&agentCmdState.InstallTrustHostKey,
+		"trust_host_key", false, "automatically add host keys to the ~/.ssh/known_hosts file",
+	)
+	agentCTFAWSEC2ICCmd.Flags().StringArrayVarP(&agentCmdState.CTFIncludeRegions,
+		"include_regions", "r", []string{}, "list of regions to filter on",
+	)
+
+	// 'agent ctf aws ssh' flags
+	agentCTFAWSSSHCmd.Flags().StringVar(&agentCmdState.CTFInfraTagKey,
+		"infra_tag_key", "", "only install agents on infra with this tag key set",
```
afiune, line 318: why do we prefix these flags with `infra_`, can we remove it and name it `tag_key`? 
```suggestion
		"tag_key", "", "only install agents on infrastructure with this tag key",
```
finally, why is this a "tag key set"? `set` sounds like two tag keys 🤔

cli/cmd/agent.go
```diff
@@ -213,6 +295,53 @@ func init() {
...
+		"infra_tag", []string{}, "only install agents on infra with this tag",
+	)
+	agentCTFAWSEC2ICCmd.Flags().StringVar(&agentCmdState.InstallAgentToken,
+		"token", "", "agent access token",
+	)
+	agentCTFAWSEC2ICCmd.Flags().BoolVar(&agentCmdState.InstallTrustHostKey,
+		"trust_host_key", false, "automatically add host keys to the ~/.ssh/known_hosts file",
+	)
+	agentCTFAWSEC2ICCmd.Flags().StringArrayVarP(&agentCmdState.CTFIncludeRegions,
+		"include_regions", "r", []string{}, "list of regions to filter on",
```
afiune, line 313: Will there be an `exclude_regions` flag in the future? or why prefix this with `include_`?

If the answer is yes, leave it, if no, I rather drop the prefix.
```suggestion
		"regions", "r", []string{}, "list of regions to filter on",
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+	agentCTFAWSEC2ICCmd = &cobra.Command{
+		Use:   "ec2ic",
+		Short: "Use EC2InstanceConnect to install the datacollector agent on all remote AWS hosts",
+		RunE:  awsCaptureTheFlagEC2IC,
+		Long: `This command installs the agent on all remote AWS EC2 hosts in your account
+using EC2InstanceConnect.
+
+To only install on instances in a specified list of regions:
+
+	lacework agent ctf aws ec2ic --include_regions us-west-2 us-east-2
```
afiune, line 184: When mentioning an example command use spaces, not tabs.

```suggestion
    lacework agent ctf aws ec2ic --include_regions us-west-2 us-east-2
```

4 spaces to be explicit so that this translates into Markdown like this https://docs.lacework.com/cli/commands/lacework_agent_install/

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
 NOTE: New agents could take up to an hour to report back to the platform.`,
 		RunE: installRemoteAgent,
 	}
+
+	agentCTFCmd = &cobra.Command{
+		Use:   "ctf",
```
afiune, line 166: We tend to be very explicit with the purpose of the commands, ctf (capture the flag) means nothing to the end-user, but if we try things like `deploy`, `bulk-install`, `cloud-install`. it might help the user understand what this "namespace" is about. 

Another option here is to drop the entire `ctf` command/namespace/directory and put the commands as:
* `lacework agent aws-deploy ec2ic`
* `lacework agent aws-deploy ssh`
Or:
* `lacework agent aws-install ec2ic`
* `lacework agent aws-install ssh`

In short, this definitely needs to change, I will think more about which name to use here but it might be nice to ask UX for help.

cli/cmd/agent_installctf.go
```diff
@@ -0,0 +1,254 @@
...
+		return err
+	}
+
+	for _, runner := range runners {
+		cli.Log.Debugw("runner: ", "runner", runner)
+		cli.Log.Debugw("runner user: ", "user", runner.Runner.User)
+		cli.Log.Debugw("runner region: ", "region", runner.Region)
+		cli.Log.Debugw("runner az: ", "az", runner.AvailabilityZone)
+		cli.Log.Debugw("runner instance ID: ", "instance ID", runner.InstanceID)
+		cli.Log.Debugw("runner: ", "runner hostname", runner.Runner.Hostname)
```
afiune, line 95: Can we make debug messages like this one a single line? Like this:
```suggestion
		cli.Log.Debugw("runner: ", "user", runner.Runner.User, "region", runner.Region, "az", runner.AvailabilityZone, "instance_id", runner.InstanceID, "hostname", runner.Runner.Hostname)
```

cli/cmd/agent_installctf.go
```diff
@@ -0,0 +1,254 @@
...
+
+		err := runner.Runner.UseIdentityFile(agentCmdState.InstallIdentityFile)
+		if err != nil {
+			return errors.Wrap(err, "unable to use provided identity file")
+		}
+
+		if err := verifyAccessToRemoteHost(&runner.Runner); err != nil {
+			cli.Log.Debugw("verifyAccessToRemoteHost failed")
```
afiune, line 53: nitpick: No need to log these kind of log statements after a `return err` since ultimately we will bubble up the error message to the user.

cli/cmd/agent_installctf.go
```diff
@@ -0,0 +1,254 @@
...
+			return err
+		}
+
+		if alreadyInstalled := isAgentInstalledOnRemoteHost(&runner.Runner); alreadyInstalled != nil {
+			cli.Log.Debugw("agent already installed on host, skipping")
+			continue
```
afiune, line 59: Now, these log messages make perfect sense! 👍🏽

cli/cmd/agent_installctf.go
```diff
@@ -0,0 +1,254 @@
...
+	input := &ec2.DescribeRegionsInput{
+		Filters: filters,
+	}
+	cfg, err := config.LoadDefaultConfig(context.TODO())
+	if err != nil {
+		return nil, err
+	}
+	svc := ec2.New(ec2.Options{
+		Credentials: cfg.Credentials,
+		Region:      "us-west-2", // use us-west-2 for lack of a better region
```
afiune, line 174: Can we check first the `AWS_REGION` environment variable?

afiune (commented):
Really great stuff, Nick! 

tenor-175385889

A few initial comments, I will test this locally tomorrow, but lets talk about the sub-command structure over Slack. 👍🏽

cli/cmd/agent.go
cirego: Do we have anyone on the UX team that has helped with command naming in the past?

I had proposed `lacework agent install ec2-ssh` and `lacework agent install ec2-instance-connect` in the `ctf` document. I also like the `aws-deploy` command name myself.

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+	}
+
+	agentCTFAWSCmd = &cobra.Command{
+		Use:   "aws",
+		Short: "Install the datacollector agent on all remote AWS hosts",
+	}
+
+	agentCTFAWSEC2ICCmd = &cobra.Command{
+		Use:   "ec2ic",
+		Short: "Use EC2InstanceConnect to install the datacollector agent on all remote AWS hosts",
```
cirego, line 177:
```suggestion
		Short: "Use EC2InstanceConnect to securely connect to EC2 instances",
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
 	}
+
+	agentCTFCmd = &cobra.Command{
+		Use:   "ctf",
+		Short: "Install the datacollector agent on all remote hosts",
+	}
+
+	agentCTFAWSCmd = &cobra.Command{
+		Use:   "aws",
+		Short: "Install the datacollector agent on all remote AWS hosts",
```
cirego, line 172:
```suggestion
		Short: "Install the datacollector agent on all EC2 instances in an AWS account",
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+	}
+
+	agentCTFAWSEC2ICCmd = &cobra.Command{
+		Use:   "ec2ic",
+		Short: "Use EC2InstanceConnect to install the datacollector agent on all remote AWS hosts",
+		RunE:  awsCaptureTheFlagEC2IC,
+		Long: `This command installs the agent on all remote AWS EC2 hosts in your account
+using EC2InstanceConnect.
```
cirego, line 180:
```suggestion
		Long: `This command installs the agent on all EC2 instances in an AWS account
using EC2InstanceConnect.
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+
+To only install on instances in a specified list of regions:
```
cirego, line 182:
```suggestion
To filter by one or more regions:
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+
+	lacework agent ctf aws ec2ic --include_regions us-west-2 us-east-2
+
+To filter by AWS tag:
```
cirego, line 186:
```suggestion
To filter by instance tag:
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+
+    lacework agent ctf aws ec2ic --infra_tag TagName TagValue
+
+To filter by AWS tag key:
```
cirego, line 190:
```suggestion
To filter by instance tag key:
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+
+	lacework agent ctf aws ec2ic --infra_tag_key TagName
+
+The environment should contain AWS credentials in the following variables:
```
cirego, line 194:
```suggestion
AWS credentials are read from the following environment variables:
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+- AWS_ACCESS_KEY_ID
+- AWS_SECRET_ACCESS_KEY
+- AWS_SESSION_TOKEN (optional)`,
+	}
+
+	agentCTFAWSSSHCmd = &cobra.Command{
+		Use:   "ssh",
+		Short: "Provide and use SSH authentication to install the datacollector agent on all remote AWS hosts",
```
cirego, line 202:
```suggestion
		Short: "Use SSH to securely connect to EC2 instances",
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+		Long: `This command installs the agent on all remote AWS EC2 hosts in your account
+using existing SSH authentication.
```
cirego, line 204:
```suggestion
		Long: `This command installs the agent on all EC2 instances in an AWS account
using SSH.
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+
+To only install on instances in a specified list of regions:
```
cirego, line 206:
```suggestion
To filter by one or more regions:
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+
+	lacework agent ctf aws ssh --include_regions us-west-2 us-east-2
+
+To filter by AWS tag:
```
cirego, line 210:
```suggestion
To filter by instance tag:
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+
+    lacework agent ctf aws ec2ic --infra_tag TagName TagValue
+
+To filter by AWS tag key:
```
cirego, line 214:
```suggestion
To filter by instance tag key:
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+
+	lacework agent ctf aws ec2ic --infra_tag_key TagName
+
+You will need to provide an SSH authentication method. This authentication method
+should work for all instances that your tag or region filters select.
```
cirego, line 219: This method requires a routable IP address, correct?

```suggestion
You will need to provide an SSH authentication method. This authentication method
should work for all instances that your tag or region filters select. Instances must
be routable from your local host.
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+
+To authenticate to your selected hosts with a username and password:
```
cirego, line 221:
```suggestion
To authenticate using username and password:
```

cli/cmd/agent.go
```diff
@@ -158,6 +161,77 @@ To list all active agents in your environment.
...
+
+    lacework agent ctf aws ssh --ssh_username <your-user> --ssh_password <secret>
+
+To authenticate to your selected hosts with an identity file instead:
```
cirego, line 225:
```suggestion
To authenticate using an identity file:
```

cli/cmd/agent.go
cirego, line 321: I was going to suggest `filter instances using this tag`, but perhaps instead we could say something like this?

```suggestion
		"infra_tag", []string{}, "only select instances with this tag",
```

cli/cmd/agent.go
```diff
@@ -213,6 +295,53 @@ func init() {
...
+	agentCTFAWSSSHCmd.Flags().StringVarP(&agentCmdState.InstallIdentityFile,
+		"identity_file", "i", defaultSshIdentityKey,
+		"identity (private key) for public key authentication",
+	)
+	agentCTFAWSSSHCmd.Flags().StringVar(&agentCmdState.InstallAgentToken,
+		"token", "", "agent access token",
+	)
+	agentCTFAWSSSHCmd.Flags().BoolVar(&agentCmdState.InstallTrustHostKey,
+		"trust_host_key", false, "automatically add host keys to the ~/.ssh/known_hosts file",
+	)
```
cirego, line 332: Should we add this flag as an option or should we just pass `-o StrictHostKeychecking=no` to the SSH command?

cli/cmd/agent_installctf.go
```diff
@@ -0,0 +1,254 @@
...
+	})
+
+	output, err := svc.DescribeRegions(context.TODO(), input)
+	if err != nil {
+		return nil, err
+	}
+	return output.Regions, nil
+}
+
+func awsFindRunnersInRegion(region string) ([]*lwrunner.AWSRunner, error) {
```
cirego, line 184: Rather than inventing our own terminology, I think we might prefer to use the AWS terminology:

```suggestion
func awsRegionDescribeInstances(region string) ([]*lwrunner.AWSRunner, error) {
```

cli/cmd/agent_installctf.go
```diff
@@ -0,0 +1,254 @@
...
+		if err != nil {
+			cli.Log.Debugw("runInstallCommandOnRemoteHost failed")
+			return err
+		}
+	}
+
+	return nil
+}
+
+func awsFindRunnersToCapture() ([]*lwrunner.AWSRunner, error) {
```
cirego, line 133: Perhaps use AWS terminology?

```suggestion
func awsDescribeInstances() ([]*lwrunner.AWSRunner, error) {
```

cirego (commented):
I'm sorry that this is going to look like a lot of nit-picks. Overall, this PR is awesome! The feedback really only falls into two categories:

1. As Afiune mentioned, it would be good to run the command names and help text by the UI / docs teams for help on wording. I've supplied some of my own suggestions.

2. Boring terminology is easier for other engineers to understand. I think we should default to using the cloud provider terminology when possible, instead of providing our own names.

lwrunner/awsrunner.go
```diff
@@ -0,0 +1,141 @@
...
+	}
+	result, err := svc.DescribeImages(context.Background(), &input)
+	if err != nil {
+		return nil, err
+	}
+	if len(result.Images) != 1 {
+		return nil, fmt.Errorf("expected to find only one AMI")
+	}
+
+	// Heuristically assign SSH username based on AMI name
```
nschmeller, line 65: Note: allow user to supply username as flag for EC2IC. Give user the power to override the heuristic

cli/cmd/agent.go
nschmeller: I went with `lacework agent install ec2-ssh` and `lacework agent install ec2ic` for now, happy to change this! I decided on `ec2ic` over `ec2-instance-connect` for the sake of brevity, but I can see why we'd want to do `ec2-instance-connect` to have style consistent with `ec2-ssh`

cli/cmd/agent.go
nschmeller: Correct, good to clarify

cli/cmd/agent.go
nschmeller: There will be an `exclude_regions` flag, tracked at https://lacework.atlassian.net/browse/RAIN-37868. Not sure if I'll get it in this PR though

cli/cmd/agent.go
nschmeller: I'm happy to do this -- but I think `StringSliceVar` might not support `:` as a delimiter (https://pkg.go.dev/github.com/spf13/pflag@v1.0.5#FlagSet.StringSliceVar). I ran a test where I tried `:` and `,`, only `,` worked. So this might be a bug with `agent list`

cli/cmd/agent.go
nschmeller: Good point, let's keep it simple

nschmeller:
What should we do with the Guardrails error? It says there's no patch available, and I moved to the latest AWS SDK version available

cli/cmd/aws.go
```diff
@@ -0,0 +1,159 @@
...
+		allRunners = append(allRunners, regionRunners...)
+	}
+
+	return allRunners, nil
+}
+
+// awsFindRegions queries the AWS API to list all the regions that
+// are enabled for the user's AWS account. Use the "include_regions"
+// command-line flag to only get regions in this list.
+func awsFindRegions() ([]types.Region, error) {
```
cirego, line 53: Nit?

```suggestion
func awsDescribeRegions() ([]types.Region, error) {
```

cli/cmd/agent_install_ec2ic_test.go
```diff
@@ -0,0 +1,71 @@
...
+
+	cli.LogLevel = "DEBUG"
+	agentCmdState.InstallTrustHostKey = true
+	agentCmdState.InstallTagKey = "CaptureTheFlagPlayer"
+	cli.NonInteractive()
+
+	runners, err := awsDescribeInstances()
+	assert.NoError(t, err)
+
+	wg := new(sync.WaitGroup)
```
cirego, line 50: Why parallelize EC2 instance connect installation and not SSH?

cirego (approved):
Looks awesome!

cli/cmd/agent_install_ec2ic_test.go
nschmeller: This is just to make the test run faster -- if the tester's account has a ton of instances, then it would be tedious to iterate over them serially. And we don't have a unit test for SSH. I guess we could add one, but this is basically an integration test, and we already have those in the agent repo.

nschmeller: Maybe it would be worthwhile to do this in the actual CLI. I'm not sure how this will work at scale...

cirego: Oh, sorry, this is just the test. Please file this one away for later then.

Probably a good idea to consider adding some sort of parallelization (think 10000 instances)

nschmeller:
We need to get https://lacework.atlassian.net/browse/RAIN-38534 into this PR. Integration tests are taking 15min for CTF, and ten of those minutes is the CLI execution to go over instances.

nschmeller:
> We need to get https://lacework.atlassian.net/browse/RAIN-38534 into this PR. Integration tests are taking 15min for CTF, and ten of those minutes is the CLI execution to go over instances.

Let's do this in another PR -- it's not blocking our integration tests anymore, and we should keep changes separate

afiune:
Make it so! 🙌🏽

cli/cmd/agent_install_ec2ic.go
```diff
@@ -0,0 +1,126 @@
...
+		"tag_key", "", "only install agents on infra with this tag key set",
+	)
+	agentInstallAWSEC2ICCmd.Flags().StringSliceVar(&agentCmdState.InstallTag,
+		"tag", []string{}, "only install agents on infra with this tag",
+	)
+	agentInstallAWSEC2ICCmd.Flags().StringVar(&agentCmdState.InstallAgentToken,
+		"token", "", "agent access token",
+	)
+	agentInstallAWSEC2ICCmd.Flags().BoolVar(&agentCmdState.InstallTrustHostKey,
+		"no_trust_host_key", true, "do not automatically add host keys to the ~/.ssh/known_hosts file",
```
afiune, line 73: Why is this flag set to `true`? shouldn't it be `false`? At least the default behavior in the help message says that we trust the hosts, so I think this should be false 🤔

cli/cmd/agent_install_ec2ic.go
```diff
@@ -0,0 +1,126 @@
...
+	)
+	agentInstallAWSEC2ICCmd.Flags().StringSliceVarP(&agentCmdState.InstallIncludeRegions,
+		"include_regions", "r", []string{}, "list of regions to filter on",
+	)
+	agentInstallAWSEC2ICCmd.Flags().StringVar(&agentCmdState.InstallSshUser,
+		"ssh_username", "", "username to login with",
+	)
+}
+
+func installAWSEC2IC(_ *cobra.Command, args []string) error {
```
afiune, line 83: nit: you don't use `args`
```suggestion
func installAWSEC2IC(_ *cobra.Command, _ []string) error {
```

cli/cmd/agent_install_ec2ic.go
```diff
@@ -0,0 +1,126 @@
...
+			var err error
+			token, err = selectAgentAccessToken()
+			if err != nil {
+				return err
+			}
+		}
+		cmd := fmt.Sprintf("sudo sh -c \"curl -sSL %s | sh -s -- %s\"", agentInstallDownloadURL, token)
+		err = runInstallCommandOnRemoteHost(&runner.Runner, cmd)
+		if err != nil {
+			cli.Log.Debugw("runInstallCommandOnRemoteHost failed")
```
afiune, line 120: nit: we don't need this debug log since we are returning the error
```suggestion
```

afiune:
make it so

cli/cmd/agent_install_ec2ssh.go
```diff
@@ -0,0 +1,145 @@
...
+			var err error
+			token, err = selectAgentAccessToken()
+			if err != nil {
+				return err
+			}
+		}
+		cmd := fmt.Sprintf("sudo sh -c \"curl -sSL %s | sh -s -- %s\"", agentInstallDownloadURL, token)
+		err = runInstallCommandOnRemoteHost(&runner.Runner, cmd)
+		if err != nil {
+			cli.Log.Debugw("runInstallCommandOnRemoteHost failed")
```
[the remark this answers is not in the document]
nschmeller: Good idea, thanks!

cli/cmd/agent_install_ec2ic.go
nschmeller: Yup, I agree, fixed in the other comment

afiune:
make it so!

nschmeller:
> make it so!

Lux fiat! 😄

nschmeller:
Lux fiat!

nschmeller:
make it so!

afiune:
Make it so

integration/test_resources/help/agent_aws-install_ec2-ssh
```diff
@@ -0,0 +1,64 @@
+This command installs the agent on all EC2 instances in an AWS account
+using SSH.
+
+To filter by one or more regions:
+
+    lacework agent install ec2-ssh --include_regions us-west-2,us-east-2
```
selvats, line 6: Please change this to "lacework agent aws-install ec2-ssh

integration/test_resources/help/agent_aws-install_ec2ic
```diff
@@ -0,0 +1,52 @@
+This command installs the agent on all EC2 instances in an AWS account using EC2InstanceConnect.
+
+To filter by one or more regions:
+
+    lacework agent install ec2ic --include_regions us-west-2,us-east-2
```
selvats, line 5: Please change all instances of "lacework agent install ec2ic" to "lacework agent aws-install ec2ic" in this file.

PR #938, opened 2022-10-11
fix: azure Lacework provider & profile flag

jon-stewart:
## Summary

Azure TF requires the Lacework provider and the Lacework API profile defined, if a default profile hasn't been configured.

```
> terraform plan
╷
│ Error: Invalid provider configuration
│
│ Provider "registry.terraform.io/lacework/lacework" requires explicit configuration. Add a provider block to the root module and configure the provider's required arguments as described in
│ the provider documentation.
│
╵
╷
│ Error: Unable to create Lacework API client
│
│   with provider["registry.terraform.io/lacework/lacework"],
│   on <empty> line 0:
│   (source code not available)
│
│ profile 'default' not found.
│
│ Try using the Lacework CLI command 'lacework configure --profile default'
```

This error doesn't occur if `.lacework.toml` contains a default block.  It is a problem for the onboarding service as the bundle (bash script) doesn't configure a default profile, it configures a profile called 'onboarding'.

- unit
- integration
- manual

## Issue

[RAIN-38163](https://lacework.atlassian.net/browse/RAIN-38163)

### DIFF
--- a/lwgenerate/azure/azure.go
+++ b/lwgenerate/azure/azure.go
@@ -54,6 +54,8 @@ type GenerateAzureTfConfigurationArgs struct {
 
 	// Azure region where the storage account for logging resides
 	StorageLocation string
+
+	LaceworkProfile string
 }
 
 // Ensure all combinations of inputs are valid for supported spec
@@ -189,6 +191,12 @@ func WithStorageLocation(location string) AzureTerraformModifier {
 	}
 }
 
+func WithLaceworkProfile(name string) AzureTerraformModifier {
+	return func(c *GenerateAzureTfConfigurationArgs) {
+		c.LaceworkProfile = name
+	}
+}
+
 // Generate new Terraform code based on the supplied args.
 func (args *GenerateAzureTfConfigurationArgs) Generate() (string, error) {
 	// Validate inputs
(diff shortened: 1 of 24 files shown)


lwgenerate/azure/azure.go
```diff
@@ -242,6 +256,11 @@ func (args *GenerateAzureTfConfigurationArgs) Generate() (string, error) {
 
 func createRequiredProviders() (*hclwrite.Block, error) {
 	return lwgenerate.CreateRequiredProviders(
+		lwgenerate.NewRequiredProvider(
+			"lacework",
+			lwgenerate.HclRequiredProviderWithSource(lwgenerate.LaceworkProviderSource),
+			lwgenerate.HclRequiredProviderWithVersion(lwgenerate.LaceworkProviderVersion),
```
rmoles, line 262: I wonder if we can bump this Lacework provider version. Latest is 0.26 https://github.com/lacework/terraform-provider-lacework/tags

rmoles (approved):
Two comments. Otherwise, this looks good.

lwgenerate/azure/azure.go
rmoles: Do we also need to implement this for AWS/GCP?

jon-stewart: AWS and GCP already had the lacework provider section in place, this issue only affecting Azure when running `terraform plan/apply`

PR #939, opened 2022-10-11
fix: azure interactive caching

jon-stewart:
## Summary

CLI command, generate Azure TF, wasn't making use of the values loaded from cache.

- unit testing
- integration testing
- manual testing

## Issue

[ALLY-1217](https://lacework.atlassian.net/browse/ALLY-1217)

### DIFF
--- a/cli/cmd/generate_azure.go
+++ b/cli/cmd/generate_azure.go
@@ -317,6 +317,7 @@ By default, this command will function interactively, prompting for the required
 )
 
 type AzureGenerateCommandExtraState struct {
+	AskAdvanced    bool
 	Output         string
 	TerraformApply bool
 }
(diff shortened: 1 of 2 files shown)


rmoles (approved):
Looks good to me

cli/cmd/generate_azure.go
```diff
@@ -317,6 +317,7 @@ By default, this command will function interactively, prompting for the required
 )
 
 type AzureGenerateCommandExtraState struct {
+	AskAdvanced    bool
```
ipcrm, line 320: Looks like this is done inconsistently (before and now).  GCP/Azure (with your change) both cache this, whereas AWS does not. I think it makes sense to cache this. If the customer answered advanced questions previously and is resuming the cache, probably makes sense to re-prompt them with everything they may have selected before.

Not sure if we should raise a different PR for that to not muddy the water here; but figured it was worth calling out.

ipcrm (approved):
Looks good! One question

cli/cmd/generate_azure.go
jon-stewart: Good spot, I agree as we cache each of the advanced options.  I'll include in this PR.

PR #946, opened 2022-10-13
fix(lwcomponent): better JSON handling components

hazedav:
## Summary
When attempting to `json.Unmarshal` the `lwcomponent.Component` object several quirks are present.
```
{
    "name": "abc",
    "description": "def",
    "type": "BINARY",
    "version": {},
    "artifacts": [
        {
            "os": "darwin",
            "arch": "amd64",
            "url": "",
            "signature": ""
        }
    ],
    "breadcrumbs": {},
    "command_name": ""
}
```

1.  Some items should be omitted if empty (i.e. `url` and `command_name`)
2. `version` should be a string not a dict

Unit tests added

## Issue
https://lacework.atlassian.net/browse/ALLY-1229

### DIFF
--- a/lwcomponent/component.go
+++ b/lwcomponent/component.go
@@ -239,7 +239,7 @@ var (
 type Artifact struct {
 	OS        string `json:"os"`
 	ARCH      string `json:"arch"`
-	URL       string `json:"url"`
+	URL       string `json:"url,omitempty"`
 	Signature string `json:"signature"`
 	//Size ?
 }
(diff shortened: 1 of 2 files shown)


lwcomponent/component.go
```diff
@@ -255,12 +255,45 @@ type Component struct {
 	Name          string         `json:"name"`
 	Description   string         `json:"description"`
 	Type          Type           `json:"type"`
-	LatestVersion semver.Version `json:"version"`
+	LatestVersion semver.Version `json:"-"`
```
afiune, line 258: We need this field.

afiune (changes requested):

lwcomponent/component.go
afiune, line 258: Maybe:
```suggestion
	LatestVersion semver.Version `json:"version,omitempty"`
```

hazedav: Please see `MarshalJSON` and `UnmarshalJSON` which properly handle this field.

afiune: You're way to smart!

afiune (approved):
Way to smart!!

hazedav:
Thanks for the speedy review @afiune

PR #948, opened 2022-10-15
Parallelize installation with goroutines

nschmeller:
Fixes RAIN-38534
Fixes RAIN-38695

## Summary

`aws-install` only does one install at a time. For a dozen instances, this is fine. For 10k instances, this doesn't work. This commit uses goroutines to parallelize agent installation.

Added a new "unit" test that runs the parallelized install. That passes with AWS credentials. Also tested manually. Output appears much faster.

## Issue

https://lacework.atlassian.net/browse/RAIN-38695


cirego (changes requested):
This is awesome!

I think we should add a `--max_parallelism` factor to ensure that we don't hit API rate-limiting.

nschmeller:
While testing this change, I ran across this error:
```
                Error:          Received unexpected error:
                                operation error EC2 Instance Connect: SendSSHPublicKey, exceeded maximum number of attempts, 3, https response error StatusCode: 400, RequestID: 57980602-ad8e-448a-a358-0ff6533e289c, ThrottlingException: Server is busy! Please retry after a moment.
                Test:           TestAwsEC2ICDescribeInstances
```
So now we know this is what we're trying to prevent!

nschmeller:
I'm looking at libraries to manage goroutine pools (concurrency managers). So far I've found:

* https://github.com/zenthangplus/goccm
  * This was a bit too primitive for me, required using `sync.WaitGroup` to work correctly
* https://github.com/panjf2000/ants
  * This was a bit too built-out for me, on `init` of the program (so any time the CLI was invoked) it would create a thread pool. 
* https://github.com/korovkin/limiter
  * I thought this had a pretty nice balance of features / simplicity. Only downside is that you can only pass in one type of function, of type `func()`. So need to do an extra trick with `sync.WaitGroup` to copy memory from the caller scope to the callee scope

Would be curious what others think about these libraries!

nschmeller:
I tried doing some refactoring to use one function for the install function body for each method (`ec2ic`,  `ssh`). This got a little messy, so I abandoned it, but I checkpointed the work at https://github.com/lacework/go-sdk/tree/nick/parallelize-aws-ctf-tmp.

cirego (approved):

cirego:
Looks great to me @nschmeller. Nice job finding a minimal library for the "threadpool" part.

lwrunner/awsrunner.go
```diff
@@ -44,6 +44,7 @@ func NewAWSRunner(amiImageId, host, region, availabilityZone, instanceID string,
 	if err != nil {
 		return nil, err
 	}
+	cfg.Region = region
```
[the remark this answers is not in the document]
nschmeller: Fair point, addressing this in https://github.com/lacework/go-sdk/pull/952

cli/cmd/agent_install_ec2ssh.go
```diff
@@ -99,6 +100,7 @@ func init() {
 	agentInstallAWSSSHCmd.Flags().IntVar(&agentCmdState.InstallSshPort,
 		"ssh_port", 22, "port to connect to on the remote host",
 	)
+	agentInstallAWSSSHCmd.Flags().IntVarP(&agentCmdState.InstallMaxParallelism, "max_parallelism", "n", 50, "maximum number of workers executing AWS API calls, set if rate limits are lower or higher than normal")
```
afiune, line 103: nit line too long

cli/cmd/agent_install_ec2ssh.go
```diff
@@ -107,44 +109,60 @@ func installAWSSSH(_ *cobra.Command, _ []string) error {
...
+				"user", threadRunner.Runner.User,
+				"region", threadRunner.Region,
+				"az", threadRunner.AvailabilityZone,
+				"instance ID", threadRunner.InstanceID,
+				"hostname", threadRunner.Runner.Hostname,
+			)
+
+			err := threadRunner.Runner.UseIdentityFile(agentCmdState.InstallIdentityFile)
+			if err != nil {
+				cli.Log.Warnw("unable to use provided identity file", "err", err, "threadRunner", threadRunner.InstanceID)
```
afiune, line 135: the standard for fields in the CLI log entries is `snake_case` rather than `CamelCase`
```suggestion
				cli.Log.Warnw("unable to use provided identity file", "err", err, "thread_runner", threadRunner.InstanceID)
```

cli/cmd/agent_install_ec2ssh.go
```diff
@@ -107,44 +109,60 @@ func installAWSSSH(_ *cobra.Command, _ []string) error {
...
+		runnerCopyWg.Add(1)
+
+		_, err := cl.Execute(func() {
+			threadRunner := *runner
+			runnerCopyWg.Done()
+			cli.Log.Debugw("threadRunner info: ",
+				"user", threadRunner.Runner.User,
+				"region", threadRunner.Region,
+				"az", threadRunner.AvailabilityZone,
+				"instance ID", threadRunner.InstanceID,
```
afiune, line 129: same here, this key should be
```suggestion
				"instance_id", threadRunner.InstanceID,
```

cli/cmd/agent_install_ec2ic.go
```diff
@@ -20,8 +20,9 @@ package cmd
 
 import (
 	"fmt"
+	"sync"
 
-	"github.com/pkg/errors"
+	"github.com/korovkin/limiter"
```
afiune, line 25: why adding this MIT Go package rather than implementing the parallelism ourselves?

cli/cmd/agent_install_ec2ic.go
```diff
@@ -69,7 +70,7 @@ func init() {
 		"tag", []string{}, "only install agents on infra with this tag",
 	)
 	agentInstallAWSEC2ICCmd.Flags().StringVar(&agentCmdState.InstallAgentToken,
-		"token", "", "agent access token",
+		"token", "", "agent access token (mandatory)",
```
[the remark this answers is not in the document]
nschmeller: Good point, will do

cli/cmd/agent_install_ec2ic.go
afiune: yeah,lets move this back to Slack 😸

cli/cmd/agent_install_ec2ic.go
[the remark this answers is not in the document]
nschmeller: Resolved in https://github.com/lacework/go-sdk/pull/952

cli/cmd/agent_install_ec2ssh.go
nschmeller: Fixed with dec996131b7e4eda7f87ebe0a3e8ad44ab4a899b

cli/cmd/agent_install_ec2ssh.go
nschmeller: Fixed with eb423daaf99b4291448e156ea4ffe8d3e0e0a3af

cli/cmd/agent_install_ec2ssh.go
nschmeller: Fixed with ca8c5fbbbdc6cd61d2120d2307188b113ca724e9

cli/cmd/agent_install_ec2ic.go
nschmeller: Resolve this with 4ae6efdfe27251ff26b4041fa07cdea55229d048

nschmeller:
Manually ran integration tests, and these changes are also being tested on the agent repo CI. `aws-vault exec default -- go test -run TestAwsInstallEC2IC` passes. This is ready to merge!

PR #952, opened 2022-10-17
Change token from flag into arg for `aws-install`, region fix

nschmeller:
This commit:
* Removes the `--token` flags and makes it a mandatory argument
* Adds a `Region` field to the `DescribeImages`
  config to prevent crashes.

This reverts commit 6b983fe69537c7f8601c20bd038c0d90c25ca925.

CTF integration tests pass with the region fix, they were failing before. Help text integration tests are passing.

## Issue

https://lacework.atlassian.net/browse/RAIN-38750


cirego (approved):

afiune (approved):
tenor-239772961

PR #961, opened 2022-10-18
feat(cli): adding alert list and open commands

hazedav:
## Summary
Adding `alert list` and `alert open` commands

`alert list` uses time range flags (`--range`, `--start`, and `--end` consistent with lql)

Integration testing

## Issue
https://lacework.atlassian.net/browse/ALLY-1183


hazedav:
When reviewing this please see changes from 6 commits (as `lacework alert show` was encapsulated in a separate PR before being merged in here).  This can be achieved with a "shift select" in the "Files changed" menu, or this link should also work:
https://github.com/lacework/go-sdk/pull/961/files/b0aefd357fc34c7008bed52a8f051e9f4c77ed8f

api/alerts.go
```diff
@@ -63,18 +70,58 @@ type Alert struct {
 	Reachability  string             `json:"reachability"`
 }
 
+type Alerts []Alert
+
+func (a Alerts) SortDescending() Alerts {
```
afiune, line 75: Can we sort by severity instead?

cli/cmd/alert_list.go
```diff
@@ -0,0 +1,262 @@
...
+		Aliases: []string{"ls"},
+		Short:   "List all alerts",
+		Long: `List all alerts.
+
+By default, alerts are shown for the last 24 hours.
+Use a custom time range by suppling a range flag...
+
+    lacework alert ls --range "last 7 days"
+
+...Or by specifing start and end flags.
```
afiune, line 48:
```suggestion
Or by specifying start and end flags.
```

cli/cmd/alert_list.go
```diff
@@ -0,0 +1,262 @@
...
+	cli.OutputHuman(
+		renderCustomTable(
+			[]string{"Alert ID", "Type", "Name", "Severity", "Start Time", "End Time", "Status"},
+			alertListTable(alerts),
+			tableFunc(func(t *tablewriter.Table) {
+				t.SetAutoWrapText(false)
+				t.SetBorder(false)
+			}),
+		),
+	)
```
afiune, line 177: Can we copy the table options from "events2?
```go
// easily add or remove borders to all event details tables
func eventsTableOptions() tableOption {
	return tableFunc(func(t *tablewriter.Table) {
		t.SetBorders(tablewriter.Border{
			Left:   false,
			Right:  false,
			Top:    true,
			Bottom: false,
		})
		t.SetColumnSeparator(" ")
	})
}
```

What do you think?

cli/cmd/alert_comment.go
```diff
@@ -26,10 +26,10 @@ import (
 )
 
 var (
-	// alertCommentCmd represents the lql validate command
+	// alertCommentCmd represents the alert comment command
 	alertCommentCmd = &cobra.Command{
 		Use:   "comment <alert_id>",
-		Short: "Add a comment.",
+		Short: "Add a comment",
 		Long: `Post a user comment on an alert's timeline .
```
afiune, line 33: should we validate that the alert ID exists before trying to add a comment? that is, before typing the comment via the EDITOR.

afiune (approved):
tenor-207694606

api/alerts.go
hazedav: Sure thing.

cli/cmd/alert_list.go
hazedav: Good eye.

cli/cmd/alert_comment.go
hazedav: This requires another round trip to the API.  We've gone back and forth with what sort of pre-validation we want to do on the client vs. server.  I will defer to your judgement here.

cli/cmd/alert_list.go
hazedav: Do you want all the alert tables to use these options?

afiune: It really depends, I wish we could have someone from UX to look at all these tables. When I did
`lacework event` I tried to use the terminal space as much as possible. I will leave it to your criteria
but use as much space as possible when it makes sense and when it doesn't make the output look
confusing, that is, when there are big long text.

cli/cmd/alert_comment.go
afiune: I think that is what we did with `lacework query update <id>` 🤔 - I would suggest us doing that
extra API call to make sure the customer doesn't start writing the comment and then, find out
that the id was wrong.

hazedav: I think this could also apply to `lacework alert close` so I will file a Jira for us to follow up on this as an enhancement.

hazedav: https://lacework.atlassian.net/browse/ALLY-1261

api/alerts.go
hazedav: Fixed.

cli/cmd/alert_list.go
hazedav: Fixed.

cli/cmd/alert_list.go
hazedav: I'm going to create something for follow up on this.

PR #964, opened 2022-10-18
test(cli): disable compliance tests

afiune:
Essentially, our CI has been failing since yesterday, we need to continue doing development
in this repo so we decided to disable the affected tests.

tenor-60735828


cirego (approved):
Thanks @afiune! I look forward to having these tests enabled again soon!

PR #976, opened 2022-10-21
fix: compliance evaluation search cmd, searches greater than last 7 days

dmurray-lacework:
## Summary

The compliance evaluation search cmd was only searching the last 7 days. The max amount of days allowed for this search is 92, with a window of 7 days. 

This change adds a new windowedSearch func to perform a new search of a specific time frame size until response data is found or the max searchable days is reached

Run `lacework aws comp search <resource_arn>` with  the resource arn from JIRA acceptance criteria

## Issue

https://lacework.atlassian.net/browse/ALLY-1196


afiune (changes requested):
Lets do the interface instead https://github.com/lacework/go-sdk/pull/977

PR #977, opened 2022-10-21
Dmurray lacework/ally 1196 alt

dmurray-lacework:
alternate approach to https://github.com/lacework/go-sdk/pull/976


api/v2_search_filters.go
```diff
@@ -54,3 +54,42 @@ type TimeFilter struct {
...
+	GetTimeFilter() *TimeFilter
+	SetStartTime(*time.Time)
+	SetEndTime(*time.Time)
+}
+
+type search func(response interface{}, filters SearchableFilter) error
+
+// WindowedSearch performs a new search of a specific time frame size,
+// until response data is found or the max searchable days is reached
+func WindowedSearch(fn search, size int, max int, response SearchResponse, filter SearchableFilter) error {
```
afiune, line 72: Please, add a few (a number of them) unit tests.

cli/cmd/compliance_aws.go
```diff
@@ -448,7 +448,7 @@ The output from status with the --json flag can be used in the body of PATCH api
 					Csp:     api.AwsInventoryType,
 				}
 			)
-			err := cli.LwApi.V2.Inventory.Search(&awsInventorySearchResponse, filter)
+			err := api.WindowedSearch(cli.LwApi.V2.Inventory.Search, 7, 92, &awsInventorySearchResponse, &filter)
```
afiune, line 451: Can we set constants with the real min, max values rather than arbitrary numbers?

api/v2_search_filters.go
```diff
@@ -54,3 +54,42 @@ type TimeFilter struct {
...
+	for i := 0; i < max; i += size {
+		err := fn(&response, filter)
+		if err != nil {
+			return err
+		}
+		if response.GetDataLength() != 0 {
+			return nil
+		}
+
+		//adjust window
```
afiune, line 82: nit 
```suggestion
		// adjust window
```

api/v2_search_filters.go
```diff
@@ -54,3 +54,42 @@ type TimeFilter struct {
...
+	SetEndTime(*time.Time)
+}
+
+type search func(response interface{}, filters SearchableFilter) error
+
+// WindowedSearch performs a new search of a specific time frame size,
+// until response data is found or the max searchable days is reached
+func WindowedSearch(fn search, size int, max int, response SearchResponse, filter SearchableFilter) error {
+	for i := 0; i < max; i += size {
+		err := fn(&response, filter)
```
afiune, line 74: Are we not overriding the response every time we run this function?

afiune (changes requested):

cli/cmd/compliance_aws.go
dmurray-lacework: Yes, I'll create consts

api/v2_search_filters.go
dmurray-lacework: Yes, we are overwriting the response. But this is because the response result is empty. 

Example:
We're performing a search for a value "x"
We make a search request in the last 7 days, "x" is not found the response is empty `{ "data": [] }`
So we search again, this time moving the timeframe to start -14 days and end - 7 days
A result is found in that window, and then we return it.

api/v2_search_filters.go
```diff
@@ -54,3 +58,60 @@ type TimeFilter struct {
...
+}
+
+type SearchableFilter interface {
+	GetTimeFilter() *TimeFilter
+	SetStartTime(*time.Time)
+	SetEndTime(*time.Time)
+}
+
+// V2ApiMaxSearchHistory defines the maximum number of days in the past api v2 allows to be searched
+const V2ApiMaxSearchHistory = 92 //days
```
afiune, line 73: lets add "days" in the variable name rather than a comment
```suggestion
const V2MaxSearchHistoryDays = 92 //days
```
something like that.

api/v2_search_filters.go
```diff
@@ -54,3 +58,60 @@ type TimeFilter struct {
...
+
+// V2ApiMaxSearchWindow defines the maximum number of days in a single request api v2 allows to be searched
+const V2ApiMaxSearchWindow = 7 //days
```
afiune, line 76: same here ☝🏽

afiune (approved):
The only think we could improve is the function name, but out of that, this is great.

tenor-52311597

api/v2_search_filters.go
afiune: any of these should work
```go
func WindowedSearchFirst(fn search, size int, max int, response SearchResponse, filter SearchableFilter) error {
```
```go
func WindowedSearchAndReturnFirst(fn search, size int, max int, response SearchResponse, filter SearchableFilter) error {
```
```go
func WindowedFindFirst(fn search, size int, max int, response SearchResponse, filter SearchableFilter) error {
```
```go
func WindowedSearchForFirst(fn search, size int, max int, response SearchResponse, filter SearchableFilter) error {
```
```go
func SearchFirst(fn search, size int, max int, response SearchResponse, filter SearchableFilter) error {
```
```go
func SearchOccurrence(fn search, size int, max int, response SearchResponse, filter SearchableFilter) error {
```

PR #984, opened 2022-10-26
fix(cdk): parse global flags when running components

afiune:
## Summary

Before this change, when someone would try to use global flags such as `--profile`, `--debug`, etc. while running a component, the CLI would bypass the parsing of all flags and it would send all arguments directly to the underlying component. This caused problems when switching profiles, activating debug logs, etc. (no global flag was working)

This change is implementing a parsing mechanism that will filter all real global flags from the provided arguments and pass the rest as arguments to the component. The filtered global flags are also parsed and applied.

Wrote unit tests for the filter function, note that I mocked a few things such as a couple structs and global
flags that will be generated at runtime by the underlying [Go package we use named cobra.](https://github.com/spf13/cobra)

## Issue

https://lacework.atlassian.net/browse/ALLY-1243

### DIFF
--- a/cli/cmd/component.go
+++ b/cli/cmd/component.go
@@ -22,12 +22,14 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 
 	"github.com/AlecAivazis/survey/v2"
 	"github.com/fatih/color"
 	"github.com/olekukonko/tablewriter"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
 
 	"github.com/lacework/go-sdk/lwcomponent"
 )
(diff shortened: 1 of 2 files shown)


cli/cmd/component.go
```diff
@@ -158,16 +160,40 @@ func (c *cliState) LoadComponents() {
...
+						globalFlags := []*pflag.Flag{}
+						cmd.Flags().VisitAll(func(f *pflag.Flag) {
+							// At runtime, we visit all global flags defined at the root command
+							// and we essentially make a slice of the flag names with their shorthand
+							// so that we can remove them all from the raw arguments (`args`)
+							globalFlags = append(globalFlags, f)
+						})
+
+						// Use the list of global CLI flags to filter the provided arguments and return
+						// the filteres arguments to pass to the underlying component and the real list
```
dmurray-lacework, line 176:
```suggestion
						// the filters arguments to pass to the underlying component and the real list
```

cli/cmd/component.go
```diff
@@ -158,16 +160,40 @@ func (c *cliState) LoadComponents() {
...
+						// of CLI flags. The later is used to parse and then run the global CLI init func
```
dmurray-lacework, line 177:
```suggestion
						// of CLI flags. The latter is used to parse and then run the global CLI init func
```

dmurray-lacework (approved):

cli/cmd/component.go
```diff
@@ -158,16 +160,40 @@ func (c *cliState) LoadComponents() {
...
+
+						cli.Log.Debugw("running component", "component", cmd.Use,
+							"args", filteredArgs, "cli_flags", filteredCLIFlags)
 						f, ok := cli.LwComponents.GetComponent(cmd.Use)
 						if ok {
 							// @afiune what if the component needs other env variables
 							envs := []string{fmt.Sprintf("LW_COMPONENT_NAME=%s", cmd.Use)}
 							envs = append(envs, c.envs()...)
-							return f.RunAndOutput(args, envs...)
+							return f.RunAndOutput(filteredArgs, envs...)
```
hazedav, line 196: @afiune , How does a component go about respecting some of these global args (i.e. `--json`)?  Should the component specify (i.e. via specs.json) whether or not args should be filtered?

afiune: I think components should load these options from environmental variables.

hazedav: @afiune , I think this makes sense.  Just to clarify I'm not seeing the CLI consistently set global flags as environment variables.  For instance, on `0.44.1` when I specify:

```
lacework remediate alert 1233214 --nocolor --debug --nocache
```

My environment looks like:
```
[HOSTNAME=f3829371e1fe LW_SUBACCOUNT= LW_API_TOKEN=_007b8a090893066043c34dd85778023e PWD=/ HOME=/root TERM=xterm SHLVL=1 LW_ACCOUNT=dev7.dev7.corp PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin _=/usr/local/bin/lacework NO_COLOR=true LW_COMPONENT_NAME=remediate LW_API_KEY= LW_API_SECRET=]
```

I get `NO_COLOR=true`, but no environment flags around `DEBUG` or `NO_CACHE`.

afiune: yup, good catch!

PR #987, opened 2022-10-27
docs: update compliance get-reports --type valid inputs help text

dmurray-lacework:
update valid types in help text for compliance `--type` flag


dmurray-lacework:
This is no longer relevant

PR #994, opened 2022-10-31
fix(cli): do not `aws-install` on nil runners

nschmeller:
Currently, if the CLI can find an instance but is not able to connect to
it and create a `lwrunner.AWSRunner` struct to represent it, it will
still try to install an agent on that instance.

This commit:
* Changes the parallelism code to not add the runner if it does not
successfully initialize
* Addresses the root cause that triggered this problem by adjusting the
AMI username heuristic for Amazon Linux 2
* Removes outdated comments around the parallelism code area

Fixes RAIN-39248
Fixes RAIN-39251

Updating integration tests to add runners across OSes.

## Issue

https://lacework.atlassian.net/browse/RAIN-39248
https://lacework.atlassian.net/browse/RAIN-39251

### DIFF
--- a/lwrunner/awsrunner.go
+++ b/lwrunner/awsrunner.go
@@ -63,7 +63,8 @@ func NewAWSRunner(amiImageId, host, region, availabilityZone, instanceID string,
 	var user string
 	if strings.Contains(*result.Images[0].Name, "ubuntu") {
 		user = "ubuntu"
-	} else if strings.Contains(*result.Images[0].Name, "amazon_linux") {
+	} else if strings.Contains(*result.Images[0].Name, "amazon_linux") ||
+		strings.Contains(*result.Images[0].Name, "amzn2-ami") {
 		user = "ec2-user"
 	} else {
 		return nil, fmt.Errorf("expected either Ubuntu or Amazon Linux 2 AMI, got AMI %s", *result.Images[0].Name)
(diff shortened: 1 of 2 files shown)


cirego (changes requested):
Looks good to me for fixing the crash that we saw with `aws-install ec2ic`.

I may be missing something because I think we still need to address the issue with `aws-install ssh` filtering out ec2 instances due to this heuristic.

lwrunner/awsrunner.go
```diff
@@ -61,16 +61,17 @@ func NewAWSRunner(amiImageId, host, region, availabilityZone, instanceID string,
...
+	imageName := *result.Images[0].Name
+	if os.Getenv("LW_SSH_USER") != "" {
+		user = os.Getenv("LW_SSH_USER")
+	} else if strings.Contains(imageName, "ubuntu") {
 		user = "ubuntu"
-	} else if strings.Contains(*result.Images[0].Name, "amazon_linux") {
+	} else if strings.Contains(imageName, "amazon_linux") {
+		user = "ec2-user"
+	} else if strings.Contains(imageName, "amzn2-ami") {
 		user = "ec2-user"
```
cirego, line 72: You might consider making this a lookup table, instead of having multiple `else if` statements.

cirego (approved):
Looks good to me!

lwrunner/awsrunner.go
nschmeller: How would this work with the `string.Contains` logic? Would we populate the lookup table with all the combinations of strings that will return a hit?

cirego: You can iterate over the map instead of performing a seek.

nschmeller: Done in 81c2c751a662a06174f728c9faee0f519c8a283e

PR #1005, opened 2022-11-07
fix(cdk) pass valid LW_API_TOKEN to components

slshen:
## Summary

When the cached token expires running `lacework iac ...` does not pass LW_API_TOKEN.

With the iac component installed, I run `lacework access-token -d 10` to generate a very short lived token.

Initially a command like `lacework iac org list` will succeed.  But after ~10 seconds you will get:

```
samshen@Sams-Lacework-Laptop go-sdk % lacework iac org list
[Error] GET https://api.demo.soluble.cloud/api/v1/users/profile returned 401 in 463ms
[Error] {
  "ok" : false,
  "status" : 401,
  "error" : "Unauthorized",
  "timestamp" : "2022-11-07T17:21:12.923519Z",
  "message" : "401"
}
[ Info] Are you not logged in?  Use lacework iac auth profile to verify.
[ Info] See https://docs.lacework.com/iac/ for more information.
Error: https://api.demo.soluble.cloud/api/v1/users/profile returned 401
ERROR unable to run component: exit status 1
```

You can look under `~/.config/lacework/cache` to see that the token has not been refreshed.

## Issue

RAIN-42604

### DIFF
--- a/cli/cmd/cache.go
+++ b/cli/cmd/cache.go
@@ -43,7 +43,6 @@ const MaxCacheSize = 1024 * 1024 * 1024
 // cli.Cache.Write("data", []byte("cool update"))     // Update
 // cli.Cache.Erase("data")                            // Delete
 // ```
-//
 func (c *cliState) InitCache(d ...string) {
 	if len(d) == 0 {
 		dir, err := cache.CacheDir()
@@ -151,7 +150,7 @@ func (c *cliState) WriteCachedToken() error {
 		return nil
 	}
 
-	if c.Token == "" {
+	if c.Token == "" || c.tokenCache.ExpiresAt.Before(time.Now().Add(-10*time.Second)) {
 		response, err := c.LwApi.GenerateToken()
 		if err != nil {
 			return err
@@ -232,9 +231,11 @@ func (c *cliState) WriteAssetToCache(key string, expiresAt time.Time, data inter
 //
 // ```go
 // var report vulnReport
-// if expired := cli.ReadCachedAsset("my-report", &report); !expired {
-//     fmt.Printf("My report: %v\n", report)
-// }
+//
+//	if expired := cli.ReadCachedAsset("my-report", &report); !expired {
+//	    fmt.Printf("My report: %v\n", report)
+//	}
+//
 // ```
 func (c *cliState) ReadCachedAsset(key string, data interface{}) bool {
 	if c.noCache {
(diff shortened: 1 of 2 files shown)


cli/cmd/component.go
```diff
@@ -180,7 +180,7 @@ func (c *cliState) LoadComponents() {
 						// Parse all global CLI flags provided (filtered) by the user, then run the global
 						// CLI init function to initialize our logger, api client, and other global config
 						err := cmd.Flags().Parse(filteredCLIFlags)
-						initConfig() // @afiune NOTE we purposely run this func first and then check the err
```
slshen, line 183: initConfig() has already been run by the persistent pre run on root; I don't see why it has to be run again.

Running it again has the undesirable side-effect of clearing out cli.Token.

cli/cmd/cache.go
```diff
@@ -151,7 +150,7 @@ func (c *cliState) WriteCachedToken() error {
 		return nil
 	}
 
-	if c.Token == "" {
+	if c.Token == "" || c.tokenCache.ExpiresAt.Before(time.Now().Add(-10*time.Second)) {
```
slshen, line 153: I believe the intent of this code is to ensure cliState has a valid token.

hazedav:
@afiune ,

I think that as part of a component's spec it should be allowed to communicate to the CLI that it wants a token with a certain TTL/expiry.  This will ensure that the CLI gives the component a token that will be valid for an appropriate amount of time.

Thoughts?

David

cli/cmd/component.go
afiune: We need this if-not none of the CLI global flags will be parsed.

afiune: Instead of removing `initConfig()` add this on line `189`
```go
if err := cli.NewClient(); err != nil {
	return err
}
```

afiune:
@hazedav I could see a case where we need such communication, but today, it is not a must have but more a nice to have. Lets add notes to https://lacework.atlassian.net/browse/ALLY-1026

afiune (approved):
tenor-84740076

hazedav:
> @hazedav I could see a case where we need such communication, but today, it is not a must have but more a nice to have. Lets add notes to https://lacework.atlassian.net/browse/ALLY-1026

May not be needed as long as the token properly refreshes itself once a valid client is established...

PR #1009, opened 2022-11-08
refactor(cli): fixup + test `aws-install` SSH user

nschmeller:
This commit refactors and adds unit tests for the SSH username
assignment logic in the `aws-install` CLI subcommand.

Specifically, this commit:
* Breaks down the AMI image name fetch and the lookup table code into
  helper functions
* Adds unit tests for these helper functions
* Enforces using the username passed as the `--ssh_username` CLI arg

Fixes RAIN-42812

Added unit tests, ran integration tests from the agent repo

## Issue

https://lacework.atlassian.net/browse/RAIN-42812

### DIFF
--- a/lwrunner/awsrunner.go
+++ b/lwrunner/awsrunner.go
@@ -38,48 +38,17 @@ type AWSRunner struct {
 	InstanceID       string
 }
 
-func NewAWSRunner(amiImageId, host, region, availabilityZone, instanceID string, callback ssh.HostKeyCallback) (*AWSRunner, error) {
+func NewAWSRunner(amiImageId, userFromCLIArg, host, region, availabilityZone, instanceID string, callback ssh.HostKeyCallback) (*AWSRunner, error) {
 	// Look up the AMI name of the runner
-	cfg, err := config.LoadDefaultConfig(context.Background())
+	imageName, err := getAMIName(amiImageId, region)
 	if err != nil {
 		return nil, err
-	}
-	cfg.Region = region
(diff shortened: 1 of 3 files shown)


lwrunner/awsrunner.go
```diff
@@ -144,3 +113,65 @@ func (run AWSRunner) SendPublicKey(pubBytes []byte) error {
...
+	return "", fmt.Errorf("no SSH username found for AMI %s, set as arg or shell env", imageName)
+}
+
+// getSSHUsernameLookupTable returns a lookup table for heuristically
+// determining SSH username based on AMI.
+// The first row of the table it returns is a function that checks
+// `LW_SSH_USER` in the shell environment.
+func getSSHUsernameLookupTable() []func(string) (bool, string) {
+	return []func(string) (bool, string){
+		func(_ string) (bool, string) { return os.Getenv("LW_SSH_USER") != "", os.Getenv("LW_SSH_USER") }, // THIS ROW MUST BE FIRST IN THE TABLE
```
cirego, line 170: I would be tempted to move the `LW_SSH_USER` check between lines 153 and 154. Otherwise it could be really easy to introduce a regression by reordering the elements in this array and/or adding a new AMI type before the LW_SSH_USER check.

cirego (approved):

lwrunner/awsrunner.go
nschmeller: I'd agree, but we test this case in a unit test. So I'd like to keep everything all neatly packaged up if possible

PR #1010, opened 2022-11-09
feat(cli)!: pass agent token as flag for `aws-install`

nschmeller:
This commit switches the method of passing the agent token to
`aws-install`, and is a breaking change for the CLI.

The old method that was released in `v0.43.0` was to pass it as
a required arg. The new method in this PR will rely on passing the
token as an optional flag (`--token`) and asking the user to provide
it interactively if that flag is not present.

Fixes RAIN-42457

## Summary

We initially decided to release the `aws-install` subcommand with the agent token as a required arg. We've changed our minds on this and have decided that it's better to ask the customer interactively (once) to select an agent token, or to explicitly pass it as a flag. The initial rationale to pass the agent token as a required arg was based on a poor design decision around initializing AWS runners.

This is a breaking change, and we'd like to get it released before v1.0.0 ships.

Updating CTF integration tests in the agent repo

## Issue

https://lacework.atlassian.net/browse/RAIN-42457

### DIFF
--- a/cli/cmd/agent_aws-install_ec2ssh.go
+++ b/cli/cmd/agent_aws-install_ec2ssh.go
@@ -23,28 +23,35 @@ import (
 	"sync"
 
 	"github.com/gammazero/workerpool"
+	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 var (
 	agentInstallAWSSSHCmd = &cobra.Command{
-		Use:   "ec2ssh <token>",
-		Args:  cobra.ExactArgs(1),
+		Use:   "ec2ssh",
+		Args:  cobra.NoArgs,
 		Short: "Use SSH to securely connect to EC2 instances",
 		Long: `This command installs the agent on all EC2 instances in an AWS account
 using SSH.
 
 To filter by one or more regions:
 
-    lacework agent aws-install ec2ssh <token> --include_regions us-west-2,us-east-2
+    lacework agent aws-install ec2ssh --include_regions us-west-2,us-east-2
 
 To filter by instance tag:
 
-    lacework agent aws-install ec2ssh <token> --tag TagName,TagValue
+    lacework agent aws-install ec2ssh --tag TagName,TagValue
 
 To filter by instance tag key:
 
-    lacework agent aws-install ec2ssh <token> --tag_key TagName
+    lacework agent aws-install ec2ssh --tag_key TagName
+
+To provide an agent access token of your choice, use the command 'lacework agent token list',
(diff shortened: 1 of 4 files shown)


cli/cmd/agent_aws-install_ec2ssh.go
```diff
@@ -23,40 +23,47 @@ import (
...
+    lacework agent aws-install ec2ssh --tag TagName,TagValue
 
 To filter by instance tag key:
 
-    lacework agent aws-install ec2ssh <token> --tag_key TagName
+    lacework agent aws-install ec2ssh --tag_key TagName
+
+To provide an agent access token of your choice, use the command 'lacework agent token list',
+select a token and pass it to the '--token' flag. This flag must be selected if the
+'--noninteractive' flag is set.
```
cirego, line 52:
```suggestion
To provide an existing access token, use the '--token' flag. This flag is required
when running non-interactively (`--noninteractive` flag). The interactive command
`lacework agent token list` can be used to query existing tokens.
```

cli/cmd/agent_aws-install_ec2ssh.go
```diff
@@ -107,6 +117,21 @@ func init() {
...
+			// user didn't provide an agent token
+			cli.Log.Debugw("agent token not provided, asking user to select one now")
+			var err error
+			token, err = selectAgentAccessToken()
+			if err != nil {
+				return err
+			}
+		} else {
+			return errors.New("user did not provide or interactively select an agent token")
+		}
```
cirego, line 132:
```suggestion
		} else {
			return errors.New("--token flag is required when --noninteractive flag is set")
		}
```

integration/test_resources/help/agent_aws-install_ec2ssh
```diff
@@ -3,27 +3,33 @@ using SSH.
...
+    lacework agent aws-install ec2ssh --tag TagName,TagValue
 
 To filter by instance tag key:
 
-    lacework agent aws-install ec2ssh <token> --tag_key TagName
+    lacework agent aws-install ec2ssh --tag_key TagName
+
+To provide an agent access token of your choice, use the command 'lacework agent token list',
+select a token and pass it to the '--token' flag. This flag must be selected if the
+'--noninteractive' flag is set.
```
cirego, line 18:
```suggestion
To provide an existing access token, use the '--token' flag. This flag is required
when running non-interactively (`--noninteractive` flag). The interactive command
`lacework agent token list` can be used to query existing tokens.
```

cirego (approved):

PR #1013, opened 2022-11-09
fix(cli): minor fix related to refreshing tokens 

hazedav:
## Summary
Logic was recently added to refresh eminent/expired tokens.  This logic does not take into consideration whether or not we actually have a cached token…

Manually

## Issue
https://lacework.atlassian.net/browse/ALLY-1278

### DIFF
--- a/cli/cmd/cache.go
+++ b/cli/cmd/cache.go
@@ -145,12 +145,21 @@ func (c *cliState) ReadCachedToken() {
 	}
 }
 
+// return true if the cached token is expired or will expire within
+// the eminent duration (i.e. 10 seconds)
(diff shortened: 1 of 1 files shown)


afiune (approved):

cli/cmd/cache.go
```diff
@@ -145,12 +145,21 @@ func (c *cliState) ReadCachedToken() {
 	}
 }
 
+// return true if the cached token is expired or will expire within
+// the eminent duration (i.e. 10 seconds)
+func (c *cliState) cachedTokenExpiryEminent() bool {
+	eminentDuration := -10 * time.Second
```
slshen, line 151: I'm late to the party here but the word you want is "imminent".

PR #1014, opened 2022-11-09
fix(cdk): improve component arg parsing

slshen:
* Be more precise in parsing arguments for components

* Always pass --help flag to component instead of showing CLI help

* ~~Pass LW_PROFILE and LW_DEBUG env vars to components~~

## Summary

Be more precise in parsing arguments for components.  For example:

 * `lacework iac auth profile` should run `iac auth profile`
 * `lacework iac org --debug --help` should run `iac org --help` with debug logging on instead of printing lacework CLI help
 * `lacework iac tf-scan --debug -a qan -d ~/work/stuff --json --upload=false` should run `iac tf-scan -d ~/work/stuff --upload=false`

Also pass LW_PROFILE and LW_DEBUG env vars to components.

See unit test.  Also ran examples by hand.

## Issue

RAIN-42779


cli/cmd/cli_state.go
```diff
@@ -430,6 +430,8 @@ func (c *cliState) envs() []string {
 		fmt.Sprintf("LW_NOCOLOR=%s", os.Getenv("NO_COLOR")),
 		fmt.Sprintf("LW_LOG=%s", c.LogLevel),
 		fmt.Sprintf("LW_JSON=%v", c.jsonOutput),
+		fmt.Sprintf("LW_PROFILE=%s", c.Profile),
```
afiune, line 433: I don't think we should pass `LW_PROFILE` 

I think the CLI has the logic to parse the profile and pass the necessary information to
the component, the problem I have is for example, if we pass the profile foo via `--profile`
but also change the sub-account via `--subaccount`, how does the component know
when to use the profile and when to use the rest of the flags?

Do you think we should still pass it?

cli/cmd/cli_state.go
```diff
@@ -430,6 +430,8 @@ func (c *cliState) envs() []string {
...
+		fmt.Sprintf("LW_DEBUG=%v", c.LogLevel == "DEBUG"),
```
afiune, line 434: Instead of `LW_DEBUG` I proposed to pass `LW_LOG`.

`LW_LOG` is better since we could pass different log levels such as `INFO`, `DEBUG`, etc.

The `--debug` flag was introduced before we created the [lwlogger](https://github.com/lacework/go-sdk/tree/main/lwlogger) Go package. 

Do you still think we should pass both `LW_DEBUG` and `LW_LOG`?

afiune (changes requested):
The re implementation looks great but I have concerns about passing both environment variables you are adding. cc @hazedav What's your take on this?

cli/cmd/cli_state.go
slshen: To answer your narrow question the component knows because it looks at LW_SUBACCONT.

The IAC component has its own configuration which it stores in its own config file -- the most important config value is the IAC organization to use.  This will be the wrong IAC org if you run `lacework configure switch-profile`.

The IAC component can store which profile it was used with, and if it's different then it can do something other than proceed blindly ahead with what surely will be wrong values.

For the most part this are issues that only internal users (or those with multiple profiles) will hit.  But it'll be really confusing when they do hit them.

cli/cmd/cli_state.go
slshen: It's better to pass both I think on the idea that we forward all the global flags as environment variables.

I do have some concern with what happens when the CLI adds TRACE.  One answer is that it never does, which is fine.  But LW_DEBUG=true is very clear, and the CLI can still set LW_DEBUG=true even if it adds LW_LOG=TRACE.

cli/cmd/cli_state.go
hazedav: I don't know much about the iac component, but FWIW it would seem that `LW_PROFILE` is a CLI implementation detail and that `LW_ACCOUNT` and `LW_SUBACCOUNT` are what's important here...

cli/cmd/cli_state.go
hazedav: FWIW the `remediate` component just ended up using `LW_LOG`:
```
Log:        lwlogger.New("").Sugar(), // get from "LW_LOG"
```

And when I got my api client I used:
```
	c.Log.Debugw("initializing lacework api client")
	writer := &zapio.Writer{Log: c.Log.Desugar(), Level: zapLogLevel(c.LogLevel)}
	defer writer.Close()

	// Initialize Lacework API client
	apiopts := []api.Option{
		api.WithSubaccount(c.Subaccount),
		api.WithApiV2(),
		api.WithLogLevelAndWriter(c.LogLevel, writer),
	}
```

afiune (approved):
tenor-200856970

PR #1015, opened 2022-11-09
fix(cli): rootCmd to intelligently PreRun components 

hazedav:
ALLY-1279

## Summary
For CDK components which are type `CLI_COMMAND` the CLI automatically creates child commands for them.  The `RunE` portion of these child commands has special handling for global flags.  However, the `PersistentPreRunE` portion of the rootCmd performs tasks (i.e. establishes a NewClient or does Migration) without the knowledge of these flags.

Since `RunE` is essentially “pre run” for CDK components we should:

1. have the root command check if we’re running a component, and if so avoid certain tasks
2. perform the tasks avoided by the root command within the RunE of the component

Manually

## Issue
https://lacework.atlassian.net/browse/ALLY-1279

### DIFF
--- a/cli/cmd/component.go
+++ b/cli/cmd/component.go
@@ -34,6 +34,8 @@ import (
 	"github.com/lacework/go-sdk/lwcomponent"
 )
 
+const componentTypeAnnotation string = "component"
+
 var (
 	// componentsCmd represents the components command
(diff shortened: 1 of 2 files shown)


afiune (approved):
tenor-243708983

PR #1022, opened 2022-11-14
feat(cli): find and fix GCP instances w/o agents

nschmeller:
Tasks left to do at checkpoint (from https://lacework.atlassian.net/browse/RAIN-37376?focusedCommentId=171275):
- [x] Get credentials from environment (done: leveraging application default credentials)
- [x] Read GCP project ID from metadata server or CLI flag (done with 36e1940821e0a5c0ce7d2d6dc64618be622f2ce7)
- [x] Read GCP username from credentials or from CLI flag (done with dcddb8b0c358eb22d98927166a4dc2a2a8c79a4f)
- [x] Working `gcp-install` (done 0515c4698087ac01b0a0ab69546bc709efd94feb)
- [x] Add filters by tag, tag key, region, ~~and project allow / deny (skipList?)~~ (done with e17648be0ae9297a5c8c1e8fdf119575e84b65d5)
- [x] Write Long help text (done with f073485879412479ab535f39a41f5b2e44235af3)
- [x] Double check command-line flags are provided (done with 37c13db2ab99990c9c7b9c8fbeda7e1b8ae89f60)
- [x] Add CLI integration test for help text output (done with a0761a53e0bec97a01b9105c901012a3f342d8f1)
- [x] Add license headers to source files (done with d3685559f0ce86e4b26089dedba42d5ee8551852)
- [x] Test with agent integration tests (done with https://github.com/lacework/agent/pull/1556)

## Summary

GCP version of `aws-install`. 

Planning on testing with agent integration tests

## Issue

https://lacework.atlassian.net/browse/RAIN-37376


nschmeller:
Here's a design decision I'm making: how to get the org ID?

I can think of two solutions:
* Read the org ID from the GCE metadata service (following the pattern of Stitch: https://github.com/lacework-dev/sidekick/blob/c9adf188bf72b9bf3877c7c856c24cc6a735de91/cloud/gcp/resources/metadata/metadata.go#L25)
* If the CLI is not being run in a Cloud Shell or GCE instance, add the option to pass in the org ID as a CLI flag

I'm not entirely happy with this decision, because I wish that there were a service or trick to get org ID and other metadata from GCP credentials. But I've been looking for a few days now and I haven't found anything, so this is workable trick to keep on moving.

nschmeller:
Actually, https://pkg.go.dev/google.golang.org/api/cloudresourcemanager/v1beta1 seems like a pretty good service for this.

nschmeller:
Let's just installs within the project for now. Org can be a straightforward change that we make once we get some customer asks. We'll try to read project ID from the metadata server and if not, assume that the user passed it in as a CLI arg.

cirego:
> Let's just installs within the project for now. Org can be a straightforward change that we make once we get some customer asks. We'll try to read project ID from the metadata server and if not, assume that the user passed it in as a CLI arg.

Sounds good Nick. Happy to review once you make this change!

nschmeller:
I found https://pkg.go.dev/google.golang.org/api@v0.103.0/oslogin/v1, which is a deprecated version of OSLogin. I don't think we can use it, but it sure would be nice if we didn't need to programmatically supply the username.

cirego (approved):
:fire: :shipit:

nschmeller:
I merged in the integration tests PR. 

Last step to do before merging in this PR is to go through the Guardrails warnings and look for anything bad. I took a look, and all of the warnings are "transitive dependencies", which seems like it's GCP's job to fix. I thought of fixing the dependencies that had "high" severity, but in all of those dependencies, there was no fixed patch available. So I think we should merge in the PR and let those 96 Guardrails warnings be added to the pre-existing 145 vulnerabilities. 

So I think we're ready to merge!

cirego:
> I merged in the integration tests PR.
> 
> Last step to do before merging in this PR is to go through the Guardrails warnings and look for anything bad. I took a look, and all of the warnings are "transitive dependencies", which seems like it's GCP's job to fix. I thought of fixing the dependencies that had "high" severity, but in all of those dependencies, there was no fixed patch available. So I think we should merge in the PR and let those 96 Guardrails warnings be added to the pre-existing 145 vulnerabilities.
> 
> So I think we're ready to merge!

Ship it!

PR #1027, opened 2022-11-18
fix(cdk): component arg/flag parsing

afiune:
## Summary

The new component arg/flag parsing is not letting flags like `-foo` or `-bar` pass to the component.

Added tests that shows the issue.

## Issue

N/A

### DIFF
--- a/cli/cmd/component_args.go
+++ b/cli/cmd/component_args.go
@@ -45,6 +45,7 @@ func (p *componentArgParser) parseArgs(globalFlags *pflag.FlagSet, args []string
 			p.componentArgs = append(p.componentArgs, s)
 			continue
 		}
+
 		if len(s) == 0 || s[0] != '-' || len(s) == 1 {
 			// not a flag, passthrough
 			p.componentArgs = append(p.componentArgs, s)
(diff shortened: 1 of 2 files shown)


cli/cmd/component_args.go
```diff
@@ -145,7 +146,7 @@ func (p *componentArgParser) parseShortArg(flags *pflag.FlagSet, s string, args
 			p.cliArgs = append(p.cliArgs, s)
 		} else if len(shorthands) > 1 {
 			// '-farg'
-			p.cliArgs = append(p.cliArgs, s)
+			p.componentArgs = append(p.componentArgs, s)
```
slshen, line 149: So cobra accepts shorthand args in the form -farg, meaning -source for the lacework CLI is equivalent to -s ource which is the same thing as --api_secret ource.

So the code before the patch was doing the right thing I think.  It may look a little strange, but what we agreed upon was the the lacework CLI will consume the flags that it understands, and because it understands -s then it does indeed consume -source.

A component can't redefine the lacework CLI's to mean something different.  In this case it would have to define --source as a long flag.

I don't think we can special case this without introducing divergent behavior depending on whether or not command in builtin the the CLI or lives in a component.   Which is to say this:

lacework query -s_7cedxxxxx ...

Would do one thing (use _7xedxxxx as the api secret), but:

lacework iac ... -s _7cedxxxxx

Would likely error out as the iac component doesn't have a -s shorthand flag.

afiune: You are so right. Thanks for reminding me about what we agreed. I appreciate your feedback.

PR #1037, opened 2022-11-25
refactor: remove all unused functions, types and tests

dmurray-lacework:
Remove all remaining unused functions, types and tests left over from v1 migration.


djmctavish (approved):
Big tidy up LGTG

PR #1045, opened 2022-11-29
feat(cli): CTF `aws-install ec2ssm`

nschmeller:
This PR contains the implementation for the `ec2ssm` subcommand of Project Capture
The Flag.

Fixes RAIN-38281

## Summary

Install agents on EC2 instances using SSM.

Integration test at https://github.com/lacework/agent/pull/1699

## Issue

https://lacework.atlassian.net/browse/RAIN-38281


nschmeller:
Design decision: whether to use `Command`s or shells for SSM...

I feel like a Command might be the kosher way of doing things with SSM. In order to use Commands, I think we'd have to create the Commands as infrastructure, and we'd have to have a Command for each operation we want to do (verify access, check if the agent is already installed, and install the agent). But that would be a lot of overhead to set up. 

Alternatively, we could use shells. I think that would be a lot simpler. But it might be more difficult to audit / less official. 

I haven't made up my mind yet, but I'm going to first investigate using shells, and then take a look at Commands if shells don't fit just right.

nschmeller:
Another design decision: using Pulumi or Terraform to create SSM Documents / Commands and the IAM role?

I don't know enough about Pulumi to completely answer this. But my impression of Pulumi is that we won't have to specify separate Terraform files that we then apply and maintain state for. I think that Pulumi will allow us to specify infrastructure in the CLI code, apply that infra during the CLI execution, and maintain state in memory. And if all of those things are true, then I think Pulumi is the right tool for the job here. 

Plus points for Terraform are that I'm familiar with it, and that it seems a little more established than Pulumi, so customers might be familiar with it too.

nschmeller:
I took a look at the GuardRails issues, the ones that are "Medium" or "High" all do not have a patch available. So I guess we can ignore them.

nschmeller:
TODO at checkpoint for today: attach policies to the role, attach the role to the instance profile, test the role (Terratest?)

nschmeller:
Design decision: how to handle pre-existing instance profiles?

Annoyingly, AWS has a rule that an EC2 instance can only have one instance profile associated with it. And, an instance profile can only have one role associated with it. So, if we want to install an agent on an EC2 instance, we had better hope there isn't an instance profile already associated with it. 

Two ways we could approach the problem when the instance already has an instance profile associated with it:
* Remove the existing instance profile, add ours, run the operation, remove ours, re-add the old instance profile
  * This would be pretty risky: 1) the customer may need the instance profile to stay attached to the instance continuously, 2) if the `aws-install ssm` execution fails or is interrupted before we can clean up and re-add the old instance profile, we've just vandalized the customer's infrastructure
* Quietly fail on the instance. Give it a best-effort try, but if there's an instance profile, then we can't install an agent.

I'm thinking we should take the second option, since it's a lot less risky.

nschmeller:
Getting a strange error when associating the instance profile to the instance:
```
{
  "level": "debug",
  "ts": "2022-12-01T20:25:17Z",
  "caller": "cmd/agent_aws-install_ssm.go:151",
  "msg": "failed to attach instance profile to runner",
  "error": "operation error EC2: AssociateIamInstanceProfile, https response error StatusCode: 400, RequestID: c4d2ef17-efeb-4745-835a-d2ea1e100ff7, api error InvalidParameterValue: Value (arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile) for parameter iamInstanceProfile.arn is invalid. Invalid IAM Instance Profile ARN",
  "instance ID": "i-0f273822d1847344e",
  "role": {
    "Arn": "arn:aws:iam::561021084946:role/Lacework-Agent-SSM-Install-Role",
    "CreateDate": "2022-12-01T20:25:16Z",
    "Path": "/",
    "RoleId": "AROAYFH3WMUJEXV4PGX2X",
    "RoleName": "Lacework-Agent-SSM-Install-Role",
    "AssumeRolePolicyDocument": "%7B%0A%09%22Version%22%3A%20%222012-10-17%22%2C%0A%09%22Statement%22%3A%20%5B%0A%09%09%7B%0A%09%09%09%22Effect%22%3A%20%22Allow%22%2C%0A%09%09%09%22Principal%22%3A%20%7B%20%22AWS%22%3A%20%22arn%3Aaws%3Asts%3A%3A561021084946%3Aassumed-role%2FIntegrationTestsRole%2Fnick_integration_tests_role%22%20%7D%2C%0A%09%09%09%22Action%22%3A%20%22sts%3AAssumeRole%22%0A%09%09%7D%0A%09%5D%0A%7D",
    "Description": null,
    "MaxSessionDuration": null,
    "PermissionsBoundary": null,
    "RoleLastUsed": null,
    "Tags": null
  },
  "instance profile": {
    "Arn": "arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile",
    "CreateDate": "2022-12-01T20:25:17Z",
    "InstanceProfileId": "AIPAYFH3WMUJBLWYT3FFP",
    "InstanceProfileName": "Lacework-Agent-SSM-Install-Instance-Profile",
    "Path": "/",
    "Roles": [],
    "Tags": null
  }
}
```
When I use these values in the CLI (`aws ec2 associate-iam-instance-profile --iam-instance-profile Arn=arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile --instance-id i-07d5abe315618b5aa`) it works fine. 

Still digging into this...

nschmeller:
I managed to reproduce the above error in the CLI:
```
❯ aws ec2 associate-iam-instance-profile --iam-instance-profile Arn=arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile --instance-id i-07d5abe315618b5aa

An error occurred (InvalidParameterValue) when calling the AssociateIamInstanceProfile operation: Value (arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile) for parameter iamInstanceProfile.arn is invalid. Invalid IAM Instance Profile ARN
```

This happened when the given instance profile didn't exist. So now I'm going to check if the CLI is deleting the instance profile before we can use it...

nschmeller:
I confirmed that the instance profile we're creating does exist when we try to associate it with the instance. 
```{
  "level": "debug",
  "ts": "2022-12-01T21:37:14Z",
  "caller": "cmd/agent_aws-install_ssm.go:175",
  "msg": "DEBUG instance profile associations",
  "output": null,
  "error": "operation error EC2: AssociateIamInstanceProfile, https response error StatusCode: 400, RequestID: 2fc208b9-2d72-44b7-9bcf-ec1facd8c56b, api error InvalidParameterValue: Value (arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile) for parameter iamInstanceProfile.arn is invalid. Invalid IAM Instance Profile ARN",
  "instance profile": {
    "Arn": "arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile",
    "CreateDate": "2022-12-01T21:37:13Z",
    "InstanceProfileId": "AIPAYFH3WMUJA6X5HDLBC",
    "InstanceProfileName": "Lacework-Agent-SSM-Install-Instance-Profile",
    "Path": "/",
    "Roles": [],
    "Tags": null
  },
  "fresh lookup of instance profile": {
    "InstanceProfile": {
      "Arn": "arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile",
      "CreateDate": "2022-12-01T21:37:13Z",
      "InstanceProfileId": "AIPAYFH3WMUJA6X5HDLBC",
      "InstanceProfileName": "Lacework-Agent-SSM-Install-Instance-Profile",
      "Path": "/",
      "Roles": [
        {
          "Arn": "arn:aws:iam::561021084946:role/Lacework-Agent-SSM-Install-Role",
          "CreateDate": "2022-12-01T21:37:13Z",
          "Path": "/",
          "RoleId": "AROAYFH3WMUJNE2BTZZUW",
          "RoleName": "Lacework-Agent-SSM-Install-Role",
          "AssumeRolePolicyDocument": "%7B%22Version%22%3A%222012-10-17%22%2C%22Statement%22%3A%5B%7B%22Effect%22%3A%22Allow%22%2C%22Principal%22%3A%7B%22AWS%22%3A%22arn%3Aaws%3Asts%3A%3A561021084946%3Aassumed-role%2FIntegrationTestsRole%2Fnick_integration_tests_role%22%7D%2C%22Action%22%3A%22sts%3AAssumeRole%22%7D%5D%7D",
          "Description": null,
          "MaxSessionDuration": null,
          "PermissionsBoundary": null,
          "RoleLastUsed": null,
          "Tags": null
        }
      ],
      "Tags": []
    },
    "ResultMetadata": {}
  }
}
```

Interestingly, if I create an instance profile by hand (using the AWS CLI), and I hardcode its ARN into the CLI debugging code I added, it works fine...

```
{
  "level": "debug",
  "ts": "2022-12-01T21:48:50Z",
  "caller": "cmd/agent_aws-install_ssm.go:180",
  "msg": "DEBUG instance profile associations",
  "output": {
    "IamInstanceProfileAssociation": {
      "AssociationId": "iip-assoc-05ffd85cf5eda9774",
      "IamInstanceProfile": {
        "Arn": "arn:aws:iam::561021084946:instance-profile/Test-Instance-Profile-Debugging",
        "Id": "AIPAYFH3WMUJLTNO3GRGU"
      },
      "InstanceId": "i-05dcbf5301fb8e547",
      "State": "associating",
      "Timestamp": null
    },
    "ResultMetadata": {}
  },
  "error": null,
  "instance profile": {
    "Arn": "arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile",
    "CreateDate": "2022-12-01T21:48:50Z",
    "InstanceProfileId": "AIPAYFH3WMUJGKNSFD47T",
    "InstanceProfileName": "Lacework-Agent-SSM-Install-Instance-Profile",
    "Path": "/",
    "Roles": [],
    "Tags": null
  },
  "fresh lookup of instance profile": {
    "InstanceProfile": {
      "Arn": "arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile",
      "CreateDate": "2022-12-01T21:48:50Z",
      "InstanceProfileId": "AIPAYFH3WMUJGKNSFD47T",
      "InstanceProfileName": "Lacework-Agent-SSM-Install-Instance-Profile",
      "Path": "/",
      "Roles": [
        {
          "Arn": "arn:aws:iam::561021084946:role/Lacework-Agent-SSM-Install-Role",
          "CreateDate": "2022-12-01T21:48:50Z",
          "Path": "/",
          "RoleId": "AROAYFH3WMUJCAO4CVJ6W",
          "RoleName": "Lacework-Agent-SSM-Install-Role",
          "AssumeRolePolicyDocument": "%7B%22Version%22%3A%222012-10-17%22%2C%22Statement%22%3A%5B%7B%22Effect%22%3A%22Allow%22%2C%22Principal%22%3A%7B%22AWS%22%3A%22arn%3Aaws%3Asts%3A%3A561021084946%3Aassumed-role%2FIntegrationTestsRole%2Fnick_integration_tests_role%22%7D%2C%22Action%22%3A%22sts%3AAssumeRole%22%7D%5D%7D",
          "Description": null,
          "MaxSessionDuration": null,
          "PermissionsBoundary": null,
          "RoleLastUsed": null,
          "Tags": null
        }
      ],
      "Tags": []
    },
    "ResultMetadata": {}
  }
}
```

nschmeller:
Looks like it's a problem with the way we're creating the instance profile -- when I create it via the CLI, it works... so it's something about the way we're creating it in CTF
```
{
  "level": "debug",
  "ts": "2022-12-01T23:18:43Z",
  "caller": "cmd/agent_aws-install_ssm.go:180",
  "msg": "DEBUG instance profile associations",
  "output": {
    "IamInstanceProfileAssociation": {
      "AssociationId": "iip-assoc-0bd8f9a41fd68573b",
      "IamInstanceProfile": {
        "Arn": "arn:aws:iam::561021084946:instance-profile/Test-Debugging-Instance-Profile",
        "Id": "AIPAYFH3WMUJDZTUEWUK4"
      },
      "InstanceId": "i-047d98483e704633a",
      "State": "associating",
      "Timestamp": null
    },
    "ResultMetadata": {}
  },
  "error": null,
  "instance profile": {
    "Arn": "arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile",
    "CreateDate": "2022-12-01T23:18:42Z",
    "InstanceProfileId": "AIPAYFH3WMUJLSHXL6MZT",
    "InstanceProfileName": "Lacework-Agent-SSM-Install-Instance-Profile",
    "Path": "/",
    "Roles": [],
    "Tags": null
  },
  "fresh lookup of instance profile": {
    "InstanceProfile": {
      "Arn": "arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile",
      "CreateDate": "2022-12-01T23:18:42Z",
      "InstanceProfileId": "AIPAYFH3WMUJLSHXL6MZT",
      "InstanceProfileName": "Lacework-Agent-SSM-Install-Instance-Profile",
      "Path": "/",
      "Roles": [
        {
          "Arn": "arn:aws:iam::561021084946:role/Lacework-Agent-SSM-Install-Role",
          "CreateDate": "2022-12-01T23:18:42Z",
          "Path": "/",
          "RoleId": "AROAYFH3WMUJHUL37ED5R",
          "RoleName": "Lacework-Agent-SSM-Install-Role",
          "AssumeRolePolicyDocument": "%7B%22Version%22%3A%222012-10-17%22%2C%22Statement%22%3A%5B%7B%22Effect%22%3A%22Allow%22%2C%22Principal%22%3A%7B%22AWS%22%3A%22arn%3Aaws%3Asts%3A%3A561021084946%3Aassumed-role%2FIntegrationTestsRole%2Fnick_integration_tests_role%22%7D%2C%22Action%22%3A%22sts%3AAssumeRole%22%7D%5D%7D",
          "Description": null,
          "MaxSessionDuration": null,
          "PermissionsBoundary": null,
          "RoleLastUsed": null,
          "Tags": null
        }
      ],
      "Tags": []
    },
    "ResultMetadata": {}
  }
}
```

nschmeller:
I'm trying to figure out what is different between the instance profile I create by hand and the one that CTF creates.

I thought it might be that the ARN is much longer for the one that CTF creates, so I tried creating the one by hand with a much longer name, and it worked fine. So that's that theory done.

```
{
  "level": "debug",
  "ts": "2022-12-01T23:50:42Z",
  "caller": "cmd/agent_aws-install_ssm.go:181",
  "msg": "DEBUG instance profile associations",
  "output": {
    "IamInstanceProfileAssociation": {
      "AssociationId": "iip-assoc-00f76fda818afbfd8",
      "IamInstanceProfile": {
        "Arn": "arn:aws:iam::561021084946:instance-profile/Test-Debugging-Instance-Profile-Additional-Chars-Lacework-Agent-SSM-Install-Instance-Profile",
        "Id": "AIPAYFH3WMUJOMAWW5W5X"
      },
      "InstanceId": "i-04540037809aae452",
      "State": "associating",
      "Timestamp": null
    },
    "ResultMetadata": {}
  },
  "error": null,
  "instance profile": {
    "Arn": "arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile",
    "CreateDate": "2022-12-01T23:18:42Z",
    "InstanceProfileId": "AIPAYFH3WMUJLSHXL6MZT",
    "InstanceProfileName": "Lacework-Agent-SSM-Install-Instance-Profile",
    "Path": "/",
    "Roles": [
      {
        "Arn": "arn:aws:iam::561021084946:role/Lacework-Agent-SSM-Install-Role",
        "CreateDate": "2022-12-01T23:18:42Z",
        "Path": "/",
        "RoleId": "AROAYFH3WMUJHUL37ED5R",
        "RoleName": "Lacework-Agent-SSM-Install-Role",
        "AssumeRolePolicyDocument": "%7B%22Version%22%3A%222012-10-17%22%2C%22Statement%22%3A%5B%7B%22Effect%22%3A%22Allow%22%2C%22Principal%22%3A%7B%22AWS%22%3A%22arn%3Aaws%3Asts%3A%3A561021084946%3Aassumed-role%2FIntegrationTestsRole%2Fnick_integration_tests_role%22%7D%2C%22Action%22%3A%22sts%3AAssumeRole%22%7D%5D%7D",
        "Description": null,
        "MaxSessionDuration": null,
        "PermissionsBoundary": null,
        "RoleLastUsed": null,
        "Tags": null
      }
    ],
    "Tags": []
  },
  "fresh lookup of instance profile": {
    "InstanceProfile": {
      "Arn": "arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile",
      "CreateDate": "2022-12-01T23:18:42Z",
      "InstanceProfileId": "AIPAYFH3WMUJLSHXL6MZT",
      "InstanceProfileName": "Lacework-Agent-SSM-Install-Instance-Profile",
      "Path": "/",
      "Roles": [
        {
          "Arn": "arn:aws:iam::561021084946:role/Lacework-Agent-SSM-Install-Role",
          "CreateDate": "2022-12-01T23:18:42Z",
          "Path": "/",
          "RoleId": "AROAYFH3WMUJHUL37ED5R",
          "RoleName": "Lacework-Agent-SSM-Install-Role",
          "AssumeRolePolicyDocument": "%7B%22Version%22%3A%222012-10-17%22%2C%22Statement%22%3A%5B%7B%22Effect%22%3A%22Allow%22%2C%22Principal%22%3A%7B%22AWS%22%3A%22arn%3Aaws%3Asts%3A%3A561021084946%3Aassumed-role%2FIntegrationTestsRole%2Fnick_integration_tests_role%22%7D%2C%22Action%22%3A%22sts%3AAssumeRole%22%7D%5D%7D",
          "Description": null,
          "MaxSessionDuration": null,
          "PermissionsBoundary": null,
          "RoleLastUsed": null,
          "Tags": null
        }
      ],
      "Tags": []
    },
    "ResultMetadata": {}
  }
}
```

nschmeller:
The one that CTF creates (this one has the role attached, that might be part of it):

```
❯ aws-vault exec integration-tests -- aws iam get-instance-profile --instance-profile-name Lacework-Agent-SSM-Install-Instance-Profile
❯ {
    "InstanceProfile": {
        "Path": "/",
        "InstanceProfileName": "Lacework-Agent-SSM-Install-Instance-Profile",
        "InstanceProfileId": "AIPAYFH3WMUJLSHXL6MZT",
        "Arn": "arn:aws:iam::561021084946:instance-profile/Lacework-Agent-SSM-Install-Instance-Profile",
        "CreateDate": "2022-12-01T23:18:42+00:00",
        "Roles": [
            {
                "Path": "/",
                "RoleName": "Lacework-Agent-SSM-Install-Role",
                "RoleId": "AROAYFH3WMUJHUL37ED5R",
                "Arn": "arn:aws:iam::561021084946:role/Lacework-Agent-SSM-Install-Role",
                "CreateDate": "2022-12-01T23:18:42+00:00",
                "AssumeRolePolicyDocument": {
                    "Version": "2012-10-17",
                    "Statement": [
                        {
                            "Effect": "Allow",
                            "Principal": {
                                "AWS": "arn:aws:sts::561021084946:assumed-role/IntegrationTestsRole/nick_integration_tests_role"
                            },
                            "Action": "sts:AssumeRole"
                        }
                    ]
                }
            }
        ],
        "Tags": []
    }
}
```

The one I created by hand:

```
{
    "InstanceProfile": {
        "Path": "/",
        "InstanceProfileName": "Test-Debugging-Instance-Profile-Additional-Chars-Lacework-Agent-SSM-Install-Instance-Profile",
        "InstanceProfileId": "AIPAYFH3WMUJOMAWW5W5X",
        "Arn": "arn:aws:iam::561021084946:instance-profile/Test-Debugging-Instance-Profile-Additional-Chars-Lacework-Agent-SSM-Install-Instance-Profile",
        "CreateDate": "2022-12-01T23:48:21+00:00",
        "Roles": [],
        "Tags": []
    }
}
```

nschmeller:
I had thought this was working, but it wasn't. I think I had just created the instance profile name via the AWS CLI instead of the SDK, so it worked fine. And it worked on repeat runs because I had commented out the code to tear down infra in CTF. So when I commented the teardown code back in, it worked on the first try, but then it tore down the instance profile that I created via the AWS CLI. So when I did a second try, it created the instance profile using the SDK, which then failed to associate.

nschmeller:
I just confirmed that when I associate the SDK-created instance profile to an instance using the AWS CLI, it works fine. So the problem is when I try to associate the SDK-created instance profile to an instance using the SDK. 

The other combinations (associate CLI-created instance profile using the SDK, associate SDK-created instance profile using the CLI, associate CLI-created instance profile using the CLI) all work. So strange...

nschmeller:
Another piece of interesting info: associating an SDK-created instance profile using the SDK works when the SDK is re-using the instance profile (aka, it was created on one SDK run and not torn down, then on another SDK run the SDK found it and reused it). 

When I did this, I also associated the instance profile to one instance using the AWS CLI. So on the second SDK run, it did not associate the already-associated instance profile, but it successfully associated the other two instance profiles that the SDK created in a previous run. I need to test if it works if I don't do this.

nschmeller:
Jackpot -- I just ran it without the complicating factor that I mentioned above (about associating with one instance but not the others). The same result happened. So it's either 1) re-using does not return the same result as creating, and re-using is correct, or 2) there's a timing issue, and the instance profile needs a few seconds (perhaps to become eventually-consistent?) before it's ready to be associated. 

I'll add a 1min sleep and see if that does anything. If it works with that's then it's (2). Otherwise I'll take a look at the reuse vs create logic.

nschmeller:
It's a timing thing -- I added a 1min sleep, and it worked without any reusing. Woohoo!!!!!!!

I tried seeing if we could do retries every second to see if the instance profile was settled in, but the only API I could think of for this, `get-instance-profile`, showed it was available right after creation (when it clearly isn't, since we can't associate it). So we just have to pick the smallest sleep time that will work 99% of the time and hope that's enough. 

Now the goal is to figure out what that time is. 30sec and 15sec worked. I'll do some more experiments tomorrow. But for now, this is working!!

nschmeller:
Another thought I have is, instead of sleeping for 30 seconds, we could do retries when associating the instance profile. This has the advantage that we don't need to sleep when creating the instance profile, but it might take longer to fail when associating the instance profile, and we might run into rate limits if all the goroutines are making API calls every second. 

This article (http://www.it-automation.com/2021/06/06/how-to-deal-with-eventual-consistency-in-AWS-IAM.html) has a pretty good writeup about the problem that we're facing here. Their suggestion was retries or sleep for 45 seconds (apparently this was what AWS recommended, but I haven't cross-referenced it). 

For now, I'm tempted to go with the sleep for 45 seconds option, since it's simpler, and it's working right now.

nschmeller:
Currently the machines I'm using to test this code are not working with SSM. I'm not sure if this is a problem with this code or a problem with the machines. I'm hoping it's a problem with this code. 

The error specifically is `operation error SSM: SendCommand, https response error StatusCode: 400, RequestID: 9ea1cda8-55f0-4f6a-a130-aa9cce4ba933, InvalidInstanceId: Instances [[i-0629883c23865f080]] not in a valid state for account 561021084946`. I tried to see if SSM was working in general for these machines, so I ran `aws ssm start-session`, and got `An error occurred (TargetNotConnected) when calling the StartSession operation: i-0629883c23865f080 is not connected.`. I reran `aws ssm start-session` with fresh machines, and I got the same error. This makes me think that there is a problem with SSM on the testing machines. 

This is just an update on progress. Next step is to figure out why SSM doesn't work for the testing machines (ideally it should work for all machines!).

nschmeller:
The last update I wrote on this PR was discussing why we were getting the error `InvalidInstanceId: Instances [[i-0629883c23865f080]] not in a valid state for account 561021084946`. I thought at the time that it was an error with the test machines because I reproduced the error without running CTF SSM. But actually it is most likely an error with CTF SSM, because the "instance is not connected" error is a blanket error for any IAM misconfiguration. 

I figured out a few more things in the last two days with this implementation. Although the same bug is occurring, hopefully it's now occurring for different reasons. The first realization was that I had forgotten to write code to actually assume the role we were creating. The second was that the calling user should not assume the role, but rather the EC2 instance should assume the role (which is done when associating the instance profile). This also meant I had to edit the trust relationship in the role's trust policy so that the EC2 service could assume instead of the calling user. 

These are two critical things to realize for a working implementation, but I'm still not sure why we're getting the `InvalidInstanceId` error...

nschmeller:
I think the IAM configuration may actually be correct in this case -- I created the IAM role and instance profile using CTF SSM, but didn't tear them down after. Then I manually attached that instance profile using the CLI, and then I successfully ran `aws ssm start-session` using the CLI. So I think CTF SSM is creating the correct IAM infra. This means that there are two possibilities for what's going wrong: 1) CTF SSM is not associating the instance profiles correctly, or 2) CTF is not calling `SendCommand` correctly. 

I tested (1) by rerunning CTF SSM (without the teardown) and manually running `aws ssm start-session` using the CLI on the instance that CTF SSM configured. This failed with the same error (`An error occurred (TargetNotConnected) when calling the StartSession operation: i-0df549e9d2380150a is not connected.`). So it looks like CTF SSM is not associating the instance profiles correctly.

nschmeller:
Now I don't have confidence that the IAM configuration is correct here. I reran the experiment I did above, where I created the IAM role and instance profile using CTF SSM and then manually associated and tested using the CLI. Previously SSM worked, but this time SSM did not work...

nschmeller:
Okay, I'm going to do the CTF SSM steps using the CLI, and if that works, bisect down to what part breaks. 

Steps using CLI:

1. Create a role with the trust relationship of:
   ```{
	"Version": "2012-10-17",
	"Statement": [
		{
			"Effect": "Allow",
			"Principal": { "Service": "ec2.amazonaws.com" },
			"Action": "sts:AssumeRole"
		}
	]
   }
   ```
2. Attach the SSM policy `arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore` to this role
3. Create an instance profile
4. Add the role to the instance profile
5. Associate the instance profile to the instance

Execution output:

1. `aws iam create-role --role-name DebugCTFSSMRoleDeleteMe --assume-role-policy-document file://EC2Trust.json`
    Output:
   ```
   {
    "Role": {
        "Path": "/",
        "RoleName": "DebugCTFSSMRoleDeleteMe",
        "RoleId": "AROAYFH3WMUJDZFL6TIXX",
        "Arn": "arn:aws:iam::561021084946:role/DebugCTFSSMRoleDeleteMe",
        "CreateDate": "2023-01-05T21:21:30+00:00",
        "AssumeRolePolicyDocument": {
            "Version": "2012-10-17",
            "Statement": [
                {
                    "Effect": "Allow",
                    "Principal": {
                        "Service": "ec2.amazonaws.com"
                    },
                    "Action": "sts:AssumeRole"
                }
            ]
        }
    }
   }
   ```
2. `aws-vault exec integration-tests -- aws iam attach-role-policy --role-name DebugCTFSSMRoleDeleteMe --policy-arn arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore`
3. `aws-vault exec integration-tests -- aws iam create-instance-profile --instance-profile-name DebugCTFSSMInstanceProfileDeleteMe`
    Output:
    ```
   {
    "InstanceProfile": {
        "Path": "/",
        "InstanceProfileName": "DebugCTFSSMInstanceProfileDeleteMe",
        "InstanceProfileId": "AIPAYFH3WMUJPMYRJXAD6",
        "Arn": "arn:aws:iam::561021084946:instance-profile/DebugCTFSSMInstanceProfileDeleteMe",
        "CreateDate": "2023-01-05T21:24:25+00:00",
        "Roles": []
    }
   }
   ```
4. `aws-vault exec integration-tests -- aws iam add-role-to-instance-profile --instance-profile-name DebugCTFSSMInstanceProfileDeleteMe --role-name DebugCTFSSMRoleDeleteMe`
5. `aws-vault exec integration-tests -- aws ec2 associate-iam-instance-profile --iam-instance-profile Name=DebugCTFSSMInstanceProfileDeleteMe --instance-id i-078072e926510f45a`
   Output:
   ```
   {
    "IamInstanceProfileAssociation": {
        "AssociationId": "iip-assoc-035352d8c9424f4ae",
        "InstanceId": "i-078072e926510f45a",
        "IamInstanceProfile": {
            "Arn": "arn:aws:iam::561021084946:instance-profile/DebugCTFSSMInstanceProfileDeleteMe",
            "Id": "AIPAYFH3WMUJPMYRJXAD6"
        },
        "State": "associating"
    }
   }
   ```
6. `aws ec2 describe-iam-instance-profile-associations --filters Name=instance-id,Values=i-078072e926510f45a`
   Output:
   ```
   {
    "IamInstanceProfileAssociations": [
        {
            "AssociationId": "iip-assoc-035352d8c9424f4ae",
            "InstanceId": "i-078072e926510f45a",
            "IamInstanceProfile": {
                "Arn": "arn:aws:iam::561021084946:instance-profile/DebugCTFSSMInstanceProfileDeleteMe",
                "Id": "AIPAYFH3WMUJPMYRJXAD6"
            },
            "State": "associated"
        }
    ]
   }
   ```
7. `aws ssm start-session --target i-078072e926510f45a`
   Output:
   `An error occurred (TargetNotConnected) when calling the StartSession operation: i-078072e926510f45a is not connected.`
8. I then ran `aws ssm start-session --target i-078072e926510f45a` again, and it worked (gave me a shell in the instance). This was a few minutes later. So we can't rely on the state being correct when the `describe-iam-instance-profile-associations` command says that the instance profile is `associated`, which is a little frustrating. But at least it's working.

nschmeller:
Now I'm going to run the same experiment as above, but use the role created by CTF SSM, with the policy already attached to the role by CTF SSM:

1. `aws iam create-instance-profile --instance-profile-name DebugCTFSSMInstanceProfileDeleteMe2`
   Output:
   ```
   {
    "InstanceProfile": {
        "Path": "/",
        "InstanceProfileName": "DebugCTFSSMInstanceProfileDeleteMe2",
        "InstanceProfileId": "AIPAYFH3WMUJGXGF6Y2MV",
        "Arn": "arn:aws:iam::561021084946:instance-profile/DebugCTFSSMInstanceProfileDeleteMe2",
        "CreateDate": "2023-01-05T21:49:12+00:00",
        "Roles": []
    }
   }
   ```
2. `aws-vault exec integration-tests -- aws iam add-role-to-instance-profile --instance-profile-name DebugCTFSSMInstanceProfileDeleteMe2 --role-name Lacework-Agent-SSM-Install-Role`
3. `aws-vault exec integration-tests -- aws ec2 associate-iam-instance-profile --iam-instance-profile Name=DebugCTFSSMInstanceProfileDeleteMe2 --instance-id i-0c159fa19e6b36094`
   Output:
   ```
   {
    "IamInstanceProfileAssociation": {
        "AssociationId": "iip-assoc-0ce2bf4c72cc0ef3e",
        "InstanceId": "i-0c159fa19e6b36094",
        "IamInstanceProfile": {
            "Arn": "arn:aws:iam::561021084946:instance-profile/DebugCTFSSMInstanceProfileDeleteMe2",
            "Id": "AIPAYFH3WMUJGXGF6Y2MV"
        },
        "State": "associating"
    }
   }
   ```
4. `aws ssm start-session --target i-0c159fa19e6b36094` worked!!

nschmeller:
I ran a new experiment where I tried associating the instance profile created by CTF SSM to the instance:

1. `aws-vault exec integration-tests -- aws ec2 associate-iam-instance-profile --iam-instance-profile Name=Lacework-Agent-SSM-Install-Instance-Profile --instance-id i-0bfc34480a45456ca`
2. `aws-vault exec integration-tests -- aws ssm start-session --target i-0bfc34480a45456ca` did not succeed after 2min, but it did succeed after 4min...

So it looks like we need to try waiting for a while for CTF SSM, like 5min. That seems extravagant, but whatever it takes to work.

nschmeller:
I tested the 5min sleep yesterday, and it worked (!!). It's a frustatingly long amount of time to sleep. Maybe we can do retries...

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,249 @@
...
+		"n",
+		50,
+		"maximum number of workers executing AWS API calls, set if rate limits are lower or higher than normal",
+	)
+	agentInstallAWSSSMCmd.Flags().StringVar(
+		&agentCmdState.InstallBYORole,
+		"iam_role_name",
+		"",
+		"IAM role name (not ARN) with SSM policy, if not provided then an ephemeral role will be created",
+	)
```
quodlibetor, line 93: One other option -- add a flag that just skips installing or verifying and SSM instance profile. If customers are using instance profiles then it's possible that they already have enabled SSM access.

Documenting that they can either have the CLI create the role/profile _or_ provide it themselves seems like it would bump customer use cases to maybe 90% or 95%. The big thing left out would be customers who require that SSM only be provided with specific Command documents.

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,249 @@
...
+			// Check if agent is already installed on the host, skip if yes
+			// Sleep for up to 5min to wait for instance profile to associate with instance
+			var ssmError error
+			var commandOutput ssm.GetCommandInvocationOutput
+			for i := 0; i < 5; i++ {
+				cli.Log.Debugw("waiting for instance profile to associate with instance, sleeping 1min",
+					"iteration number (time slept in minutes)", i,
+					"instance ID", threadRunner.InstanceID,
+				)
+				time.Sleep(1 * time.Minute)
```
quodlibetor, line 179: it seems like this should be at the end of the loop to only execute on failure, (and I guess check for `i < 6` to keep it for 5 minutes)

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,249 @@
...
+			defer wg.Done()
+
+			threadRunner := *runner
+			runnerCopyWg.Done()
+
+			cli.Log.Debugw("runner info: ",
+				"user", threadRunner.Runner.User,
+				"region", threadRunner.Region,
+				"az", threadRunner.AvailabilityZone,
+				"instance_id", threadRunner.InstanceID,
```
quodlibetor, line 150: nit: I'd be inclined to make the `instance_id` log output label consistent throughout this function (it's named variously `instance_id`, `instance ID`, and `runner`). IMO one of `instance_id` or `instanceID` are the best names for it, but as long as it's consistent it's fine.

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,249 @@
...
+			)
+
+			// Attach an instance profile with our new role to the runner
+			err = threadRunner.AssociateInstanceProfileWithRunner(cfg, instanceProfile)
+			if err != nil {
+				cli.Log.Debugw("failed to attach instance profile to runner",
+					"error", err,
+					"instance ID", threadRunner.InstanceID,
+					"role", role,
+					"instance profile", instanceProfile,
```
quodlibetor, line 161: light nit, TIOLI: logs are usually easier to parse if labels don't have spaces in them.

cli/cmd/awsiam.go
```diff
@@ -0,0 +1,304 @@
...
+	]
+}`
+	input := &iam.CreateRoleInput{
+		AssumeRolePolicyDocument: &trustPolicyDocument,
+		RoleName:                 aws.String(roleName),
+		Description: aws.String(
+			`Ephemeral role to install Lacework agents using SSM; created by the Lacework CLI.
+Safe to delete if found`,
+		),
+	}
```
quodlibetor, line 195: It's polite to add tags that clearly identify this role as lacework-derived ephemeral for automation. There are tools that folks use that specifically look at tags, so this can surface things for folks that just having a good name doesn't (untested code):

```suggestion
	input := &iam.CreateRoleInput{
		AssumeRolePolicyDocument: &trustPolicyDocument,
		RoleName:                 aws.String(roleName),
		Description: aws.String(
			`Ephemeral role to install Lacework agents using SSM; created by the Lacework CLI.
Safe to delete if found`,
		),
		Tags: []*iam.Tag{
		{
		    Key:   aws.String("LaceworkAutomation"),
		    Value: aws.String("agent-ssm-install")
		}
		}
	}
```

nschmeller: Good idea, done with bffd96d7613cbc53e83f996f0b922ae2b2f6c74b

cli/cmd/agent_aws-install_ec2ssm.go
nschmeller: I agree, done with cd07b01bdd62bb29ec37766b74d0b71f25e7df1c

cli/cmd/agent_aws-install_ec2ssm.go
nschmeller: This is a good point. Here are the tradeoffs as I see them:

Pros: 
* Only sleep on failure (if customer has instance profile already associated, this saves them a minute of execution)

Cons:
* The last minute of sleep isn't useful, since we've already tried the command and failed, so we're waiting a minute just to return an error. 

I think the solution that has the best of both worlds is to wrap the sleep with a conditional that executes the sleep if it's not the last iteration in the loop. Which seems a little messy to me, but sleeps aren't exactly pristine code in the first place. 

I'm open to other suggestions too!

nschmeller: Done with 84b48008fe1f12807fad27b1546d1499e38dafb5

cli/cmd/agent_aws-install_ec2ssm.go
nschmeller: Yeah, that makes sense, basically saying that if the customer has everything setup for SSM already, then give them the option. It's basically what the logic is doing by checking if the existing infra works.

nschmeller: 927f83530ea0052ed3aa8519fb41afb909494444

nschmeller:
The CLI integration tests are failing for an unrelated reason (https://g.codefresh.io/build/63be2cfa65ff901d108d6511?step=Integration&tab=output&logs=terminal). I asked about this in Slack: https://lacework.slack.com/archives/C01RSNZ0C5V/p1673409107239759. Hopefully it's resolved quickly!

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,267 @@
...
+	}
+
+	var role types.Role
+	var instanceProfile types.InstanceProfile
+	if !agentCmdState.InstallSkipCreatInfra {
+		var err error
+		role, instanceProfile, err = SetupSSMAccess(cfg, agentCmdState.InstallBYORole, token)
+		defer func() {
+			err := TeardownSSMAccess(cfg, role, instanceProfile, agentCmdState.InstallBYORole) // clean up after ourselves
+			cli.Log.Warnw("got an error while tearing down IAM infra", "error", err)
```
cirego, line 136: We should be more precise. `IAM infra` doesn't mean anything by itself.

```suggestion
			cli.Log.Warnw("got an error while tearing down IAM infra", "error", err)
```

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,267 @@
...
+		&agentCmdState.InstallBYORole,
+		"iam_role_name",
+		"",
+		"IAM role name (not ARN) with SSM policy, if not provided then an ephemeral role will be created",
+	)
+	agentInstallAWSSSMCmd.Flags().BoolVar(
+		&agentCmdState.InstallSkipCreatInfra,
+		"skip_infra_creation",
+		false,
+		"set this flag to skip IAM / SSM infra creation. Assumes all instances have SSM enabled",
```
cirego, line 99: Nit: `infra_creation` doesn't describe what this command is actually doing. I think that "attach profile" more accurately describes the action that this flag prevents.

```suggestion
	agentInstallAWSSSMCmd.Flags().BoolVar(
		&agentCmdState.InstallNoAttachProfile,
		"no_attach_profile",
		false,
		"set this flag to skip attaching an SSM instance profile to EC2 instances. All instances should have SSM already enabled",
```

cli/cmd/awsiam.go
```diff
@@ -0,0 +1,324 @@
...
+	output, err := c.GetRole(context.Background(), input)
+	if err != nil {
+		return types.Role{}, err
+	}
+
+	return *output.Role, nil
+}
+
+// createSSMRole makes a call to the AWS API to create an IAM role.
+// This role allows the current user to assume it.
```
cirego, line 161: I might be missing something but I don't see how this role is scoped such that *only* the current user can assume this role. We need to make sure that only the current user can assume the role...

vendor/github.com/aws/aws-sdk-go-v2/service/iam/api_client.go
```diff
@@ -0,0 +1,434 @@
+// Code generated by smithy-go-codegen DO NOT EDIT.
```
CatZimmermann-Lacework, line 1: Do the `smith-go-codegen` generated files have to be committed or can they be generated when building the SDK?

nschmeller: I believe they have to be committed as the SDK vendors all dependencies

cli/cmd/agent_aws-install_ec2ssm.go
nschmeller: I think it technically does both, since if the flag is set, the role and instance profile will not be created. I'll reword this

nschmeller: Done with f0167d2234c0ad1ab5be6ecb4350d6f10cc7216b

cli/cmd/awsiam.go
nschmeller: Good catch, this comment was when I (mistakenly) thought that the role was meant for the current user to assume. Actually, it's for the EC2 service to assume. I updated this with 16643d2428e68ecffa310d251596b1bb5990066b

cli/cmd/agent_aws-install_ec2ssm.go
nschmeller: Actually, I think this is accurate (although I suppose we could change from `IAM infra` to `IAM role / instance profile`. This function does tear down the role and instance profile, if they were created. It does not detach the temporary instance profile. I think this is done automatically when the instance profile is deleted (https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_manage_delete.html).

cli/cmd/agent_aws-install_ec2ssm.go
quodlibetor: Yeah I think the largest issue is actually that if you have 1000 hosts and a workgroup of size 10 then you will sleep for about 100 minutes, even though you don't need to sleep for more than 1 minute -- the sleep should only need to happen the first time through.

Your fix looks good to me.

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,347 @@
...
+instances already have associated instance profiles, this command will not change
+their state. This command will teardown the IAM role and instance profile before exiting.
+
+This command authenticates with AWS credentials from well-known locations on the user's
+machine. The principal associated with these credentials should have the
+'AmazonEC2FullAccess', 'IAMFullAccess' and 'AmazonSSMFullAccess' policies attached.
+
+To skip IAM role / instance profile creation and instance profile association:
+
+    lacework agent aws-install ec2ssm --skip_infra_creation
```
cirego, line 55: Given that the other flag is called `--iam_role_name`, perhaps call this:

```suggestion
    lacework agent aws-install ec2ssm --skip_iam_role_creation
```

I still think `infra` is too generic and does not inform the customer about what creation is being skipped.

cli/cmd/agent_aws-install_ec2ssm.go
cirego: Sure, calling this `IAM role / instance profile` would be a good change.

cli/cmd/agent_aws-install_ec2ssm.go
nschmeller: That makes sense, fixed this in 1659b5a60b61f3d313a22c7ce13c8777548521bc

cirego:
It would be great if we can get a review from the security team on the IAM stuff.

nschmeller:
> It would be great if we can get a review from the security team on the IAM stuff.

I agree, but I'm curious what specifically you think they'd care about? Is the thought that we want to make sure we're creating least-privilege role + instance profile?

cli/cmd/awsiam.go
```diff
@@ -0,0 +1,323 @@
...
+			)
+		}
+	}
+
+	cli.Log.Debugw("adding role to instance profile", "role", role, "instance profile", instanceProfile)
+	addInput := &iam.AddRoleToInstanceProfileInput{
+		InstanceProfileName: instanceProfile.InstanceProfileName,
+		RoleName:            role.RoleName,
+	}
+	_, err := c.AddRoleToInstanceProfile(context.Background(), addInput)
```
pmoody-, line 320: maybe not for the POC, but internally i'd been meaning to look at any existing applied instance profile and see if i can just hork the ssm bits into it.

integration/test_resources/help/agent_aws-install_ec2ssm
```diff
@@ -0,0 +1,70 @@
+This command installs the agent on all EC2 instances in an AWS account using SSM.
```
pmoody-, line 1: is there anything in here about reporting on instances that don't support ssm, or that don't support ssm by default? IIRC, suse and a few other distros were 'download the ssm agent with this curl command and manually install'

cli/cmd/awsiam.go
nschmeller: I was thinking about that too. It's not too difficult:
* Use the EC2 API to get the already associated instance profile
* Use the IAM API to get more details about the instance profile (such as the added role name)
* Use the IAM API to get more details about the added role (such as which policies are attached)
* If the role doesn't have the proper SSM policies attached (and doesn't trust the EC2 service to assume) then fix that

I didn't do this because it's trickier when the requirement is "ephemeral infra" for external customers, because then we have to tear down, which means we have to remember what the state looked like before we messed around with it. But if customers want that, then we could do that.

integration/test_resources/help/agent_aws-install_ec2ssm
nschmeller: If the instance doesn't have the SSM agent running, then we'll skip it and print a warning to the customer. The idea is "best effort". It's a good idea to add a caveat for that though, done in 6ae711ea10f99a52a33f5bc82bf4619e6b81a822

That said, I'm curious how you downloaded the SSM agent without SSM? Did you do this for the lambda?

pmoody-: Internally we just throw up our hands and say darn. I think the eventual plan, long since discarded by all of the other work that's come up, was to try and force people to use ami's that already either had the agent installed by default or already had ssm support (either by installing it via packer or like, requiring people use a supported distro).

nschmeller: Personally, I wouldn't mind being forced to use a distro that has SSM support, but I can see how the project priority drops once you have 80% coverage 😁

cirego:
> > It would be great if we can get a review from the security team on the IAM stuff.
> 
> I agree, but I'm curious what specifically you think they'd care about? Is the thought that we want to make sure we're creating least-privilege role + instance profile?

Yes, I wanted a sanity check that we're creating a role with least-privileges required and that we're not going to open up a security hole in customer environments.

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,350 @@
...
+	)
+	agentInstallAWSSSMCmd.Flags().StringVar(
+		&agentCmdState.InstallBYORole,
+		"iam_role_name",
+		"",
+		"IAM role name (not ARN) with SSM policy, if not provided then an ephemeral role will be created",
+	)
+	agentInstallAWSSSMCmd.Flags().BoolVar(
+		&agentCmdState.InstallSkipCreatInfra,
+		"skip_iam_role_creation",
```
quodlibetor, line 119: also very nitpicky, TIOLI: it's more common for unix commands to use kebab-case instead of snake_case for flags. Bash/zsh/etc don't distinguish between underscores and dashes, they're both just word chars.

cli/cmd/aws.go
```diff
@@ -49,6 +51,15 @@ func awsDescribeInstances() ([]*lwrunner.AWSRunner, error) {
 	return allRunners, nil
 }
 
+func GetConfig() (aws.Config, error) {
+	cfg, err := config.LoadDefaultConfig(context.Background())
+	if err != nil {
+		return cfg, err
+	}
+
+	return cfg, nil
```
quodlibetor, line 60: Is this different to just `return config.LoadDefaultConfig(context.Background())`?

cli/cmd/awsiam.go
```diff
@@ -0,0 +1,325 @@
...
+// - Deletes the instance profile
+// - Detaches all managed policies from the role (assumes no inline policies attached)
+// - Deletes the role
+func teardownSSMAccess(cfg aws.Config, role types.Role, instanceProfile types.InstanceProfile, byoRoleName string) error {
+	c := iam.New(iam.Options{
+		Credentials: cfg.Credentials,
+		Region:      cfg.Region,
+	})
+
+	cli.Log.Debug("only destroying instance profile if it's ours")
```
quodlibetor, line 69: This log doesn't seem useful now that the code is written (just log "destroying" or "not destroying")

cli/cmd/awsiam.go
```diff
@@ -0,0 +1,325 @@
...
+	if *instanceProfile.InstanceProfileName == instanceProfileName {
```
quodlibetor, line 70: I think that this will be safer if you examine a tag created by this script rather than the role name, I could imagine refactorings very easily forgetting that we care specifically about the role name and accidentally deleting a customer's instance profile.

cli/cmd/awsiam.go
```diff
@@ -0,0 +1,325 @@
...
+		}
+		_, err = c.DeleteInstanceProfile(context.Background(), deleteProfileInput)
+		if err != nil {
+			return err
+		}
+	}
+
+	if byoRoleName != "" || *role.RoleName != roleName {
+		return nil // we didn't create this role, we should not delete it
+	}
```
quodlibetor, line 93: Same: examining an explicit tag rather than an implicit property of the script seems safer.

lwrunner/awsrunner.go
```diff
@@ -116,6 +122,180 @@ func (run AWSRunner) SendPublicKey(pubBytes []byte) error {
...
+
+	// The runner has an instance profile attached, the instance profile has a role,
+	// and the role does not have the policy we need for SSM. We can't install on
+	// this instance, return an error
+	return false, fmt.Errorf(
+		"runner %v already has an instance profile (%v) attached, does not have policy %s",
+		run,
+		getInstanceProfileOutput.InstanceProfile,
+		SSMInstancePolicy,
+	)
```
quodlibetor, line 232: fwiw in future work we could still attach policies to existing roles: https://docs.aws.amazon.com/sdk-for-go/api/service/iam/#AttachRolePolicyInput . I _do_ think it should be future work, checking all that and cleaning it up is a bunch of busywork, and what we already have here is great and useful.

quodlibetor (commented):
I think the only meaningful comment that I have here is that we should probably use an explicit tag rather than properties of the Role or InstanceProfile name to determine if we should destroy it.

Looks really good otherwise!

lwrunner/awsrunner.go
nschmeller: Totally agree, I filed a ticket for this, https://lacework.atlassian.net/browse/RAIN-47063

cli/cmd/aws.go
nschmeller: I agree that that's simpler, we can just do that. But, I actually want to change this to not have a `GetConfig` function, and just pass the `cfg` struct down from the toplevel caller. I'll probably do this in a separate PR.

cli/cmd/agent_aws-install_ec2ssm.go
nschmeller: I agree, this is just following the existing flag style for the Lacework CLI

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,350 @@
...
+	"github.com/lacework/go-sdk/lwrunner"
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+var (
+	agentInstallAWSSSMCmd = &cobra.Command{
+		Use:   "ec2ssm",
+		Args:  cobra.NoArgs,
+		Short: "Use SSM to securely install on EC2 instances",
```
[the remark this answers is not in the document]
nschmeller: Good idea, done with 3d3dc43587a40eedaa939f530798d06419623fc5

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,350 @@
...
+		RunE:  installAWSSSM,
+		Long: `This command installs the agent on all EC2 instances in an AWS account using SSM.
```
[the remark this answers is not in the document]
nschmeller: Done with 09737e9752df858cf278cf0015118ca62a5c9fdb

cli/cmd/awsiam.go
nschmeller: I agree, plus tags seem a bit more kosher. Done with 162cf4a1ca3a9722a25341408478d7f6edaf476c

cli/cmd/awsiam.go
nschmeller: Done with 162cf4a1ca3a9722a25341408478d7f6edaf476c

cli/cmd/awsiam.go
nschmeller: Yeah, the code speaks for itself. I had the idea that I would turn comments into debug logs, in the spirit of "self documenting code" (or maybe "self documenting execution"?). Removed with fda4b8e9ba880406023d61d0d9aa3b0d10218522

quodlibetor (approved):
looks great!

nschmeller:
One last change to make: the run at https://github.com/lacework/agent/actions/runs/3942421877/jobs/6746065889#step:6:8207 is failing because we're not describing EC2 instances that we can install on correctly for SSM. 

Specifically, https://github.com/lacework/go-sdk/blob/64142fa67c8e5853da30e19ce654e07bb2580102/lwrunner/awsrunner.go#L152 will check that we have a specific SSH username for the instance so we can use SSH-based install methods (i.e. `ec2ic` and `ec2ssh`). But since SSM doesn't use SSH, we don't need to provide an SSH username. 

I'll do this tomorrow!

cli/cmd/agent_aws-install_ec2ic_test.go
```diff
@@ -30,6 +31,7 @@ func TestAwsInstallEC2IC(t *testing.T) {
 		t.Skip("aws credentials not found in environment, skipping test")
 	}
 
+	cli.Log = lwlogger.New("DEBUG").Sugar()
```
afiune, line 34: Is this intentional? Why do we hard code this?

cli/cmd/agent_aws-install_ec2ssm.go
```diff
@@ -0,0 +1,350 @@
...
+			// user didn't provide an agent token
+			cli.Log.Debugw("agent token not provided, asking user to select one now")
+			var err error
+			token, err = selectAgentAccessToken()
+			if err != nil {
+				return err
+			}
+		} else {
+			return errors.New("user did not provide or interactively select an agent token")
+		}
```
afiune, line 145: nit; reduce if statement logic and better error message to the end-user
```suggestion
		if !cli.InteractiveMode() {
			return errors.New("agent token not provided. Use '--token' when running in non interactive mode")
		}
		// user didn't provide an agent token
		cli.Log.Debug("agent token not provided, asking user to select one now")
		var err error
		token, err = selectAgentAccessToken()
		if err != nil {
			return err
		}
```

cli/cmd/aws.go
afiune: I am with @quodlibetor here, if we created a ticket to refactor it later, let's put this back. (My $0.02)

cli/cmd/agent_aws-install_ec2ic_test.go
nschmeller: Done with 61813be7983295cc225feb943a0722b6f5276763

afiune (commented):
@cirego @nschmeller at what point are we going to start writing Unit Tests for this code?

I know we talked that integration tests would be done over in the agent repo but, unit tests should still exist in this repo, specially since some of the functions have increased in complexity and small modifications could impact other features/functionalities. Also, it is easier to review.

cli/cmd/aws.go
nschmeller: Looking at it again, I agree with this. Removed with 16aa209ac553b404c32d9f3e97eeef2c2c560630

cli/cmd/agent_aws-install_ec2ssm.go
nschmeller: Good idea, done in 17a10310cea74528271622b52e1303a6d298a00e

cirego:
> @cirego @nschmeller at what point are we going to start writing Unit Tests for this code?
> 
> I know we talked that integration tests would be done over in the agent repo but, unit tests should still exist in this repo, specially since some of the functions have increased in complexity and small modifications could impact other features/functionalities. Also, it is easier to review.

Hi @afiune, can you help define what kind of unit tests you're looking for? Many of these functions are making direct API calls to AWS services, which makes them complicated to unittest. We could look a building a mock AWS SDK for testing and I think we should talk about the relative priority of that work. Each of these PRs have been accompanied by an integration test in the Agent repo, which we should look at moving into this repo soon. I will talk with @nschmeller about using a mock service.

PR #1053, opened 2022-12-07
docs(cli): list supported hosts for `ec2ic`

nschmeller:
EC2InstanceConnect supports only some hosts for agent installs using
`aws-install`. This commit adds information about these hosts to the
`aws-install ec2ic` long help text.

Fixes RAIN-44562

Integration tests for the new help text

## Issue

https://lacework.atlassian.net/browse/RAIN-44562

### DIFF
--- a/cli/cmd/agent_aws-install_ec2ic.go
+++ b/cli/cmd/agent_aws-install_ec2ic.go
@@ -63,6 +63,11 @@ AWS credentials are read from the following environment variables:
 - AWS_SESSION_TOKEN (optional)
 - AWS_REGION (optional)
 
+This command will only install the agent on hosts that are supported by
+EC2InstanceConnect. The supported AMI types are Amazon Linux 2 and Ubuntu
+16.04 and later. There may also be a region restriction. You must be able
+to access the host's network.
+
 This command will automatically add hosts with successful connections to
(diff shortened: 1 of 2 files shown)


cli/cmd/agent_aws-install_ec2ic.go
```diff
@@ -63,6 +63,11 @@ AWS credentials are read from the following environment variables:
 - AWS_SESSION_TOKEN (optional)
 - AWS_REGION (optional)
 
+This command will only install the agent on hosts that are supported by
+EC2InstanceConnect. The supported AMI types are Amazon Linux 2 and Ubuntu
+16.04 and later. There may also be a region restriction. You must be able
+to access the host's network.
```
cirego, line 69: I think we can drop the `You must be able to access the host's network.` bit. I believe the correct wording is something like `The instance must be listening on an IP Address routable from the host running this command.` I think it may be better to drop this verbiage.

cirego (approved):

cli/cmd/agent_aws-install_ec2ic.go
nschmeller: Makes sense to me, removed it

cirego (approved):

PR #1057, opened 2022-12-07
fix: revert compliance csv report_type field format

dmurray-lacework:

## Summary

Api v2 returns the report type in a different format. Revert the format change in cvs output 

The v1 api csv reports had the report_type field format as `AWS_CIS_S3`
The v2 api report_type field format was`CIS Amazon Web Services Foundations Benchmark v1.4.0`

This change reverts back to `AWS_CIS_S3`

Run the compliance get-report cmd with csv flag

## Issue

https://lacework.atlassian.net/browse/ALLY-1306


afiune (approved):
Needed some tests...

PR #1063, opened 2022-12-09
build(cli): Support aarch64 architecture

kolbeinn:
## Summary

On some arm64 platforms, the `uname -m` command will return 'aarch64' and not 'arm64'. This results in the incorrect lacework cli executable to be installed, namely a 386 version of it. This commit ensures that aarch64 is treated identically to arm64.

Test plan:
1. On an M1 mac, I open a Docker container:
```
docker run -it --rm --entrypoint /bin/bash -v `pwd`:/go-sdk golang:1.18
```
2. Verify that `uname -m` outputs aarch64.
3. I install the cli:
```
cd /go-sdk
make prepare
make install-cli
```
4. Verify that the installed version works:
```
lacework version
```

The fact that the Lacework version worked means that we're no longer invoking a 386 binary (which of course immediately crashed)/


hazedav (approved):
@kolbeinn with that knockoff macbook...

PR #1065, opened 2022-12-13
chore: Fix struct for create and update policy

varming:
## Description
Make alertProfile optional and add three fields:
1. additionalInformation
2. infoLink
3. references

## Links
https://lacework.atlassian.net/browse/RAIN-45238

### DIFF
--- a/api/policy.go
+++ b/api/policy.go
@@ -46,19 +46,22 @@ func NewV2PolicyService(c *Client) *PolicyService {
 var ValidPolicySeverities = []string{"critical", "high", "medium", "low", "info"}
 
 type NewPolicy struct {
-	PolicyID      string   `json:"policyId,omitempty" yaml:"policyId,omitempty" `
-	PolicyType    string   `json:"policyType" yaml:"policyType"`
-	QueryID       string   `json:"queryId" yaml:"queryId"`
-	Title         string   `json:"title" yaml:"title"`
-	Enabled       bool     `json:"enabled" yaml:"enabled"`
-	Description   string   `json:"description" yaml:"description"`
-	Remediation   string   `json:"remediation" yaml:"remediation"`
-	Severity      string   `json:"severity" yaml:"severity"`
-	Limit         int      `json:"limit,omitempty" yaml:"limit,omitempty"`
-	EvalFrequency string   `json:"evalFrequency,omitempty" yaml:"evalFrequency,omitempty"`
-	AlertEnabled  bool     `json:"alertEnabled" yaml:"alertEnabled"`
-	AlertProfile  string   `json:"alertProfile" yaml:"alertProfile"`
-	Tags          []string `json:"tags,omitempty" yaml:"tags,omitempty"`
+	PolicyID              string   `json:"policyId,omitempty" yaml:"policyId,omitempty" `
+	PolicyType            string   `json:"policyType,omitempty" yaml:"policyType,omitempty"`
+	QueryID               string   `json:"queryId" yaml:"queryId"`
+	Title                 string   `json:"title" yaml:"title"`
+	Enabled               bool     `json:"enabled" yaml:"enabled"`
+	Description           string   `json:"description" yaml:"description"`
+	Remediation           string   `json:"remediation" yaml:"remediation"`
+	Severity              string   `json:"severity" yaml:"severity"`
+	Limit                 int      `json:"limit,omitempty" yaml:"limit,omitempty"`
+	EvalFrequency         string   `json:"evalFrequency,omitempty" yaml:"evalFrequency,omitempty"`
+	AlertEnabled          bool     `json:"alertEnabled" yaml:"alertEnabled"`
+	AlertProfile          string   `json:"alertProfile,omitempty" yaml:"alertProfile,omitempty"`
+	Tags                  []string `json:"tags,omitempty" yaml:"tags,omitempty"`
+	References            []string `json:"references,omitempty" yaml:"references,omitempty"`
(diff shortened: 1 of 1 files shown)


api/policy.go
```diff
@@ -97,19 +100,22 @@ This would prevent someone from toggling something to disabled or 0 respectively
 As such we are using pointers instead of primitives for booleans and integers in this struct
 */
 type UpdatePolicy struct {
```
hazedav, line 102: We should validate what types are optional here as some of them may need to be pointer types.  For instance, if someone has set `AlertProfile`, how do they properly update a policy to "empty out" the `AlertProfile` (since `omitempty` will simply omit `AlertProfile`)...?

varming: Interesting. We should not infer empty strings and we should not omit empty strings in any of the policy apis (IMHO). So the fields in the struct probably needs to look like:
```golang
PolicyID *string `json:"policyId,omitempty"`
```
to be able to parse the user provided text and faithfully pass it on to the api (the existing `string` is broken as it will remove the `policyId` key from the json object if `PolicyID` is the empty string, and thus the customer may get an odd error message that is pertains to input that is different from what he wrote). In other words I would expect the struct to be:
```golang
type NewPolicy struct {
        PolicyID              *string   `json:"policyId,omitempty" yaml:"policyId,omitempty" `
        PolicyType            *string   `json:"policyType,omitempty" yaml:"policyType,omitempty"`
        QueryID               *string   `json:"queryId,omitempty" yaml:"queryId,omitempty"`
        Title                 *string   `json:"title,omitempty" yaml:"title,omitempty"`
        Enabled               *bool     `json:"enabled,omitempty" yaml:"enabled,omitempty"`
        Description           *string   `json:"description,omitempty" yaml:"description,omitempty"`
        Remediation           *string   `json:"remediation,omitempty" yaml:"remediation,omitempty"`
        Severity              *string   `json:"severity,omitempty" yaml:"severity,omitempty"`
        Limit                 *int      `json:"limit,omitempty" yaml:"limit,omitempty"`
        EvalFrequency         *string   `json:"evalFrequency,omitempty" yaml:"evalFrequency,omitempty"`
        AlertEnabled          *bool     `json:"alertEnabled,omitempty" yaml:"alertEnabled,omitempty"`
        AlertProfile          *string   `json:"alertProfile,omitempty" yaml:"alertProfile,omitempty"`
        Tags                  []string `json:"tags,omitempty" yaml:"tags,omitempty"`
        References            []string `json:"references,omitempty" yaml:"references,omitempty"`
        AdditionalInformation *string   `json:"additionalInformation,omitempty" yaml:"additionalInformation,omitempty"`
        InfoLink              *string   `json:"infoLink,omitempty" yaml:"infoLink,omitempty"`
}
```
to not mess with the input in a material way.

Such changes are outside the scope of what I want to fix here, so I'll just make sure to not expand the problem with the type of the new fields.

api/policy.go
```diff
@@ -46,19 +46,22 @@ func NewV2PolicyService(c *Client) *PolicyService {
...
+	Description           string   `json:"description" yaml:"description"`
+	Remediation           string   `json:"remediation" yaml:"remediation"`
+	Severity              string   `json:"severity" yaml:"severity"`
+	Limit                 int      `json:"limit,omitempty" yaml:"limit,omitempty"`
+	EvalFrequency         string   `json:"evalFrequency,omitempty" yaml:"evalFrequency,omitempty"`
+	AlertEnabled          bool     `json:"alertEnabled" yaml:"alertEnabled"`
+	AlertProfile          string   `json:"alertProfile,omitempty" yaml:"alertProfile,omitempty"`
+	Tags                  []string `json:"tags,omitempty" yaml:"tags,omitempty"`
+	References            []string `json:"references,omitempty" yaml:"references,omitempty"`
+	AdditionalInformation *string  `json:"additionalInformation,omitempty" yaml:"additionalInformation,omitempty"`
```
hazedav, line 63: We don't need pointers here...

api/policy.go
```diff
@@ -143,22 +149,25 @@ func ParseUpdatePolicy(s string) (UpdatePolicy, error) {
...
+	Description           string   `json:"description" yaml:"description"`
+	Remediation           string   `json:"remediation" yaml:"remediation"`
+	Severity              string   `json:"severity" yaml:"severity"`
+	Limit                 int      `json:"limit" yaml:"limit"`
+	EvalFrequency         string   `json:"evalFrequency" yaml:"evalFrequency"`
+	AlertEnabled          bool     `json:"alertEnabled" yaml:"alertEnabled"`
+	AlertProfile          string   `json:"alertProfile" yaml:"alertProfile"`
+	Tags                  []string `json:"tags" yaml:"tags"`
+	References            []string `json:"references" yaml:"references"`
+	AdditionalInformation *string  `json:"additionalInformation" yaml:"additionalInformation"`
```
hazedav, line 166: We don't need pointers here...

api/policy.go
hazedav: The use of pointers is strictly related to the patching (`UpdatePolicy` struct) of optional values such that we can differentiate between zero vs. unspecified...

afiune:
@varming are we good closing this PR since we merged/released https://github.com/lacework/go-sdk/pull/1078?

varming:
> are we good closing this PR since we merged/released #1078?

Sure. It would be good to add the fields that are missing to the struct capturing the details of a policy

hazedav:
@varming , I filed https://lacework.atlassian.net/browse/GROW-1317 to track the addition of the new fields.  Do we know when these go live?  I do not see them here:
https://youraccount.lacework.net/api/v2/docs#tag/Policies/paths/~1api~1v2~1Policies/post

PR #1066, opened 2022-12-13
feat: Add support from hidden suppressions api

rmoles:

## Summary

However it was decided that we should add this to the CLI as a more "official" approach.

This PR adds support for the hidden suppressions API

Tested locally ✅ 

## Issue
https://lacework.atlassian.net/browse/RAIN-44577


api/v2.go
```diff
@@ -88,6 +89,11 @@ func NewV2Endpoints(c *Client) *V2Endpoints {
 		&VulnerabilityExceptionsService{c},
 		NewV2VulnerabilitiesService(c),
 		&AlertsService{c},
+		&SuppressionsServiceV2{c,
```
jon-stewart, line 92: Missing closing bracket here?

jon-stewart (approved):

api/v2.go
rmoles: the closing bracket is on line 96

api/v2_suppressions_test.go
```diff
@@ -0,0 +1,839 @@
...
+	expectedRecsWithSup := []string{"AWS_CIS_1_14", "AWS_CIS_2_7", "AWS_CIS_4_4",
+		"LW_AWS_NETWORKING_50",
+		"LW_S3_1"}
+	sort.Strings(expectedRecsWithSup)
+	sort.Strings(recsWithSuppressions)
+	assert.Equal(t, expectedRecsWithSup, recsWithSuppressions)
+}
+
+func generateSuppressions() string {
+	return ListSuppressions()
```
iok0, line 73: nit: Could collapse to one method here? I was looking for where ListSuppressions() was used - then I found this wrapper

rmoles: Fair point. I'll get that updated

PR #1068, opened 2022-12-14
feat: Display supported policy exception constraints

rmoles:

## Summary

When running a lacework policy show <policId> we should parse & display the supported policy exception constraints, as we now have this info available in the response

Tested locally

## Issue

https://lacework.atlassian.net/browse/GROW-1314

### DIFF
--- a/api/policy.go
+++ b/api/policy.go
@@ -143,22 +143,29 @@ func ParseUpdatePolicy(s string) (UpdatePolicy, error) {
 }
 
 type Policy struct {
-	PolicyID       string   `json:"policyId" yaml:"policyId"`
-	PolicyType     string   `json:"policyType" yaml:"-"`
-	QueryID        string   `json:"queryId" yaml:"queryId"`
-	Title          string   `json:"title" yaml:"title"`
-	Enabled        bool     `json:"enabled" yaml:"enabled"`
-	Description    string   `json:"description" yaml:"description"`
-	Remediation    string   `json:"remediation" yaml:"remediation"`
-	Severity       string   `json:"severity" yaml:"severity"`
-	Limit          int      `json:"limit" yaml:"limit"`
-	EvalFrequency  string   `json:"evalFrequency" yaml:"evalFrequency"`
-	AlertEnabled   bool     `json:"alertEnabled" yaml:"alertEnabled"`
-	AlertProfile   string   `json:"alertProfile" yaml:"alertProfile"`
-	Tags           []string `json:"tags" yaml:"tags"`
-	Owner          string   `json:"owner" yaml:"-"`
-	LastUpdateTime string   `json:"lastUpdateTime" yaml:"-"`
-	LastUpdateUser string   `json:"lastUpdateUser" yaml:"-"`
+	PolicyID               string                                               `json:"policyId" yaml:"policyId"`
+	PolicyType             string                                               `json:"policyType" yaml:"-"`
+	QueryID                string                                               `json:"queryId" yaml:"queryId"`
+	Title                  string                                               `json:"title" yaml:"title"`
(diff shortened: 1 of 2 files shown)


api/policy.go
```diff
@@ -143,22 +143,29 @@ func ParseUpdatePolicy(s string) (UpdatePolicy, error) {
...
+	Severity               string                                               `json:"severity" yaml:"severity"`
+	Limit                  int                                                  `json:"limit" yaml:"limit"`
+	EvalFrequency          string                                               `json:"evalFrequency" yaml:"evalFrequency"`
+	AlertEnabled           bool                                                 `json:"alertEnabled" yaml:"alertEnabled"`
+	AlertProfile           string                                               `json:"alertProfile" yaml:"alertProfile"`
+	Tags                   []string                                             `json:"tags" yaml:"tags"`
+	Owner                  string                                               `json:"owner" yaml:"-"`
+	LastUpdateTime         string                                               `json:"lastUpdateTime" yaml:"-"`
+	LastUpdateUser         string                                               `json:"lastUpdateUser" yaml:"-"`
+	ExceptionConfiguration map[string][]PolicyExceptionConfigurationConstraints `json:"exceptionConfiguration" yaml:"-"`
```
rmoles, line 162: I don't like this, but I couldn't find a better way to parse the Policy Exception constraints, as the format is: an Array of Maps in a Map. It's not clear to me why the outer map was required 🤷 
```
    "exceptionConfiguration": {
      "constraintFields": [
        {
          "dataType": "String",
          "fieldKey": "accountIds",
          "multiValue": true
        },
        {
          "dataType": "String",
          "fieldKey": "resourceNames",
          "multiValue": false
        },
        {
          "dataType": "KVTagPair",
          "fieldKey": "resourceTags",
          "multiValue": true
        }
      ]
    },
```

jon-stewart (approved):

cli/cmd/policy.go
```diff
@@ -590,6 +591,13 @@ func buildPolicyDetailsTable(policy api.Policy) string {
 		{"UPDATED BY", policy.LastUpdateUser},
 		{"EVALUATION FREQUENCY", policy.EvalFrequency},
 	}
+	// append VALID EXCEPTION CONSTRAINTS to the table when the ExceptionConfiguration is not empty
+	exceptionConstraints := strings.Join(
+		getPolicyExceptionConstraintsSlice(policy.ExceptionConfiguration), ", ")
+	if exceptionConstraints != "" {
+		entry := []string{"VALID EXCEPTION CONSTRAINTS", exceptionConstraints}
+		details = append(details, entry)
+	}
```
rmoles, line 600: @afiune any thoughts on whether we want to just drop this field from the table, or whether we should put a message  to the effect of `no constraints found`

api/policy.go
iok0: Yeah, doesn't look necessary right now. Maybe they've plans to add another field to exceptionConfiguration.
Can Go be flexible like that - or does the structure/type need to be defined up front?

iok0 (approved):

api/policy.go
rmoles: Perhaps. Still annoying for parsing!
For parsing the values into structs we need to declare the structure up front. I was hoping I could tell GoLang to chain the map keys when parsing ie. `json:"exceptionConfiguration.constraintFields" yaml:"-"` but that doesn't seem to be possible.

cli/cmd/policy.go
afiune: Let's add it always and if the policy do not have constraints then show "None" - because no constraint found implies that maybe, there could be constraints.

afiune (approved):

cli/cmd/policy.go
rmoles: Sounds good. I'll update!

PR #1069, opened 2022-12-14
feat: Add support from hidden suppressions migration command

rmoles:

## Summary

However it was decided that we should add this to the CLI as a more "official" approach.

```
lacework suppressions list aws
lacework suppressions migrate aws
```
list has the following alias `ls`

`lacework suppressions list`
Will allow the users to get their legacy suppressions. Currently they aren't able to view these anywhere else...

`lacework suppressions migrate aws`

#1066 & #1068 should be merged before this

Tested locally ✅ 

## Issue
https://lacework.atlassian.net/browse/45215

### DIFF
--- a/cli/cmd/suppressions.go
+++ b/cli/cmd/suppressions.go
@@ -29,9 +29,6 @@ var (
 )
 
 func init() {
-	// add the suppressions command
-	rootCmd.AddCommand(suppressionsCommand)
-
 	suppressionsCommand.AddCommand(suppressionsListCmd)
 	suppressionsListCmd.AddCommand(suppressionsListAwsCmd)
 
--- a/cli/cmd/suppressions.go
+++ b/cli/cmd/suppressions.go
@@ -1,0 +1,40 @@
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	// top-level cloud-account command
+	suppressionsCommand = &cobra.Command{
+		Use:     "suppressions",
+		Aliases: []string{"suppression", "sup", "sups"},
+		Short:   "Manage legacy suppressions",
+		Long:    "Manage legacy suppressions",
+	}
+
+	// suppressionsListCmd represents the list sub-command inside the suppressions command
+	suppressionsListCmd = &cobra.Command{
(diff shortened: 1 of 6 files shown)


cli/cmd/suppressions.go
```diff
@@ -0,0 +1,57 @@
...
+//
+
+package cmd
+
+import (
+	"github.com/spf13/cobra"
+)
+
+var (
+	// top-level cloud-account command
```
afiune, line 26: this needs an update

cli/cmd/suppressions.go
```diff
@@ -0,0 +1,57 @@
...
+	suppressionsCommand = &cobra.Command{
+		Use:     "suppressions",
+		Hidden:  true,
+		Aliases: []string{"suppression", "sup", "sups"},
```
afiune, line 30: do we really need aliases like `sup` and `sups`? if this is a hidden command I don't think we do need so many aliases.

afiune: But I am ok leaving them 😂 this is a nit

afiune (changes requested):
If the Lacework account does not have any AWS account configured, the list command shows the following error message:
```
ERROR unable to get legacy aws suppressions:
  [GET] https://customerdemo.lacework.net/api/v2/suppressions/aws/allExceptions
  [400] Cannot perform operation. No active AWS accounts
```

Can we improve this and make it just a message without error saying that there are no AWS accounts configured.

afiune:
Also, I feel that the commands should be flipped around, what do you think @rmoles ??
```
$ lacework suppressions aws list
$ lacework suppressions aws migrate

// in the future
$ lacework suppressions gcp list
$ lacework suppressions gcp ....
```

cli/cmd/suppressions.go
rmoles: probably not, but a the same time, why not? 🤷

afiune (approved):
This is way too much code for just one single test, I would highly recommend writing more tests, both unit tests (for functional testing) and integration tests (for end-to-end tests).

Code looks Ok.

PR #1084, opened 2022-12-27
fix(lwlogger): do not override explicit level argument

kolbeinn:
Signed-off-by: Kolbeinn Karlsson <kolbeinn.karlsson@lacework.net>

## Summary

lwlogger silently overrides any level passed to it with the value of the LW_LOG environment variable. This may not cause issues in the main CLI, but for CLI components, LW_LOG is always set, and thus there is no way to create any lwloggers without their level being overridden.

This PR changes it so that the lwlogger level is overridden with LW_LOG only if `level=""`. This is provides maximum flexibility for the logger. You can call `lwlogger.New("")` to have the level be determined by LW_LOG, or you can call `lwlogger.New(myDesiredLevel)` to explicitly set the level.

In the Lacework remediate component, I create a simple function foo with the following logic:
```
func foo(cmd *cobra.Command, args []string) {
	loggerOne := lwlogger.New("INFO")
	loggerTwo := lwlogger.New("")

	loggerOne.Info("Logger one INFO call. xyz123")
	loggerTwo.Info("Logger two INFO call. xyz123")
	loggerOne.Debug("Logger one DEBUG call. xyz123")
	loggerTwo.Debug("Logger two DEBUG call. xyz123")
}
```
Then call the function twice. Once with LW_LOG explicitly set and once without:
```shell
$ lacework-dev remediate foo 2>&1 | grep xyz123
{"level":"info","ts":"2022-12-27T11:25:33Z","caller":"cmd/foo.go:40","msg":"Logger one INFO call. xyz123"}
$ LW_LOG=DEBUG lacework-dev remediate foo 2>&1 | grep xyz123
{"level":"info","ts":"2022-12-27T11:25:37Z","caller":"cmd/foo.go:40","msg":"Logger one INFO call. xyz123"}
{"level":"info","ts":"2022-12-27T11:25:37Z","caller":"cmd/foo.go:41","msg":"Logger two INFO call. xyz123"}
{"level":"debug","ts":"2022-12-27T11:25:37Z","caller":"cmd/foo.go:43","msg":"Logger two DEBUG call. xyz123"}
```
This shows me that `loggerOne` has the INFO level regardless of LW_LOG, while `loggerTwo` follows LW_LOG.

### DIFF
--- a/lwlogger/logger.go
+++ b/lwlogger/logger.go
@@ -51,9 +51,8 @@ var (
 
 // New initialize a new logger with the provided level and options
 func New(level string, options ...zap.Option) *zap.Logger {
-	// give priority to the environment variable
-	if envLevel := LogLevelFromEnvironment(); envLevel != "" {
-		level = envLevel
+	if level == "" {
+		level = LogLevelFromEnvironment()
 	}
 
 	zapConfig := zap.Config{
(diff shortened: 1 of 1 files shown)


hazedav:
LGTM, but I'll let someone from the Growth team weigh in as well...

afiune (approved):
This makes perfect sense, I still would have loved to see some tests.

PR #1088, opened 2022-12-28
feat(GROW-1058): Improve management of component versions

edoardopirovano:
## Summary

This makes two changes to improve managing different versions of components:

- `lacework component install` and `lacework component update` now accept a `--verison` flag to force a specific version to be installed rather than the latest one.

- There is a new command `lacework component show <name>` that shows details for a component, including all available versions.

There's unit tests for the main interesting new utility functions which construct the update message and display the list of available versions to the user. I also exercised this manually to check end to end functionality.

## Issue

https://lacework.atlassian.net/browse/GROW-1058

### DIFF
--- a/cli/cmd/component.go
+++ b/cli/cmd/component.go
@@ -24,6 +24,7 @@ import (
 	"path/filepath"
 
 	"github.com/AlecAivazis/survey/v2"
+	"github.com/Masterminds/semver"
 	"github.com/fatih/color"
 	"github.com/olekukonko/tablewriter"
 	"github.com/pkg/errors"
@@ -56,6 +57,15 @@ var (
 		Short:   "List all components",
 		Long:    `List all available components and their current state`,
 		RunE:    runComponentsList,
+	}
+
+	// componentsShowCmd represents the show sub-command inside the components command
+	componentsShowCmd = &cobra.Command{
+		Use:   "show <component>",
+		Short: "Show details about a component",
+		Long:  "Show details about a component",
+		Args:  cobra.ExactArgs(1),
+		RunE:  runComponentsShow,
 	}
 
 	// componentsInstallCmd represents the install sub-command inside the components command
@@ -94,14 +104,22 @@ var (
 		Args:   cobra.ExactArgs(1),
 		RunE:   runComponentsDevMode,
 	}
+
+	versionArg string
 )
 
 func init() {
 	// add the components command
 	rootCmd.AddCommand(componentsCmd)
 
+	componentsInstallCmd.PersistentFlags().StringVar(&versionArg, "version", "",
+		"require a specific version to be installed (default is latest)")
+	componentsUpdateCmd.PersistentFlags().StringVar(&versionArg, "version", "",
+		"update to a specific version (default is latest)")
+
 	// add sub-commands to the components command
 	componentsCmd.AddCommand(componentsListCmd)
+	componentsCmd.AddCommand(componentsShowCmd)
 	componentsCmd.AddCommand(componentsInstallCmd)
 	componentsCmd.AddCommand(componentsUpdateCmd)
 	componentsCmd.AddCommand(componentsUninstallCmd)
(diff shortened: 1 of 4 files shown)


lwcomponent/component.go
```diff
@@ -603,3 +617,71 @@ func (c Component) EnterDevelopmentMode() error {
 
 	return nil
 }
+
+// ByVersion type to sort an array of versions by their semver comparison
+type ByVersion []semver.Version
```
dmurray-lacework, line 622: nit: This is an unusual name for a type. Is there anything more suitable than `ByVersion`

lwcomponent/component.go
```diff
@@ -603,3 +617,71 @@ func (c Component) EnterDevelopmentMode() error {
...
+
+func (a ByVersion) Len() int           { return len(a) }
+func (a ByVersion) Less(i, j int) bool { return a[i].LessThan(&a[j]) }
+func (a ByVersion) Swap(i, j int)      { a[i], a[j] = a[j], a[i] }
```
dmurray-lacework, line 626: Is this needed for sorting. Is it not possible to do this for the packages github readme?
https://github.com/Masterminds/semver#sorting-semantic-versions

afiune: I agree with Darren, this is exactly whats implemented in the `semver.Collection` type. Can we use it instead?

- https://github.com/Masterminds/semver/blob/master/collection.go

lwcomponent/component.go
```diff
@@ -228,9 +240,9 @@ func (s State) Install(name string) error {
 		return err
 	}
 
-	artifact, found := component.ArtifactForRunningHost()
+	artifact, found := component.ArtifactForRunningHost(version)
 	if !found {
-		return errors.New("unsupported platform")
+		return errors.New("could not find an artifact for version " + version + " on the current platform")
```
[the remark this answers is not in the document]
edoardopirovano: Thanks, done!

cli/cmd/component.go
```diff
@@ -39,6 +41,10 @@ var cdkDevState = struct {
 	Description string
 }{}
 
+var (
+	versionArg string
+)
```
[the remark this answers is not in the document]
edoardopirovano: Thanks, done!

lwcomponent/component.go
```diff
@@ -202,12 +203,23 @@ func Dir() (string, error) {
...
 	}
 
+	if version == "" {
+		version = component.LatestVersion.String()
+	}
+
+	// @edoardopirovano, December 2022: Temporary workaround while there are still old
+	// APIs out there which return only artifacts for the latest version and don't tag
+	// these. This block can be removed some time in the future.
+	if s.Version == "0.2.0" && version == component.LatestVersion.String() {
```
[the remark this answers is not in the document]
edoardopirovano: I've expanded this comment to explain the context more thoroughly. Let me know if that answers your questions! (I'd rather make the comment self-contained than elaborate here on the PR which will be lost to future readers of the code!)

lwcomponent/component.go
edoardopirovano: No longer relevant since this has been removed per your next comment.

lwcomponent/component.go
edoardopirovano: Thanks, that's a good point! I've switched to using that, which required making the array into an array of pointers instead of values to make the types work. I think this is okay, and the unit tests should've caught if this caused issues, but please do check this. (This is my first non-trivial PR in Go, so I'm somewhat lacking in the details of the semantics of the language.)

lwcomponent/component.go
```diff
@@ -603,3 +624,64 @@ func (c Component) EnterDevelopmentMode() error {
...
+	versionToBreadcrumb[&c.LatestVersion] = c.Breadcrumbs.UpdateMessage
+	allVersions := make([]*semver.Version, 0)
+	for k := range versionToBreadcrumb {
+		allVersions = append(allVersions, k)
+	}
+	sort.Sort(semver.Collection(allVersions))
+	return allVersions, versionToBreadcrumb
+}
+
+func (c Component) MakeUpdateMessage(from semver.Version, to semver.Version) string {
```
afiune, line 647: nitpick
```suggestion
func (c Component) MakeUpdateMessage(from, to semver.Version) string {
```

lwcomponent/component.go
```diff
@@ -603,3 +624,64 @@ func (c Component) EnterDevelopmentMode() error {
...
+			if breadcrumbs[version] != "" {
+				// We've found a breadcrumb in the range (from, to] which is one we care about
+				updateMessage += "\n"
+				updateMessage += breadcrumbs[version]
+			}
+		}
+		return fmt.Sprintf("Successfully upgraded component from %s to %s%s", from.String(), to.String(), updateMessage)
+	}
+	return fmt.Sprintf("Successfully downgraded component from %s to %s", from.String(), to.String())
+}
```
afiune, line 669: I love this function so much! ❤️ Thank you

afiune (approved):
Left a few comments but the change looks great! Thank you so much, @edoardopirovano !!

edoardopirovano:
Thanks for the comments @afiune! I'm updating one of the dev environments to contain https://github.com/lacework/rainbow/pull/10874 so that I can do a full end to end test of this and then I'll get this merged 🙂

edoardopirovano:
I've pushed some final minor fixes that I spotted during integration testing and squashed the commit history plus rebased to tidy this up. Will proceed to merge later today if there's no objections.

PR #1089, opened 2022-12-30
feat: default to non-interactive if stdout is not a terminal

slshen:
## Summary

Default to non-interactive output if stdout is not a terminal.

I've happen to have 1.2.0 installed and at the moment the latest version is 1.3.1.

```
; rm ~/.config/lacework/version_check
; lacework configure show account > /tmp/account.txt
; cat /tmp/account.txt
qan.qan.corp

A newer version of the Lacework CLI is available! The latest version is v1.3.1,
to update execute the following command:

  brew upgrade lacework-cli
; rm ~/.config/lacework/version_check
; go run cli/main.go configure show account > /tmp/account.txt
; cat /tmp/account.txt
qan.qan.corp
```

After thinking about it a bit it's not the daily version check that's the problem, it's that we shouldn't assume we're interactive if stdout isn't a terminal.

## Issue

https://lacework.atlassian.net/browse/RAIN-45976

### DIFF
--- a/cli/cmd/cli_state.go
+++ b/cli/cmd/cli_state.go
@@ -32,6 +32,7 @@ import (
 	"github.com/briandowns/spinner"
 	"github.com/fatih/color"
 	prettyjson "github.com/hokaccha/go-prettyjson"
+	"github.com/mattn/go-isatty"
 	"github.com/peterbourgon/diskv/v3"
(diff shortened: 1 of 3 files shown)


afiune (approved):

slshen:
Thanks @afiune .

PR #1104, opened 2023-01-10
feat(cdk): make help command work for components

slshen:
## Summary

The help command should show help for a component instead of help for the CLI.

Run `go run cli/main.go iac help org list` and note that it now displays help for `... iac org list` instead of help for the CLI itself.

## Issue

https://lacework.atlassian.net/browse/RAIN-42930

### DIFF
--- a/cli/cmd/root.go
+++ b/cli/cmd/root.go
@@ -150,6 +150,8 @@ func Execute() (err error) {
 	}()
 	defer cli.Wait()
 
+	setupRootHelpCommand()
+
 	// first, verify if the user provided a command to execute,
 	// if no command was provided, only print out the usage message
 	if noCommandProvided() {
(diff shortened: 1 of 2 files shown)


cli/cmd/root.go
```diff
@@ -166,6 +168,32 @@ func Execute() (err error) {
...
+	helpCommand.RunE = func(cmd *cobra.Command, args []string) error {
+		target, _, _ := rootCmd.Find(args)
+		if target != nil && isComponent(target.Annotations) {
+			f, ok := cli.LwComponents.GetComponent(target.Use)
+			if ok {
+				envs := []string{
+					fmt.Sprintf("LW_COMPONENT_NAME=%s", target.Use),
+				}
+				helpArgs := append([]string{"help"}, args[1:]...)
+				return f.RunAndOutput(helpArgs, envs...)
```
afiune, line 189: This change it nice, my only concern is that if a component doesn't have a `help` command, then we will error out pretty nasty so, could we run the component but capture the `STDERR` and `STDOUT` before displaying it to the user and, if the command returns and exit code other than `0` then print something more helpful, maybe the `defaultRunHelp()` instead?

Here is an example of the output we would see for components that doesn't have a help command:
```
{"level":"error","ts":"2023-01-10T12:04:52-08:00","caller":"cmd/cdk_init.go:94","msg":"LW_ACCOUNT environment variable is not set","stacktrace":"... suppressed output ..."}
Usage:
  lacework help [command] [flags]

Flags:
  -h, --help   help for help

Global Flags:
      --api_token string    access token (replaces the use of api_key and api_secret)
      --organization        access organization level data sets (org admins only)
      --subaccount string   sub-account name inside your organization (org admins only)

➜  go-sdk git:(component-help) ✗
```
Note that I suppressed the looooong stracktrace on purpose, but we all know how long this string will be 😓

afiune (changes requested):
I really like this change, I have one only ask/idea - Happy to elaborate more 1:1 over Slack.

cli/cmd/root.go
slshen: I just checked the existing components in dev and they all support a help command.

sca and sast don't work properly because they expect to always get a client.  The help command as-implemented doesn't give one, which I suppose it could.  Bit odd though to go to the trouble to get a client just to run help.

slshen: Actually if you look closely, it's the component itself complaining that LW_ACCOUNT isn't set, not that it doesn't have a help command. 

I'm assuming you're looking at sca (or sast maybe, also assuming the code is similar.)  It does have a help command, it's just that it insists on getting an api client for the help command.  See https://github.com/lacework-dev/sca/blob/main/cmd/root.go#L66.

afiune: Ill take a look at this early next week, most likely you are right 😄

afiune: Thank you so much @slshen for looking into this. I completely agree with you, I will let @linpengzhang @edoardopirovano @declanwilson-lw know that, for help messages, we don't need to generate an API client.

I am +1 for this change! 💯

afiune (approved):

PR #1120, opened 2023-01-19
feat(cli): add hidden suppressions migrate gcp command

rmoles:

## Summary

```
lacework suppressions migrate gcp
```

`lacework suppressions migrate gcp`

Additionally this PR includes a fix for the `lacework suppressions migrate aws` command. This checks legacy suppressions for ARNs in the resourceNames field and parses out the resource name from the ARN, as ARNs are not supported for policy exceptions.

Tested locally ✅ 

## Issue
https://lacework.atlassian.net/browse/RAIN-46613
https://lacework.atlassian.net/browse/RAIN-47184

### DIFF
--- a/cli/cmd/suppressions.go
+++ b/cli/cmd/suppressions.go
@@ -172,6 +172,8 @@ func convertResourceNamesSupConditions(supConditions []string) []any {
 	var conditions []any
 	for _, condition := range supConditions {
 		ok := arn.IsARN(condition)
+		// If the legacy suppression resourceNames field contains an ARN, we should parse and pull
+		// out the resource name. ARNs are not supported in policy exceptions
 		if ok {
 			parsedEntry, _ := arn.Parse(condition)
 			condition = parsedEntry.Resource
--- a/cli/cmd/suppressions.go
+++ b/cli/cmd/suppressions.go
@@ -19,7 +19,10 @@
 package cmd
 
 import (
+	"github.com/fatih/color"
+	"github.com/lacework/go-sdk/api"
 	"github.com/spf13/cobra"
+	"golang.org/x/exp/slices"
 )
 
 var (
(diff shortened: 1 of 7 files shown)


cli/cmd/suppressions.go
```diff
@@ -63,4 +68,181 @@ func init() {
...
+		for _, exception := range convertedSuppressions {
+			err := cli.OutputJSON(exception)
+			if err != nil {
+				return
+			}
+		}
+		colorizeR := color.New(color.FgRed, color.Bold)
+		cli.OutputHuman(colorizeR.Sprintf("WARNING: Before continuing, " +
+			"please thoroughly inspect the above exceptions to ensure they are valid and" +
+			" required. By continuing, you accept liability for any compliance violations" +
```
garyboyes, line 119: Are we gonna need to get Michael Moore involved for legal advice here 😆

cli/cmd/suppressions.go
```diff
@@ -63,4 +68,181 @@ func init() {
...
+		if condition != "*" {
+			// It appears that for GCP, the resourceName field for policy exceptions is in fact expecting
+			// users to provider the full GCP resource_id.
+			// Example resourceId: //compute.googleapis.com/projects/gke-project-01-c8403ba1/zones/us-central1-a/instances/squid-proxy
+			// This was not the case for legacy suppressions and in most cases it's unlikely that the
+			// users will have provided this. Instead, we are more likely to have
+			// the resource name provided. To cover this scenario we prepend the resource name
+			// from the legacy suppression with "*/" to make it match the resource name while
+			// wildcarding the rest of the resourceId
+			condition = "*/" + condition
```
iok0, line 199: were you able to double check this syntax a dev env.?

garyboyes (commented):
Looks good to me (at least with my v limited Go experience...)
Couple of questions around the functionality, when the user does the migration, does it remove the old suppressions? and, is this only going to be a 1 time operation, or could a customer/user run this multiple times? and what's the impact if they do that?

iok0:
In terms of testing, was there an environment with a bunch of suppressions setup that you could try the migration out? i.e. check the expected exceptions were actually applied on next run?

cli/cmd/suppressions.go
rmoles: Good point. I need to circle back an test this as I wasn't able to yesterday on dev5 or dev7

cli/cmd/suppressions.go
```diff
@@ -63,4 +68,183 @@ func init() {
 	// gcp
 	suppressionsCommand.AddCommand(suppressionsGcpCmd)
 	suppressionsGcpCmd.AddCommand(suppressionsListGcpCmd)
+	suppressionsGcpCmd.AddCommand(suppressionsMigrateGcpCmd)
+}
+
+func autoConvertSuppressions(convertedPolicyExceptions []map[string]api.PolicyException) {
+	cli.StartProgress("Creating policy exceptions ...")
```
dmurray-lacework, line 75: Extreme nitpick, but I don't think we usually leave space between the ellipsis
```suggestion
	cli.StartProgress("Creating policy exceptions...")
```

cli/cmd/suppressions.go
```diff
@@ -63,4 +68,183 @@ func init() {
...
+	for _, exceptionMap := range convertedPolicyExceptions {
+		for policyId, exception := range exceptionMap {
+			response, err := cli.LwApi.V2.Policy.Exceptions.Create(policyId, exception)
+			if err != nil {
+				cli.Log.Debug(err, "unable to create exception")
+				cli.OutputHuman(color.RedString(
+					"Error creating policy exception to create exception. %e"),
```
dmurray-lacework, line 82: Not sure this is the correct placeholder for your err.
https://pkg.go.dev/fmt
```suggestion
					"Error creating policy exception to create exception. %s"),
```

jon-stewart (approved):

cli/cmd/suppressions_gcp.go
```diff
@@ -58,3 +156,213 @@ func suppressionsGcpList(_ *cobra.Command, _ []string) error {
...
+		if err != nil {
+			return err
+		}
+		if confirm {
+			autoConvertSuppressions(convertedPolicyExceptions)
+			printDiscardedSuppressions(discardedSuppressions)
+			cli.OutputHuman(color.GreenString("To view the newly created Exceptions, " +
+				"try running `lacework policy-exceptions list <policyId>"))
+		} else {
+			cli.OutputHuman("Cancelled Legacy Suppression to Exception migration!")
```
dmurray-lacework, line 235: nit, need a new line here to stop % being output
```suggestion
			cli.OutputHuman("Cancelled Legacy Suppression to Exception migration!\n")
```

dmurray-lacework (approved):

cli/cmd/suppressions.go
rmoles: Confirmed on dev8 that the suppression format is working as expected. The only concern would be if a user has already provided the full resource ID. However I don't believe this format was used anywhere in the legacy compliance checks.

PR #1122, opened 2023-01-19
fix(cli): ec2ssm use runner region, not cfg region

nschmeller:
Calls to the EC2 API currently fail when the region of the runner is
different from the region of the credentials used to authenticate the
request.

This commit modifies the EC2 client that we create during `ec2ssm`
execution to use `run.Region` instead of `cfg.Region`.

Fixes RAIN-47298

## Summary

Found a bug during a customer deployment around this. There was a workaround to use credentials with the same region as the runners, but we should be region agnostic.

Adding an instance in a different region to the `ec2ssm` integration test sandbox infra, PR to come

## Issue

https://lacework.atlassian.net/browse/RAIN-47298


nschmeller:
Working integration test in https://github.com/lacework/agent/pull/1732, good thing I waited to make sure they were running properly, they caught a bug!

PR #1127, opened 2023-01-25
Avoid unnecessary memory consumption of vulnerability cmds

afiune:
## Summary

We are using a lot of resources in encoding and decoding JSON (API responses). It is really
A LOT of data that API v2 returns, that is, hundreds of thousands of data-points just for a
single command. (max limit of rows is `500,000` and some customers reach those numbers)

This PR is adding a number of changes to optimize the command:
```
lacework vulnerability container list-assessments
```

## feat(cli): improve caching using hash of filters
When caching, we depend on a key to store and retrieve data, this cache
key should change depending on filters that will affect the data
returned by our APIs.

This change implements a function `hash()` to generate unique hashes of
arbitrary values in Go which is useful to generate unique cache keys of
filters applied to commands.

It also implements a second function `generateContainerVulnListCacheKey()`
that returns the cache key for vulnerability assessments. The criteria to
generate this cache key is to use the prefix `vulnerability/container/v2_{HASH}`
with a hash value at the end of the prefix. The hash is generated based of filters
that, if changed by the user, will change the data returned from our APIs.

## fix(cli): --registry flag fails faster for list-assessments command
If the flag `--registry` is provided with a container registry that is not found, instead of
executing the command and using unnecessary resources, we fail fast with a helpful message:
```
ERROR container registry 'index.docker.io' not found

Your account has the following container registries configured:

    > abc
    > xyz

To integrate a new container registry use the command:

    lacework container-registry create
```

## refactor(cli): avoid unnecessary memory consumption
The new improved process to get the list of container assessments is:

1) Check if the user provided a list of registries and repositories,
   if so, use those filters instead of fetching the entire data from
   all registries, repositories, local scanners, etc. (This is a memory
   utilization improvement)
2) If no filter by registries and/or repos, then fetch all data from all
   registries and all local scanners, we purposely split them in two search
   requests since there could be so much data that we get to the 500,000 rows
   if data and we could potentially miss some information
3) Either 1) or 2) will generate a tree of unique container vulnerability
   assessments (see the `treeCtrVuln` type), with this tree we will generate
   one last API request to unique evaluations per image (This is a memory
   utilization improvement)
4) Finally, if we get information from the queried assessments, we build a
   summary that will ultimately get stored in the cache for subsequent commands

`treeCtrVuln` and `ctrVuln` are types that help us generate an tree of container
vulnerability assessments that are unique per image id, that is, there will
never be duplicates of the same image with different evaluation guids (evalGuid)

## fix(cli): --fixable and other filters

We were using a `switch` statement for the vulnerability filters and this implementation
was causing that only one filter can be applied and when more than one was provided,
the rest would never be executed/applied.

The right implementation is to use multiple `if` statements.

## feat(api): v2PageMetadata to log page information

A new type used to compute the total pages and the page number automatically
when reading pages using the `client.NextPage()` function

Inside the `Pageable` interface, there are a few new functions that are
automatically implemented when attaching the `v2PageMetadata` type
into any struct that implements `Pageable`, so attaching that struct is
now a requirement.

## refactor(cli): use evaluation GUID to retrieve scan results

This change improves the command `lacework vuln ctr scan` to use
the new field `evalGuid` to retrieve the assessment results.

More information at https://lacework.atlassian.net/browse/RAIN-46724

## refactor(cli): deprecate --image_id flag

Instead of having the user to guess between the Image Digest or Image ID,
the search is so fast and accurate that we can check both with the provided
sha256 argument.

As part of this refactor, I also started caching the assessments results when
running:
```
$ lacework vuln ctr show <sha256:hash>
```
Further commands will be blasting fast after the cache expires after 30 minutes.

## Issue
* https://lacework.atlassian.net/browse/GROW-1371
* https://lacework.atlassian.net/browse/RAIN-46724
* https://lacework.atlassian.net/browse/LINK-1273
* https://lacework.atlassian.net/browse/LINK-1152

### DIFF
--- a/cli/cmd/vuln_container.go
+++ b/cli/cmd/vuln_container.go
@@ -102,10 +102,14 @@ func init() {
 		vulContainerListAssessmentsCmd.Flags(),
 	)
 
+	// DEPRECATED
 	vulContainerShowAssessmentCmd.Flags().BoolVar(
 		&vulCmdState.ImageID, "image_id", false,
-		"(DEPRECATED) by default we look up both, image_id and image_digest",
+		"tread the provided sha256 hash as image id",
 	)
+	errcheckWARN(vulContainerShowAssessmentCmd.Flags().MarkDeprecated(
+		"image_id", "by default we now look up both, image_id and image_digest at once.",
+	))
 }
 
 func setPollFlag(cmds ...*flag.FlagSet) {
--- a/cli/cmd/vuln_container.go
+++ b/cli/cmd/vuln_container.go
@@ -104,7 +104,7 @@ func init() {
 
 	vulContainerShowAssessmentCmd.Flags().BoolVar(
 		&vulCmdState.ImageID, "image_id", false,
-		"tread the provided sha256 hash as image id",
+		"(DEPRECATED) by default we look up both, image_id and image_digest",
 	)
 }
 
--- a/api/agent_info.go
+++ b/api/agent_info.go
@@ -35,6 +35,8 @@ func (svc *AgentInfoService) Search(response interface{}, filters SearchFilter) 
 type AgentInfoResponse struct {
 	Data   []AgentInfo  `json:"data"`
 	Paging V2Pagination `json:"paging"`
+
+	v2PageMetadata `json:"-"`
 }
 
 // Fulfill Pageable interface (look at api/v2.go)
@@ -43,6 +45,7 @@ func (r AgentInfoResponse) PageInfo() *V2Pagination {
 }
 func (r *AgentInfoResponse) ResetPaging() {
 	r.Paging = V2Pagination{}
+	r.Data = nil
 }
 
 type AgentInfo struct {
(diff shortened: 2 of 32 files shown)


dmurray-lacework:
Possibly an issue with the fixable field not being returned in the cached response

dmurray-lacework (approved):
Approved. Left a comment about the fixable field. But it looks to be preexisting and can be tackled outside this PR.

afiune:
@dmurray-lacework I fixed the issue with the fixable field not being returned. It was really simple.

cli/cmd/vuln_container.go
```diff
@@ -104,7 +104,7 @@ func init() {
 
 	vulContainerShowAssessmentCmd.Flags().BoolVar(
 		&vulCmdState.ImageID, "image_id", false,
-		"tread the provided sha256 hash as image id",
+		"(DEPRECATED) by default we look up both, image_id and image_digest",
```
dmurray-lacework, line 107: Cobra command has a `MarkDeprecated` function for flags. Can we mark use of this?
github.com/spf13/pflag/flag.go:404

afiune: Will do.

PR #1128, opened 2023-01-25
feat(cli): `ec2ssm` no longer reuses IAM role

nschmeller:
`ec2ssm` currently will reuse existing IAM roles + instance profiles from a previous
invocation. This was useful in avoiding extra IAM operations in the
event of an interrupt or a crash. However, this functionality meant that
`ec2ssm` was not idempotent and could not be run concurrently. This
property is true now.

This commit:

* adds a random identifier to the names of new IAM roles and profiles
* makes an AWS API call to disassociate instance profiles on exit
* removes checks for existing IAM roles and profiles

Fixes RAIN-47837

Running integration tests concurrently

## Issue

https://lacework.atlassian.net/browse/RAIN-47837


cirego (approved):
Looks good to me!

PR #1129, opened 2023-01-27
feat(cli): add hidden suppressions migrate azure command

rmoles:

## Summary

```
lacework suppressions migrate azure
```

`lacework suppressions migrate azure`

Additionally this PR includes a fix for the `lacework suppressions migrate aws` command. This checks legacy suppressions for ARNs in the resourceNames field and parses out the resource name from the ARN, as ARNs are not supported for policy exceptions.

Tested locally ✅ 

## Issue
https://lacework.atlassian.net/browse/RAIN-46612
https://lacework.atlassian.net/browse/RAIN-47184


afiune (commented):
Unit tests are not happy... 😭

dmurray-lacework (approved):

PR #1135, opened 2023-02-06
feat(api): Add support for the `ComponentData` API end-points

edoardopirovano:
## Summary

This PR adds support in the Go SDK for having components upload data back to the Lacework platform via the new `/v2/ComponentData` API end-points (added in https://github.com/lacework/rainbow/pull/11260). Rather than allowing the end-points to be called individually (which there is no intended use-case for), this adds a single new function which handles the whole flow of performing an upload: that is, requesting from the API server some pre-signed URLs to upload data to S3, performing the upload, and then sending the API server a confirmation that the upload was completed successfully. This encapsulates this flow in the SDK rather than duplicating it accross components so that if we need to change it in future we only need to change it here.

Each of the network steps includes retrying with an expontential back-off, and if any step fails after this retrying then we abort and return that error. This may leave a partially completed upload in S3, which is okay: without sending the final confirmation the upload won't be consumed and the failed attempt will get removed automatically after our (fairly short) retention period expires.

New unit tests that call a mock API. Also, manual end-to-end testing against one of our development instances (still in progress at time of writing, but I won't merge until that's completed successfully).

Edit: Manual testing is now complete and working 🎉

## Issue

https://lacework.atlassian.net/browse/RAIN-46483

### DIFF
--- a/api/component_data.go
+++ b/api/component_data.go
@@ -25,45 +25,45 @@ type DocumentSpec struct {
 	Size int64  `json:"size"`
 }
 
-type InitialResponseRaw struct {
-	Data *InitialResponse `json:"data,omitempty"`
+type ComponentDataInitialResponseRaw struct {
+	Data *ComponentDataInitialResponse `json:"data,omitempty"`
 }
 
-type InitialResponse struct {
-	Guid          string          `json:"guid,omitempty"`
-	UploadMethods []*UploadMethod `json:"uploadMethods,omitempty"`
+type ComponentDataInitialResponse struct {
+	Guid          string                       `json:"guid,omitempty"`
+	UploadMethods []*ComponentDataUploadMethod `json:"uploadMethods,omitempty"`
 }
 
-type UploadMethod struct {
+type ComponentDataUploadMethod struct {
 	Method string            `json:"method,omitempty"`
 	Info   map[string]string `json:"info,omitempty"`
 }
 
-type CompleteRequest struct {
+type ComponentDataCompleteRequest struct {
 	Guid string `json:"guid"`
 }
 
-type CompleteResponseRaw struct {
-	Data *CompleteResponse `json:"data,omitempty"`
+type ComponentDataCompleteResponseRaw struct {
+	Data *ComponentDataCompleteResponse `json:"data,omitempty"`
 }
 
-type CompleteResponse struct {
+type ComponentDataCompleteResponse struct {
 	Guid string `json:"guid,omitempty"`
 }
 
 func (svc *ComponentDataService) UploadFiles(name string, tags []string, paths []string) (string, error) {
-	initialRequest, err := buildInitialRequest(name, tags, paths)
+	initialRequest, err := buildComponentDataInitialRequest(name, tags, paths)
 	if err != nil {
 		return "", err
 	}
-	var initialResponse InitialResponseRaw
-	err = doWithExponentialBackoff(func() error {
-		return svc.client.RequestEncoderDecoder(http.MethodPost, "v2/ComponentData/requestUpload", initialRequest, &initialResponse)
+	var initialResponse ComponentDataInitialResponseRaw
+	err = doWithExponentialBackoffWaiting(func() error {
+		return svc.client.RequestEncoderDecoder(http.MethodPost, apiV2ComponentDataRequest, initialRequest, &initialResponse)
 	})
 	if err != nil {
 		return "", err
 	}
-	var chosenMethod *UploadMethod
+	var chosenMethod *ComponentDataUploadMethod
 	for _, method := range initialResponse.Data.UploadMethods {
 		if method.Method == "AwsS3" {
 			chosenMethod = method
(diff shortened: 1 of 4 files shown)


api/component_data.go
```diff
@@ -0,0 +1,149 @@
...
+}
+
+func (svc *ComponentDataService) UploadFiles(name string, tags []string, paths []string) (string, error) {
+	initialRequest, err := buildInitialRequest(name, tags, paths)
+	if err != nil {
+		return "", err
+	}
+	var initialResponse InitialResponseRaw
+	err = doWithExponentialBackoff(func() error {
+		return svc.client.RequestEncoderDecoder(http.MethodPost, "v2/ComponentData/requestUpload", initialRequest, &initialResponse)
```
afiune, line 61: All endpoints should be defined inside `api/api.go`.

api/component_data.go
```diff
@@ -0,0 +1,149 @@
...
+		if err != nil {
+			return "", err
+		}
+	}
+	completeRequest := CompleteRequest{
+		Guid: initialResponse.Data.Guid,
+	}
+	var completeResponse CompleteResponseRaw
+	err = doWithExponentialBackoff(func() error {
+		return svc.client.RequestEncoderDecoder(http.MethodPost, "v2/ComponentData/completeUpload", completeRequest, &completeResponse)
```
afiune, line 88: same, this endpoints should be defined inside `api/api.go` as a constant

api/component_data.go
```diff
@@ -0,0 +1,149 @@
...
+type CompleteResponseRaw struct {
+	Data *CompleteResponse `json:"data,omitempty"`
+}
+
+type CompleteResponse struct {
+	Guid string `json:"guid,omitempty"`
+}
+
+func (svc *ComponentDataService) UploadFiles(name string, tags []string, paths []string) (string, error) {
+	initialRequest, err := buildInitialRequest(name, tags, paths)
```
afiune, line 55: can you make this function names more explicit? `buildInitialRequest()` sounds very generic, an idea could be `componentDataBuildInitialRequest()` or something like that.

api/component_data.go
```diff
@@ -0,0 +1,149 @@
...
+		return err
+	}
+	_, err = svc.client.Do(req)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func doWithExponentialBackoff(f func() error) error {
```
afiune, line 136: any way to unit test this function?

api/component_data.go
```diff
@@ -0,0 +1,149 @@
...
+	defer file.Close()
+	req, err := http.NewRequest(http.MethodPut, uploadUrls[filepath.Base(path)], file)
+	if err != nil {
+		return err
+	}
+	_, err = svc.client.Do(req)
+	if err != nil {
+		return err
+	}
+	return nil
```
afiune, line 133: nit: no need to check for error
```suggestion
	_, err = svc.client.Do(req)
	return err
```

api/component_data.go
```diff
@@ -0,0 +1,149 @@
...
+	}
+	return &ComponentDataInitialRequest{
+		Name:             name,
+		Tags:             tags,
+		SupportedMethods: []string{"AwsS3"},
+		Documents:        documents,
+	}, nil
+}
+
+func putFileToS3(svc *ComponentDataService, path string, uploadUrls map[string]string) error {
```
afiune, line 119: why not put this function attached to the `ComponentDataService` struct?

api/component_data.go
```diff
@@ -0,0 +1,149 @@
...
+	Guid          string          `json:"guid,omitempty"`
+	UploadMethods []*UploadMethod `json:"uploadMethods,omitempty"`
+}
+
+type UploadMethod struct {
+	Method string            `json:"method,omitempty"`
+	Info   map[string]string `json:"info,omitempty"`
+}
+
+type CompleteRequest struct {
```
afiune, line 42: Very generic struct name. if this is only for the component data endpoint, a prefix would be helpful.

api/component_data.go
```diff
@@ -0,0 +1,149 @@
...
+type DocumentSpec struct {
+	Name string `json:"name"`
+	Size int64  `json:"size"`
+}
+
+type InitialResponseRaw struct {
+	Data *InitialResponse `json:"data,omitempty"`
+}
+
+type InitialResponse struct {
```
afiune, line 32: same, might be better to add a prefix.

api/component_data.go
edoardopirovano: Thanks, done.

api/component_data.go
edoardopirovano: Thanks, done.

api/component_data.go
edoardopirovano: Sure, done.

api/component_data.go
edoardopirovano: Yep, we don't want to actually wait inside unit tests, but if we pass in a function that defines how the waiting should happen then we can pass in a stub for that in the unit test instead of actually calling `time.sleep`. I've done that.

api/component_data.go
edoardopirovano: Thanks, done.

api/component_data.go
edoardopirovano: Sure, done.

api/component_data.go
edoardopirovano: Thanks, changed.

api/component_data.go
edoardopirovano: Thanks, changed.

edoardopirovano:
@afiune I've pushed up some minor changes to get things working end-to-end. I can now confirm this is all working against one of our dev environments, and from my point of view this is ready to merge.

afiune (approved):

PR #1136, opened 2023-02-06
fix: change ParseRelative time to use UTC instead of local time

dmurray-lacework:
## Summary

Relative time function sets "now" to local time rather than UTC

This has been causing issues in search apis where time filter is provided.
```
  [400] endTime cannot be after current time
```

Set timezone to gmt +1, then run 
`lacework vuln ctr list-assessments`

## Issue

- https://lacework.atlassian.net/browse/GROW-1393
- https://lacework.atlassian.net/browse/LINK-1322


afiune (approved):

hazedav:
It's not clear to me how this would fix anything....

PR #1160, opened 2023-02-23
fix(cli): inline/proxy scanner Ok state

afiune:
## Summary
Both, inline/proxy scanner integrations don’t have a `state`
object at all- so we will hard code them to be always `Ok`.

Added tests.

## Issue
https://lacework.atlassian.net/browse/LINK-1361

### DIFF
--- a/api/alerts_details_integrations.go
+++ b/api/alerts_details_integrations.go
@@ -54,7 +54,7 @@ func (c AlertIntegrationChannel) StateString() string {
 	if c.State.Ok {
 		return "Ok"
 	}
-	return "Check"
+	return "Pending"
 }
 
 type AlertIntegrationContext struct {
--- a/api/cloud_accounts.go
+++ b/api/cloud_accounts.go
@@ -238,7 +238,7 @@ func (c v2CommonIntegrationData) StateString() string {
 	if c.State != nil && c.State.Ok {
 		return "Ok"
 	}
-	return "Check"
+	return "Pending"
 }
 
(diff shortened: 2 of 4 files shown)


cli/cmd/container_registry.go
```diff
@@ -63,12 +63,22 @@ func init() {
 func containerRegistriesToTable(containerRegistries []api.ContainerRegistryRaw) [][]string {
 	var out [][]string
 	for _, cadata := range containerRegistries {
+
+		// @afiune LINK-1361
+		var state string
+		switch cadata.ContainerRegistryType() {
+		case api.InlineScannerContainerRegistry, api.ProxyScannerContainerRegistry:
+			state = "Ok"
```
hazedav, line 71: nit.  Would this logic be better incorporated here:
https://github.com/lacework/go-sdk/blob/f46a8c8ec140c482e8b38230581540405bd53f94/api/cloud_accounts.go#L237

hazedav (approved):
Approved w/ comments

cli/cmd/container_registry.go
afiune: I can't do it there but, I think I can create another `StateString()` function at the `ContainerRegistryRaw` struct instead.

afiune:
@hazedav I met with @dmurray-lacework and @djmctavish to brainstorm and we came up with this! 

Pretty slick! what do you think?

hazedav (approved):

PR #1163, opened 2023-02-28
feat(cli): Add 'subtype' filter flag to 'lacework report-definitions …

dmurray-lacework:
…list' cmd

## Summary

Add a new filter flag to report-definitions list cmd to filter only specified clouds
`lacework report-definitions list --subtype AWS`
`lacework report-definitions list --subtype GCP`
`lacework report-definitions list --subtype Azure`

- [x] Manually run `lacework report-definitions list --subtype <cloud>`
- [x] Unit test `TestFilterReportDefinitionsWithSubType`
- [x] Integrations Tests `TestReportDefintionsListWithSubtype` `TestReportDefintionsListJsonWithSubtype` 

## Issue

https://lacework.atlassian.net/browse/GROW-1403

### DIFF
--- a/cli/cmd/report_definitions.go
+++ b/cli/cmd/report_definitions.go
@@ -22,12 +22,18 @@ import (
 	"strings"
 
 	"github.com/lacework/go-sdk/api"
+	"github.com/lacework/go-sdk/internal/array"
 	"github.com/olekukonko/tablewriter"
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 var (
+	reportDefinitionsCmdState = struct {
+		// filter report definitions by subtype. 'Aws', 'GCP' or 'Azure'
+		SubType string
+	}{}
+
 	// report-definitions command is used to manage lacework report definitions
 	reportDefinitionsCommand = &cobra.Command{
 		Use:     "report-definition",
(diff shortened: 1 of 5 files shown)


cli/cmd/report_definitions.go
```diff
@@ -22,12 +22,18 @@ import (
...
 	"github.com/pkg/errors"
 	"github.com/spf13/cobra"
 )
 
 var (
+	reportDefinitionsCmdState = struct {
+		// filter report definitions by subtype. 'Aws', 'GCP' or 'Azure'
+		SubType string
+	}{}
+
```
djmctavish, line 36: Just a single string definition for this struct, is this for future expansion?

integration/test_resources/help/report-definition_list
```diff
@@ -7,7 +7,8 @@ Aliases:
   list, ls
 
 Flags:
-  -h, --help   help for list
+  -h, --help             help for list
+      --subtype string   filter report definitions by subtype. 'Aws', 'GCP' or 'Azure'
```
djmctavish, line 11: The help for this command gives the following
`--subtype string   filter report definitions by subtype. 'Aws', 'GCP' or 'Azure'`

This does not work with Aws, Gcp they need to match the full AWS, GCP, Azure etc. Perhaps make the user input choice uppercase for the check for Gcp and Aws?

Also if the user inputs `lacework report-definition list --subtype AZURE `then the error is 
`ERROR 'AZURE' is not valid. Report definitions subtype can be [AWS Azure GCP]

djmctavish (commented):
Checked out this branch and run the commands, just a few minor obbservations

djmctavish:
I also see that the cli command line is actually `report-definition` NOT `report-definitions`, not sure that we want to change that in the CLI, possible change the command examples in the PR summary to match!

cli/cmd/report_definitions.go
dmurray-lacework: Yeh, this is the pattern we use for flags. Though you are right it is not needed right now. I'll leave it here for any potential future flags

dmurray-lacework:
> I also see that the cli command line is actually `report-definition` NOT `report-definitions`, not sure that we want to change that in the CLI, possible change the command examples in the PR summary to match!

We have it aliased for 'report-definitions' but you're right 'report-definition' is the actual name

djmctavish (approved):

PR #1170, opened 2023-03-06
fix(cdk): handle -v flag for components

slshen:
## Summary

A bug in the component arg parser makes `lacework sca -v` hang.

What `lacework sca -v` is supposed to do is print the version (where "supposed to do" means what cobra is trying to do.). This isn't done correctly either.

There's a unit test to check that -v is parsed correctly.    And run by hand to verify that `lacework sca -v` prints the version of the sca command.

## Issue

GROW-1438


linpengzhang (approved):
lgtm

PR #1174, opened 2023-03-07
feat: Allow logging format to be chosen by caller

edoardopirovano:
For SAST, we'd like to be able to select the logging format explicitly rather than have it be set from the environment.


linpengzhang (approved):
lgtm

PR #1187, opened 2023-03-15
fix: honeyvent flag values

jon-stewart:
## Summary

Honeycomb event flags are now only the flags

Unit testing and manual

## Issue

https://lacework.atlassian.net/browse/GROW-1436


dmurray-lacework (approved):
This looks good!

Would be to see a screenshot of the change in Honeycomb

PR #1190, opened 2023-03-15
feat: Add retrying to component listing

edoardopirovano:
With SAST and SCA getting increasing usage internally we've stated to hit some API rate limits at installation time. We've already put a mitigation place by starting to cache the installed components in the GitHub Actions cache (https://github.com/lacework/code-security-action/pull/44), but ideally we'd like the installation to be a bit more robust to occasional network glitches and/or rate limits being hit. I suggest adding the retrying code in this PR to our call to the component listing end-point. There's probably more places we could have retrying, but this is pretty much the only place I've actually had reports of failures happening, so let's stick to just here for now.

cc. @lacework/codesec-action

### DIFF
--- a/lwcomponent/component.go
+++ b/lwcomponent/component.go
@@ -31,9 +31,11 @@ import (
 	"runtime"
 	"sort"
 	"strings"
+	"time"
 
 	"aead.dev/minisign"
 	"github.com/Masterminds/semver"
+	"github.com/cenkalti/backoff/v4"
 	"github.com/pkg/errors"
 
 	"github.com/lacework/go-sdk/api"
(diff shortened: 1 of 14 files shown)


vendor/github.com/cenkalti/backoff/v4/.gitignore
jeremydubreil, line 1: Sorry for the noob question: is it expected to have these files checked in?

dmurray-lacework: It is -> https://github.com/golang/dep/blob/master/docs/FAQ.md#should-i-commit-my-vendor-directory

jeremydubreil: Okay, thanks!

lwcomponent/component.go
```diff
@@ -79,6 +83,13 @@ func LoadState(client *api.Client) (*State, error) {
 	return nil, errors.New("invalid api client")
 }
 
+func backoffStrategy() *backoff.ExponentialBackOff {
```
jeremydubreil, line 86: Is this chosen at random?

jeremydubreil (approved):
Looks good to me to fix the failures.

afiune (commented):
Would have loved to see some tests here.

edoardopirovano:
> Would have loved to see some tests here.

A fair point. Added one in https://github.com/lacework/go-sdk/pull/1193.

PR #1191, opened 2023-03-15
fix(cli): vuln host scan-pkg-manifest only show 'VULNERABLE' results

dmurray-lacework:
## Summary

This change removes "GOOD" results from `lacework vuln host scan-pkg-manifest` and only shows "VULNERABLE"

Tests:
- [x] `TestFilterHostScanPackagesVulnDetails`
- [x] `TestFilterHostScanPackagesVulnDetailsFixable`

### DIFF
--- a/cli/cmd/vuln_host_scan_package_manifest.go
+++ b/cli/cmd/vuln_host_scan_package_manifest.go
@@ -201,7 +201,6 @@ func hostScanPackagesVulnToTable(scan api.VulnerabilitySoftwarePackagesResponse)
 			"Package",
 			"Version",
 			"Fix Version",
-			"Status",
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


cli/cmd/vuln_host_scan_package_manifest.go
```diff
@@ -245,7 +244,7 @@ func filterHostScanPackagesVulnDetails(vulns []api.VulnerabilitySoftwarePackage)
 	out := make([]api.VulnerabilitySoftwarePackage, 0)
 
 	for _, vuln := range vulns {
-		if vulCmdState.Fixable && !vuln.HasFix() {
+		if vulCmdState.Fixable && !vuln.HasFix() || !vuln.IsVulnerable() {
```
afiune, line 247: This is wrong. This will only work when the user passes the `--fixable` flag.

We need this in a different if statement.

dmurray-lacework: Added tests and fixed this

afiune (approved):

PR #1198, opened 2023-03-21
fix(cli): invalid JSON for 'vuln ctr show' command

afiune:
## Summary

An array/list `[]` is NOT a valid JSON.

Now we return instead:
```
{ "vulnerabilities": [ ... ] }
```

## Issue

Talked with customer directly ❤️


cli/cmd/vuln_container_test.go
```diff
@@ -413,6 +430,157 @@ var rawListAssessments = `
 ]
 }`
 
+var mockJSONOutputFromShowAssessmentCmd = `{
+  "vulnerabilities": [
```
afiune, line 434: This is what we needed.

afiune:
Well, @hazedav just smacked me like Will...

PR #1199, opened 2023-03-22
ci: disable report_definitions windows ci

djmctavish:

### DIFF
--- a/integration/report_definitions_create_test.go
+++ b/integration/report_definitions_create_test.go
@@ -1,5 +1,6 @@
 //go:build report_definitions && !windows
 
+//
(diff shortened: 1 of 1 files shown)


dmurray-lacework (approved):

afiune:
Why we did not check windows?

afiune:
@djmctavish @dmurray-lacework also, this essentially disables ALL report definition tests since we didn't add a tag to the Makefile directive https://github.com/lacework/go-sdk/blob/main/Makefile#L59-L81

afiune:
Which also needs to be added here https://github.com/lacework/go-sdk/blob/main/scripts/codefresh.ps1#L21-L23

PR #1200, opened 2023-03-23
feat(alerts): alert open to use new UI 

hazedav:
## Summary
`lacework alert open` and the breadcrumb for `lacework alert show` should be using the generally available Alerts UI.

- From a code standpoint we will now use `alertLinkBuilder` in favor of `eventLinkBuilder`.
- The alertLinkBuilder properly supports sub-accounts by passing the `accountName` query argument.

Existing automated testing
Manual testing:
- [ ] UI Logged into Org Account and `lacework open` on Sub Account
- [ ] UI Logged into Sub Account and `lacework open` on Org Account
- [ ] UI Logged into Sub Account 1 and `lacework open` on Sub Account 2

## Issue
https://lacework.atlassian.net/browse/GROW-1233

### DIFF
--- a/cli/cmd/alert.go
+++ b/cli/cmd/alert.go
@@ -104,15 +104,8 @@ func init() {
 
 // Generates a URL similar to:
 //
-//	=> https://account.lacework.net/ui/investigate/recents/EventDossier-123
-func eventLinkBuilder(id string) string {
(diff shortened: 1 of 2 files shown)


afiune (commented):
Do we have tests around that need to be updated? if not, could we add at least one to test `alertLinkBuilder()`? (thinking about avoiding future breaking changes)

afiune (approved):
Also, after tests are added......

PR #1209, opened 2023-03-28
feat(cli): new report-definitions revert command

dmurray-lacework:
## Summary

Changes in this PR:
Support new cmd `lacework report-definitions revert <id> <version>`
Support new cmd `lacework report-definitions diff <id> <version> <version>`

Add new `--version` flag to `lacework report-definitions show` cmd: 

  `lacework report-definitions show <id> --version <version>`
  `lacework report-definitions show <id> --version all`

Manual
- [x] run `lacework report-definitions revert <id> <version>`
- [x] run `lacework report-definitions show <id> --version <version>`
- [x] run `lacework report-definitions show <id> --version all`
- [x] run `lacework report-definitions diff <id> <version> <version>`

Tests
- [x] `TestReportDefinitionsRevert`
- [x] `TestReportDefinitionsDiff`
- [x] `TestReportDiff`

## Issue

https://lacework.atlassian.net/browse/GROW-1487

### DIFF
--- a/cli/cmd/report_definitions.go
+++ b/cli/cmd/report_definitions.go
@@ -246,6 +246,7 @@ func init() {
 	reportDefinitionsCommand.AddCommand(reportDefinitionsDeleteCommand)
 	reportDefinitionsCommand.AddCommand(reportDefinitionsUpdateCommand)
 	reportDefinitionsCommand.AddCommand(reportDefinitionsRevertCommand)
+	reportDefinitionsCommand.AddCommand(reportDefinitionsDiffCommand)
 
 	// add flags to report-definition commands
 	reportDefinitionsShowCommand.Flags().StringVar(&reportDefinitionsCmdState.Version,
--- a/cli/cmd/report_definitions.go
+++ b/cli/cmd/report_definitions.go
@@ -19,6 +19,7 @@
 package cmd
 
 import (
+	"fmt"
 	"strconv"
 	"strings"
 
@@ -52,6 +53,8 @@ var (
 		SubType string
 		// create report definitions from a file input
 		File string
+		// retrieve report definition by version
+		Version string
 	}{}
 
 	// report-definitions command is used to manage lacework report definitions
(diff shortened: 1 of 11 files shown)


jon-stewart (approved):

cli/cmd/report_definitions.go
```diff
@@ -160,6 +167,64 @@ var (
...
+
+		return cli.OutputJSON(response)
+	}
+
+	if version, err = strconv.Atoi(reportDefinitionsCmdState.Version); err != nil {
+		return errors.Wrap(err, "unable to parse version")
+	}
+
+	cli.StartProgress(fmt.Sprintf("Fetching report definition version %d...", version))
+	response, err := cli.LwApi.V2.ReportDefinitions.GetVersions(id)
```
afiune, line 206: nit - If I run:
```
lacework rd show <id> --version 3432
```
I would get an error from the APIs saying that it does not exist (I imagine), but it would be better error message to say also which versions actually exist.

cli/cmd/report_definitions.go
```diff
@@ -160,6 +167,64 @@ var (
...
+	if reportDefinitionsCmdState.Version == "all" {
+		cli.StartProgress("Fetching all report definition versions...")
+		response, err := cli.LwApi.V2.ReportDefinitions.GetVersions(id)
+		cli.StopProgress()
+
+		if err != nil {
+			return err
+		}
+
+		return cli.OutputJSON(response)
```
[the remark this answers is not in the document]
afiune: Also, there is no Human Readable output?

cli/cmd/report_definitions_diff.go
```diff
@@ -0,0 +1,166 @@
...
+	"gopkg.in/yaml.v3"
+)
+
+// reportDefinitionsDiffCommand command is used to compare 2 lacework report definition versions
+var reportDefinitionsDiffCommand = &cobra.Command{
+	Use:   "diff <report_definition_id> <version> <version>",
+	Short: "Compare 2 versions of a report definition",
+	Long: `Compare 2 versions of a report definition.
+
+To see a diff of 2 report definition versions:
```
afiune, line 39: Run this by the docs team but, I think we should use numbers here.
```suggestion
	Short: "Compare two versions of a report definition",
	Long: `Compare two versions of a report definition.

To see a diff of two report definition versions:
```

cli/cmd/report_definitions_diff.go
```diff
@@ -0,0 +1,166 @@
...
+				object: r,
+			}
+		}
+	}
+
+	if reportOne == nil || reportTwo == nil {
+		return errors.New("unable to find report definition versions")
+	}
+
+	diff, err := diffAsYamlString(*reportOne, *reportTwo)
```
afiune, line 91: What if both versions are the same?

cli/cmd/report_definitions_revert.go
```diff
@@ -0,0 +1,67 @@
...
+	"strconv"
+
+	"github.com/pkg/errors"
+	"github.com/spf13/cobra"
+)
+
+// revert command is used to rollback lacework report definition to a previous version
+var reportDefinitionsRevertCommand = &cobra.Command{
+	Use:     "revert <report_definition_id> <version>",
+	Aliases: []string{"restore"},
```
afiune, line 31: I thought @marcosgm said to use `restore`? 🤔 - With alias to `revert`

The reason it is important to choose one is because the entire output of this command will say:
```
Restoring report definition from version 'X' ...
```
and
```
The report definition 'ABC' was restored to version 'X'
````

afiune:
check windows

cli/cmd/report_definitions_diff.go
dmurray-lacework: response is "" same as a git diff if results are the same
We could prevent the user comparing the same version.

cli/cmd/report_definitions.go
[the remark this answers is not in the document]
dmurray-lacework: I think it could be of use to view all versions in 1 cmd. Or at least give the option.

PR #1222, opened 2023-04-06
fix: adjust endtime for natural time

dmurray-lacework:
## Summary

Avoid issues with `now` ahead of current time when using natural and relative language flags.

When calculating the local time `now` and submitting it to the Lacework APIs, the clock on some
computers could have an offset (they could be ahead of time) with the Lacework platform, we have
seen a max of 1-2s offset so, we have decided to default `now` to `-2s` to avoid these issues.

We have also made this clock offset configurable, users can use the environment variable `LW_CLOCK_OFFSET`
to customize this parameter or put it back to the previous default (`-0s`).

Manual
- [x] Run `lacework alert ls --range "last 3 days" ` ensure end time is adjusted.

Plus unit and integrations tests.

## Issue

https://lacework.atlassian.net/browse/LINK-1507

### DIFF
--- a/lwtime/nattime.go
+++ b/lwtime/nattime.go
@@ -188,7 +188,7 @@ func ParseNatural(n string) (time.Time, time.Time, error) {
 	// time.Now() is intentional here such that snaps work properly
 	// For instance snapping to @d should snap to the start of the local day
 	startLocal, endLocal, err := parseNaturalFromTime(n, time.Now())
-	return startLocal.UTC(), endLocal.UTC(), err
+	return startLocal.UTC(), endLocal.UTC().Add(-time.Minute), err
 }
 
 func parseNaturalFromTime(n string, fromTime time.Time) (time.Time, time.Time, error) {
--- a/lwtime/nattime.go
+++ b/lwtime/nattime.go
@@ -181,14 +181,15 @@ func (nt natural) getRelativeRange() (relStart string, relEnd string, err error)
 // Start and End time objects are returned in UTC
 //
 // start, end, err := lwtime.ParseNatural("this year")
-// if err != nil {
-// 	...
-// }
+//
+//	if err != nil {
+//		...
+//	}
 func ParseNatural(n string) (time.Time, time.Time, error) {
 	// time.Now() is intentional here such that snaps work properly
 	// For instance snapping to @d should snap to the start of the local day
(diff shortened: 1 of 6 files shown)


lwtime/nattime.go
```diff
@@ -188,7 +188,7 @@ func ParseNatural(n string) (time.Time, time.Time, error) {
 	// time.Now() is intentional here such that snaps work properly
 	// For instance snapping to @d should snap to the start of the local day
 	startLocal, endLocal, err := parseNaturalFromTime(n, time.Now())
-	return startLocal.UTC(), endLocal.UTC(), err
+	return startLocal.UTC(), endLocal.UTC().Add(-time.Minute), err
```
afiune, line 191: I think that this will remove one minute to all Natural Times, what we need is actually
modifying only the snap `now` to maybe remove 30 seconds.

@hazedav what do you think?

dmurray-lacework: From the ticket 
``` 
Command used:
lacework alert ls --range "last 3 days" --severity info --status Open --json | ConvertFrom-Json
```

lwtime/nattime.go
```diff
@@ -181,9 +181,10 @@ func (nt natural) getRelativeRange() (relStart string, relEnd string, err error)
 // Start and End time objects are returned in UTC
 //
 // start, end, err := lwtime.ParseNatural("this year")
-// if err != nil {
-// 	...
-// }
+//
```
hazedav, line 184: not sure why vscode insists on auto formatting this.  will fix.

hazedav:
@dmurray-lacework / @afiune ,  Ready for your consideration...

afiune:
@hazedav Thank you so much for your help!!!

hazedav:
@afiune , another thing that I had considered was adding a distinct relative time specifier `safenow`.  This would allow `now` to remain at `+0s` and then terraform and CLI consumers could use `safenow` instead.  This change is a bit more involved since it requires that we update the consumers...

afiune:
@hazedav I couldn't agree more with you. 

I will put a ticket in our backlog to revisit this approach, for now, I am changing `now` to be `-2s` and make it
configurable with an environment variable.

afiune:
https://lacework.atlassian.net/browse/GROW-1521

afiune (approved):

PR #1230, opened 2023-04-12
fix(cli): allow multiple recommendation ID types

afiune:
## Summary

We currently have three different formats (maybe more) of Compliance recommendation IDs.

* `"REC_ID": "AWS_CIS_1_3"`
* `"REC_ID": "1.10"`
* `"REC_ID": "lacework-global-76"`

If a user tries to access the resources of the second and third format, the CLI will show the following error:

```
$ lacework compliance aws get-report <AccountID> --details 1.10

ERROR
'1.10' is not a valid recommendation id
```

This change fixes that problem by removing the regex check.

Added tests and ran a few commands. All worked out well.

## Issue

Jira https://lacework.atlassian.net/browse/RAIN-45227

### DIFF
--- a/api/reports.go
+++ b/api/reports.go
@@ -110,7 +110,7 @@ func (r *RecommendationV2) SeverityString() string {
 }
 
 type CloudComplianceReportV2 interface {
-	GetComplianceRecommendation(recommendationID string) RecommendationV2
+	GetComplianceRecommendation(recommendationID string) (*RecommendationV2, bool)
 }
 
 // ValidComplianceStatus is a list of all valid compliance status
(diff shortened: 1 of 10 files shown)


dmurray-lacework (commented):
Could remove this test now `TestRecommendationIDRegex`

dmurray-lacework:
Fixed failing test.  STDOUT should not be empty, it prints the cmd help text.

```
=== FAIL: integration TestComplianceAwsGetReportRecommendationIDNotFound (0.85s)                                                                                                                                                 compliance_aws_test.go:191:                                                                                                                                                                                                  
                Error Trace:    /codefresh/volume/go-sdk/integration/compliance_aws_test.go:191                                                                                                                                                                                                                                                                                                                                                   
                Test:           TestComplianceAwsGetReportRecommendationIDNotFound                                                                                                                                               
                Messages:       STDOUT should be empty  
 ```

PR #1231, opened 2023-04-13
fix: unsupported aws region

jon-stewart:
## Summary

Not all AWS region are supported.  The `cn` region is not supported by Lacework ingestion.

## Issue

https://github.com/lacework/services/pull/12158


djmctavish (approved):

afiune (approved):
The pipeline is being silly....

PR #1237, opened 2023-04-17
fix: increase default TLS Handshake Timeout to 63s

dmurray-lacework:
## Summary

Increase TLS timeout. This change fixes for https://lacework.atlassian.net/browse/LINK-1101

## Issue

https://lacework.atlassian.net/browse/LINK-1101

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -172,6 +172,17 @@ func WithTimeout(timeout time.Duration) Option {
 	})
 }
 
+// WithTransport changes the default transport to increase TLSHandshakeTimeout
+func WithTransport() Option {
+	transport := &http.Transport{
+		TLSHandshakeTimeout: 60 * time.Second,
+	}
+	return clientFunc(func(c *Client) error {
+		c.c.Transport = transport
+		return nil
+	})
+}
+
 // WithURL sets the base URL, this options is only available for test purposes
 func WithURL(baseURL string) Option {
 	return clientFunc(func(c *Client) error {
--- a/api/client.go
+++ b/api/client.go
@@ -173,10 +173,7 @@ func WithTimeout(timeout time.Duration) Option {
 }
 
 // WithTransport changes the default transport to increase TLSHandshakeTimeout
-func WithTransport() Option {
-	transport := &http.Transport{
-		TLSHandshakeTimeout: 60 * time.Second,
-	}
+func WithTransport(transport *http.Transport) Option {
 	return clientFunc(func(c *Client) error {
 		c.c.Transport = transport
 		return nil
--- a/api/client.go
+++ b/api/client.go
@@ -32,7 +32,10 @@ import (
 	"go.uber.org/zap"
 )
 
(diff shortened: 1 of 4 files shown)


cli/cmd/cli_state.go
```diff
@@ -235,6 +235,10 @@ func (c *cliState) NewClient() error {
 		api.WithHeader("User-Agent", fmt.Sprintf("Command-Line/%s", Version)),
 	}
 
+	if os.Getenv("LW_TRANSPORT_OVERRIDE") != "" {
```
afiune, line 238: Any implication of using it by default?

afiune (commented):
can we add any sort of tests?

api/client.go
```diff
@@ -172,6 +172,17 @@ func WithTimeout(timeout time.Duration) Option {
 	})
 }
 
+// WithTransport changes the default transport to increase TLSHandshakeTimeout
+func WithTransport() Option {
+	transport := &http.Transport{
+		TLSHandshakeTimeout: 60 * time.Second,
```
afiune, line 178: should we increase this timeout by default? I would say so but I don't know why we wouldn't want that? 🤔

dmurray-lacework:
> can we add any sort of tests?

It might be possible to add automation to simulate the failure. There are manual examples on the LINK comments, but i'm not sure if it would work on codefresh

cli/cmd/cli_state.go
dmurray-lacework: I don't know, perhaps it would be best left as an optional configuration. We could input the value as the time in seconds.

djmctavish (approved):
Thought the numeric answer to the universe and everything was 42, seems like the new one is 63 😀

cli/cmd/cli_state.go
```diff
@@ -231,6 +231,7 @@ func (c *cliState) NewClient() error {
 		api.WithLogLevel(c.Log.Level().CapitalString()),
 		api.WithSubaccount(c.Subaccount),
 		api.WithApiKeys(c.KeyID, c.Secret),
+		api.WithTransport(&http.Transport{TLSHandshakeTimeout: time.Second * 63}),
```
afiune, line 234: Could we move the default to the SDK level, that is, inside the `api` package instead of setting defaults for CLI, Terraform, etc.

internal/lacework/server.go
```diff
@@ -57,6 +57,15 @@ func MockServer() *Mock {
 	}
 }
 
+func MockUnstartedServer() *Mock {
+	mux := http.NewServeMux()
+	return &Mock{
+		Mux:        mux,
+		Server:     httptest.NewUnstartedServer(mux),
+		ApiVersion: "v1",
```
afiune, line 65: nit: we should stop using v1 even on tests 😅

dmurray-lacework: fixed

afiune (approved):

PR #1242, opened 2023-04-20
fix: vuln host scan-pkg-manifest fail_on_severity

dmurray-lacework:
## Summary

The `VulnerabilityCounts()` used in the fail_on_severity evaluation are using the unfiltered results, eg. when vuln status is 'GOOD'. This change passes the response as a pointer so results are filtered when VulnerabilityCounts() is called.

Tests
- [x] `TestHostScanPackagesFailOnSeverity`

Manual
Run `lacework vuln host scan-pkg-manifest` against manifest from amzn2-ami-hvm-2.0.20230320.0-x86_64-gp2 
* see manifest.json on LINK-1587 attachments

## Issue

https://lacework.atlassian.net/browse/LINK-1587

### DIFF
--- a/cli/cmd/vuln_host_scan_package_manifest.go
+++ b/cli/cmd/vuln_host_scan_package_manifest.go
@@ -134,7 +134,7 @@ To generate a package-manifest from the local host and scan it automatically:
 				return errors.Wrap(err, "unable to request an on-demand host vulnerability scan")
 			}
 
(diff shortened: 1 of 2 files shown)


djmctavish (approved):
Built this and ran command against the manifest in the ticket, can confirm no error in CLI

afiune (approved):

PR #1246, opened 2023-04-21
fix: readd doc gen file

dmurray-lacework:
## Summary

Readd the removed doc gen file.

## Issue

https://lacework.atlassian.net/browse/GROW-1522

### DIFF
--- a/cli/cmd/docs.go
+++ b/cli/cmd/docs.go
@@ -1,4 +1,4 @@
-//go:generate go run ../docs/main.go
+//go:generate go run ../../scripts/docs/doc_gen.go
 //
 // Author:: Salim Afiune Maya (<afiune@lacework.net>)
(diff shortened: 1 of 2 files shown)


djmctavish (approved):
Are we going to have the docs in the scripts directory, should we create a new one?

dmurray-lacework:
> Are we going to have the doc gen in the scripts directory, should we create a new one?

I've no issues with it living in scripts directory. But if we want a new dir for it I don't mind moving it.

djmctavish:
> > Are we going to have the doc gen in the scripts directory, should we create a new one?
> 
> I've no issues with it living in scripts directory. But if we want a new dir for it I don't mind moving it.

Not a burning need, just we might want to change that at a later date, more important that we have the docs generated!!

PR #1257, opened 2023-05-03
fix: ctr list-assessments not fetching latest results

dmurray-lacework:
## Summary

Issue in the `treeCtrVuln.ParseData` func, the slice of ctrVuln was not being updated by:
``` 
if ctr.StartTime.After(latestContainer.StartTime) {
  latestContainer.StartTime = ctr.StartTime
  latestContainer.EvalGUID = ctr.EvalGUID
}
```

This change adds a new func `treeCtrVuln.Replace` to correctly update the slice with new value. And ensure latest results are returned.

Manual
- [x] Run `lacework vuln ctr list-assessments` and ensure last scan field is showing the latest result

Test
- [x] TestTreeCtrVulnParseData

## Issue

https://lacework.atlassian.net/browse/LINK-1614

### DIFF
--- a/cli/cmd/vuln_container_list_assessments.go
+++ b/cli/cmd/vuln_container_list_assessments.go
@@ -257,7 +257,7 @@ func listVulnCtrAssessments(
 		// if not, then we need to fetch information from 1) all
 		// container registries and 2) local scanners in two separate
 		// searches since platform scanners might have way too much
-		// data which may cause loosing the local scanners data
+		// data which may cause losing the local scanners data
 		//
 		// find all container registries
 		// cli.StartProgress("Fetching container registries...")
@@ -309,15 +309,17 @@ func listVulnCtrAssessments(
 		treeOfContainerVuln.ParseData(response.Data)
 	}
 
-	cli.Log.Infow("evaluation guids", "count", len(treeOfContainerVuln.ListEvalGuid()))
-	if len(treeOfContainerVuln.ListEvalGuid()) != 0 {
+	evalGuids := treeOfContainerVuln.ListEvalGuid()
+
+	cli.Log.Infow("evaluation guids", "count", len(evalGuids))
+	if len(evalGuids) != 0 {
 		// Update the filter with the list of evaluation GUIDs and remove the "returns"
 		filter.Returns = nil
 		filter.Filters = []api.Filter{
 			{
 				Expression: "in",
 				Field:      "evalGuid",
-				Values:     treeOfContainerVuln.ListEvalGuid(),
+				Values:     evalGuids,
 			},
 		}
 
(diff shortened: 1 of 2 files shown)


cli/cmd/vuln_container_list_assessments.go
```diff
@@ -309,15 +309,17 @@ func listVulnCtrAssessments(
 		treeOfContainerVuln.ParseData(response.Data)
 	}
 
-	cli.Log.Infow("evaluation guids", "count", len(treeOfContainerVuln.ListEvalGuid()))
-	if len(treeOfContainerVuln.ListEvalGuid()) != 0 {
+	evalGuids := treeOfContainerVuln.ListEvalGuid()
```
dmurray-lacework, line 312: This is not part of the fix, but ListEvalGuid() was previously being called 3 times below.

dmurray-lacework, line 312: This is not part of the fix, but ListEvalGuid() was previously being called 3 times below.

djmctavish (approved):
I downloaded this branch and ran the command `lacework vuln ctr list-assessments` I can see that the latest results are returned!

cli/cmd/vuln_container_list_assessments.go
```diff
@@ -370,13 +372,16 @@ func (v *treeCtrVuln) ParseData(data []api.VulnerabilityContainer) {
 	for _, ctr := range data {
 		latestContainer, exist := v.Get(ctr.ImageID)
 		if exist {
-			if latestContainer.EvalGUID == ctr.EvalGUID {
+			if latestContainer.EvalGUID == ctr.EvalGUID && ctr.StartTime.Before(latestContainer.StartTime) {
```
afiune, line 375: why do we need this extra evaluation? it feels redundant 🤔

dmurray-lacework: It's not redundant with the current order of the code. We don't want the same evalID to fall into the continue if it's a later start time. I could restructure this function, but I'd rather make minimal changes for this fix.

cli/cmd/vuln_container_list_assessments.go
```diff
@@ -385,6 +390,18 @@ func (v *treeCtrVuln) ParseData(data []api.VulnerabilityContainer) {
 	}
 }
 
+// Replace updates an existing ctrVuln in the treeCtrVuln slice with a new ctrVuln
+func (v *treeCtrVuln) Replace(old ctrVuln, new ctrVuln) {
+	current := *v
```
afiune, line 395: If you have don't really need to use the pointer of `treeCtrVuln`, maybe switch this function to use the actual value so you can access and modify it with out doing `current := *v` and `*v = current`

cli/cmd/vuln_container_list_assessments.go
```diff
@@ -385,6 +390,18 @@ func (v *treeCtrVuln) ParseData(data []api.VulnerabilityContainer) {
...
+	for i, ctrVuln := range current {
+		if ctrVuln.EvalGUID == old.EvalGUID {
+			current[i] = new
+			break
+		}
+	}
+	*v = current
+}
```
afiune, line 399: This is what I think you could do instead:
```suggestion
func (v treeCtrVuln) Replace(old ctrVuln, new ctrVuln) {
	for i, ctrVuln := range v {
		if ctrVuln.EvalGUID == old.EvalGUID {
			v[i] = new
			break
		}
	}
}
```

cli/cmd/vuln_container_list_assessments.go
```diff
@@ -370,13 +372,12 @@ func (v *treeCtrVuln) ParseData(data []api.VulnerabilityContainer) {
 	for _, ctr := range data {
 		latestContainer, exist := v.Get(ctr.ImageID)
 		if exist {
-			if latestContainer.EvalGUID == ctr.EvalGUID {
```
afiune, line 373: This is still an important check, please add it back.

afiune (changes requested):

PR #1279, opened 2023-05-17
fix(cli): Inconsistent results from host vuln show

afiune:
## Summary

The new APIv2 for host vulnerabilities introduced the concept (exposed) of evaluation GUIDs,
this id is used to group all vulnerabilities from a single evaluation (assessment).

Before this change, we were ignoring this field and therefore, we could end up with two problems:
1. Display more vulnerabilities from other evaluations
2. Display less vulnerabilities since we were NOT reading all pages and, we were looking only to a
single day (`-24h`) of data

This PR fixes a number of issues by:
* It now parses the `EvalGUID` in the `VulnerabilityHost` struct
* The `VulnerabilityCounts()` func now removes duplicates before counting
* The CLI cmd now searching for latest host evaluation first
* Then, using the `EvalGUID` the CLI fetches vulnerabilities that match
* Finally, we look at the past 7 days instead of last 24 hours

A bunch of manual tests.

## Issue
https://lacework.atlassian.net/browse/GROW-1624

### DIFF
--- a/api/v2_vulnerabilities.go
+++ b/api/v2_vulnerabilities.go
@@ -631,7 +631,7 @@ func (hosts *VulnerabilitiesHostResponse) VulnerabilityCounts() HostVulnCounts {
 
 	for _, h := range hosts.Data {
 		// avoid counting duplicates
-		if h.VulnID == "" || array.ContainsStr(cves, h.VulnID) {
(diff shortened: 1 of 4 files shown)


dmurray-lacework:
Some failing unit tests 
`TestBuildCSVVulnHostReportsWithVulnerabilitiesPackagesViewAndNoFilters`

PR #1284, opened 2023-05-22
feat(api): Allow API client to be configured with a retrying policy

edoardopirovano:
## Summary

Some users of the Go SDK may want failed API requests to be retried. This PR adds a new configuration option to the API client to allow this to be configured. The default is still to not do retries, to preserve the current behaviour.

A new automated test has been added with setups up a mock server that fails a request a couple of times before succeeding it. We then try it with no retrying and confirm this fails, and with some retrying to confirm it succeeds.

### DIFF
--- a/api/client.go
+++ b/api/client.go
@@ -29,6 +29,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
 	"github.com/lacework/go-sdk/lwdomain"
 	"github.com/pkg/errors"
 	"go.uber.org/zap"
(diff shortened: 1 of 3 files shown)


linpengzhang (approved):
great!! lgtm

linpengzhang:
@edoardopirovano should we use the more general interface`*backoff.BackOff`instead of `*backoff.ExponentialBackOff`? This would allow to use other functions such as https://pkg.go.dev/github.com/cenkalti/backoff/v4#WithMaxRetries when passing a backoff, or other kind of backoffs as well

edoardopirovano:
> @edoardopirovano should we use the more general interface`*backoff.BackOff`instead of `*backoff.ExponentialBackOff`?

Hmm, _maybe_. I did consider this, but I kind of wanted to force users to use an exponential back-off because IMO this is best practice. It is annoying that means we can't use `WithMaxRetries`, but with the right choice of constants in an exponential back-off (and a bit of maths 😉) you can still get the right bound on the maximum number of retries.

PR #1299, opened 2023-06-27
chore(lwgenerate): deprecate force_destroy flags

afiune:
## Summary

All Terraform modules now will default to force destroy buckets, this PR removes all force
destroy flags, if users does not wish to allow these buckets to be destroyed when they are
not empty, users should switch off the variable at the Terraform code. (flags depend on
each module)

Updated tests and ran generate commands manually.

## Issue
https://lacework.atlassian.net/browse/GROW-1336


afiune (commented):
We need to first, release all Terraform modules. Then, update the CLI to use those new modules (pin the version)

lwgenerate/azure/test-data/activity-log-with-all-subscriptions.tf
```diff
@@ -22,7 +22,7 @@ module "az_ad_application" {
 
 module "az_activity_log" {
   source                      = "lacework/activity-log/azure"
-  version                     = "~> 2.0"
+  version                     = "~> 2.7"
```
dmurray-lacework, line 25: Have we updated these versions correctly?

https://g.codefresh.io/build/649c04a31af33f4ceb64e0eb?step=integration&tab=output&logs=text

PR #1300, opened 2023-06-27
refactor(api): remove access to API v1 + default to v2

afiune:
## Summary

We have already removed all API v1 endpoints but one, this PR is switching the auth access token
endpoint to be API v2 endpoint. Additionally, we set the API client version to `v2` and remove
completely `WithApiV2()` option function.

**BREAKING CHANGE**: Anyone using `WithApiV2()` will have to remove the function since v2
is now the default.

Updated all tests, built and run CLI manually.

## Issue

https://lacework.atlassian.net/browse/GROW-2277

### DIFF
--- a/api/_examples/agent-access-token/main.go
+++ b/api/_examples/agent-access-token/main.go
@@ -9,7 +9,7 @@ import (
 )
 
 func main() {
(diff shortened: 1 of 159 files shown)


jon-stewart (approved):

afiune:
If we keep `WithApiV2()` option as a no-op. The breaking change could be avoided.

PR #1302, opened 2023-06-28
feat: vuln host show-assessment filter by collector type

dmurray-lacework:
## Summary

- Add new `collector_type` flag to `lacework vuln host show-assessement` cmd. To allow filtering by Agent/Agentless results
- Default `collector_type` is `Agentless`
- If no flag is set by the user and agentless results are not found, we automatically try agent
- If agentless flag is explicitly set by the user, and no agentless integration is found return error and link to docs.

Display collector_type in host details output

Manully run:

- [x] `lacework vuln host show <> `
- [x] `lacework vuln host show <> --collector_type Agent`
- [x] `lacework vuln host show <> --collector_type Agentless`

## Issue

https://lacework.atlassian.net/browse/GROW-2294

### DIFF
--- a/api/entities_machines.go
+++ b/api/entities_machines.go
@@ -1,0 +1,161 @@
+//
+// Author:: Darren Murray (<darren.murray@lacework.net>)
+// Copyright:: Copyright 2023, Lacework Inc.
+// License:: Apache License, Version 2.0
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+
+package api
+
+import (
+	"time"
+)
+
+// ListMachines returns a list of MachineDetailEntity from the last 7 days
+func (svc *EntitiesService) ListMachines() (response MachinesEntityResponse, err error) {
+	now := time.Now().UTC()
+	before := now.AddDate(0, 0, -7) // 7 days from ago
+	err = svc.Search(&response,
+		SearchFilter{
+			TimeFilter: &TimeFilter{
+				StartTime: &before,
+				EndTime:   &now,
+			},
+		},
+	)
+	return
+}
+
+// ListMachinesWithFilters returns a list of UserEntity based on a user defined filter
+func (svc *EntitiesService) ListMachinesWithFilters(filters SearchFilter) (response MachinesEntityResponse, err error) {
+	err = svc.Search(&response, filters)
+	return
+}
+
+// ListAllMachines iterates over all pages to return all machine details at once
+func (svc *EntitiesService) ListAllMachines() (response MachinesEntityResponse, err error) {
+	response, err = svc.ListMachines()
+	if err != nil {
+		return
+	}
+
+	var (
+		all    []MachineDetailEntity
+		pageOk bool
+	)
+	for {
+		all = append(all, response.Data...)
+
+		pageOk, err = svc.client.NextPage(&response)
+		if err == nil && pageOk {
+			continue
+		}
+		break
+	}
+
+	response.ResetPaging()
+	response.Data = all
+	return
+}
+
+// ListAllMachinesWithFilters iterates over all pages to return all machine details at once based on a user defined filter
+func (svc *EntitiesService) ListAllMachinesWithFilters(filters SearchFilter) (response MachinesEntityResponse, err error) {
+	response, err = svc.ListMachinesWithFilters(filters)
+	if err != nil {
+		return
+	}
+
+	var (
+		all    []MachineDetailEntity
+		pageOk bool
+	)
+
+	for {
+		all = append(all, response.Data...)
+
+		pageOk, err = svc.client.NextPage(&response)
+		if err == nil && pageOk {
+			continue
+		}
+		break
+	}
+
+	response.ResetPaging()
+	response.Data = all
+	return
+}
+
+type MachinesEntityResponse struct {
+	Data   []MachineDetailEntity `json:"data"`
+	Paging V2Pagination          `json:"paging"`
+
+	v2PageMetadata `json:"-"`
+}
+
+// Fulfill Pageable interface (look at api/v2.go)
+func (r MachinesEntityResponse) PageInfo() *V2Pagination {
+	return &r.Paging
+}
+func (r *MachinesEntityResponse) ResetPaging() {
+	r.Paging = V2Pagination{}
+	r.Data = nil
+}
+
+type MachineEntity struct {
+	AwsInstanceID string    `json:"awsInstanceId"`
+	Hostname      string    `json:"hostname"`
+	EntityType    string    `json:"entityType"`
+	EndTime       time.Time `json:"endTime"`
+	Mid           int       `json:"mid"`
+	PrimaryIpAddr string    `json:"primaryIpAddr"`
+	StartTime     time.Time `json:"startTime"`
+	Tags          struct {
+		// Shared Tags
+		Cluster             string `json:"Cluster,omitempty"`
+		Env                 string `json:"Env,omitempty"`
+		Arch                string `json:"arch,omitempty"`
+		ExternalIP          string `json:"ExternalIp,omitempty"`
+		Hostname            string `json:"Hostname,omitempty"`
+		InstanceID          string `json:"InstanceId,omitempty"`
+		InternalIP          string `json:"InternalIp,omitempty"`
+		LwTokenShort        string `json:"LwTokenShort,omitempty"`
+		Os                  string `json:"os,omitempty"`
+		VMInstanceType      string `json:"VmInstanceType,omitempty"`
+		VMProvider          string `json:"VmProvider,omitempty"`
+		Zone                string `json:"Zone,omitempty"`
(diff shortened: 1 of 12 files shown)


cli/cmd/vuln_host.go
```diff
@@ -101,6 +101,12 @@ func init() {
 		"local", "l", false,
 		"automatically generate the package manifest from the local host",
 	)
+
+	// the collector_type of the assessment
+	vulHostShowAssessmentCmd.Flags().StringVar(&vulCmdState.CollectorType,
+		"collector_type", vulnHostCollectorTypeAgentless,
+		"filter assessments by collector type(Agent/Agentless)",
```
afiune, line 108:
```suggestion
		"filter assessments by collector type (Agent or Agentless)",
```

cli/cmd/vuln_host_show_assessment.go
```diff
@@ -58,6 +58,22 @@ Grab a CVE id and feed it to the command:
 					vulCmdState.Details = true
 				}
 			}
+
+			// validate collector_type flag
+			if vulCmdState.CollectorType != vulnHostCollectorTypeAgentless && vulCmdState.CollectorType != vulnHostCollectorTypeAgent {
+				return errors.Errorf("collector_type must be either %s or %s", vulnHostCollectorTypeAgent, vulnHostCollectorTypeAgentless)
+			}
```
afiune, line 65: nit; could be a switch statement to avoid nested ifs 
```suggestion
			switch vulCmdState.CollectorType {
			case vulnHostCollectorTypeAgentless, vulnHostCollectorTypeAgent:
			default:
				return errors.Errorf(
					"collector_type must be either %s or %s",
					vulnHostCollectorTypeAgent, vulnHostCollectorTypeAgentless,
				)
			}
```

cli/cmd/vuln_host_test.go
```diff
@@ -421,7 +428,8 @@ func mockHostVulnerabilityAssessment() api.VulnerabilitiesHostResponse {
             "evalCtx": {
                 "exception_props": [],
                 "Hostname": "host-mock",
-                "mc_eval_guid": "mock"
+                "mc_eval_guid": "mock",
+				"collector_type": "Agent"
```
afiune, line 432:
```suggestion
                "collector_type": "Agent"
```

cli/cmd/vuln_host_show_assessment.go
```diff
@@ -559,3 +595,16 @@ func sortVulnHosts(slice []api.VulnerabilityHost) {
...
 	}
 }
+
+func checkAgentlessCloudAccount() bool {
+	resp, err := cli.LwApi.V2.CloudAccounts.List()
+	if err != nil {
+		return false
+	}
+	for _, ca := range resp.Data {
+		if strings.Contains(ca.Type, "Sidekick") {
```
afiune, line 605: If we actually want to check for an integration, let's make sure to use the constant defined at the `api` package instead of hardcoding a string here. (not that the integrations type will change but, that's why we define constants, right? 😅 )

cli/cmd/vuln_host_show_assessment.go
```diff
@@ -58,6 +58,22 @@ Grab a CVE id and feed it to the command:
...
 			}
+
+			// validate collector_type flag
+			if vulCmdState.CollectorType != vulnHostCollectorTypeAgentless && vulCmdState.CollectorType != vulnHostCollectorTypeAgent {
+				return errors.Errorf("collector_type must be either %s or %s", vulnHostCollectorTypeAgent, vulnHostCollectorTypeAgentless)
+			}
+
+			if vulCmdState.CollectorType == vulnHostCollectorTypeAgentless {
+				if !checkAgentlessCloudAccount() {
+					if cmd.Flags().Changed("collector_type") {
```
afiune, line 69: This logic tricked me for a few.

We will throw an error message only, and only if
1. The user does NOT have an Agentless cloud-account integration (AND)
2. The user used the flag `--collector_type` (AND)
3. ONLY if the provided collector type is `Agentless`, which is already the default 😕, which then I thought the flag never changed..... 😖  

We might wanna find a better way to do this check.

Finally, we are already interacting with the APIs here so, maybe displaying a Spinner about the task that the CLI is doing, would be nice for folks with slow internet.

cli/cmd/vuln_host_show_assessment.go
```diff
@@ -121,6 +137,13 @@ Grab a CVE id and feed it to the command:
...
 
+				// filter by collector_type
+				filter.Filters = append(filter.Filters, api.Filter{
+					Expression: "eq",
+					Field:      "evalCtx.collector_type",
+					Value:      vulCmdState.CollectorType,
+				})
+
 				cli.StartProgress(
 					fmt.Sprintf("Fetching vulnerabilities from host evaluation '%s'...", evalGUID),
```
afiune, line 148: At this point we know what the collector type is, can we add it in the spinner like
```suggestion
					fmt.Sprintf("Fetching vulnerabilities from host evaluation '%s' (collector_type: %s) ...", evalGUID),
```

afiune: Also, there is another line above with an info message:
```go
				cli.Log.Infow("latest assessment found", "eval_guid", evalGUID)
```

could we change it to
```go
				cli.Log.Infow("latest assessment found", "eval_guid", evalGUID, "collector_type", vulCmdState.CollectorType)
```

afiune (changes requested):
We must add integration tests for both. Agent and Agentless datasets

cli/cmd/vuln_host_show_assessment.go
dmurray-lacework: There is no nested if on the line with this suggestion.

cli/cmd/vuln_host_show_assessment.go
dmurray-lacework: An explanation of the behaviour:

If the user explicitly wants agentless results eg. they've set the flag. But they've not configured an agentless integration then display an error rather than agent results. 

If no flag is set, and there is no agentless integration then we just default to agent

dmurray-lacework: I'll add comments around this to make it a little more clear what the logic is.

afiune:
tests are failing

dmurray-lacework:
This needs more work. We need to take a look at the entities/MachineDetails search in cli/cmd/vuln_host_show_assessment.go:103

api/entities_machines.go
```diff
@@ -0,0 +1,161 @@
...
+//
+
+package api
+
+import (
+	"time"
+)
+
+// ListMachines returns a list of MachineDetailEntity from the last 7 days
+func (svc *EntitiesService) ListMachines() (response MachinesEntityResponse, err error) {
```
afiune, line 26: All this file needs unit testing, please add it on a new PR.

api/entities_machines.go
```diff
@@ -0,0 +1,161 @@
...
+// limitations under the License.
+//
+
+package api
+
+import (
+	"time"
+)
+
+// ListMachines returns a list of MachineDetailEntity from the last 7 days
```
afiune, line 25:
```suggestion
// ListMachines returns a list of MachineEntity from the last 7 days
```

api/entities_machines.go
```diff
@@ -0,0 +1,161 @@
...
+
+// ListAllMachines iterates over all pages to return all machine details at once
+func (svc *EntitiesService) ListAllMachines() (response MachinesEntityResponse, err error) {
+	response, err = svc.ListMachines()
+	if err != nil {
+		return
+	}
+
+	var (
+		all    []MachineDetailEntity
```
afiune, line 54:
```suggestion
		all    []MachineEntity
```

api/entities_machines.go
```diff
@@ -0,0 +1,161 @@
...
+		break
+	}
+
+	response.ResetPaging()
+	response.Data = all
+	return
+}
+
+type MachinesEntityResponse struct {
+	Data   []MachineDetailEntity `json:"data"`
```
afiune, line 100:
```suggestion
	Data   []MachineEntity `json:"data"`
```

cli/cmd/vuln_host_show_assessment.go
```diff
@@ -121,8 +140,15 @@ Grab a CVE id and feed it to the command:
...
+				// filter by collector_type
+				filter.Filters = append(filter.Filters, api.Filter{
+					Expression: "eq",
+					Field:      "evalCtx.collector_type",
+					Value:      vulCmdState.CollectorType,
+				})
+
 				cli.StartProgress(
-					fmt.Sprintf("Fetching vulnerabilities from host evaluation '%s'...", evalGUID),
+					fmt.Sprintf("Fetching vulnerabilities from host evaluation '%s' (collector_type: %s) ...", evalGUID, vulCmdState.CollectorType),
```
afiune, line 151:
```suggestion
					fmt.Sprintf("Fetching vulnerabilities from host evaluation '%s' (collector_type: %s) ...",
						evalGUID, vulCmdState.CollectorType),
```

cli/cmd/vuln_host_test.go
afiune:
```suggestion
                "collector_type": "Agent"
```

internal/array/contains.go
```diff
@@ -40,7 +40,7 @@ func ContainsStrCaseInsensitive(array []string, expected string) bool {
 
 func ContainsPartialStr(array []string, expected string) bool {
 	for _, value := range array {
-		if strings.Contains(value, expected) {
+		if strings.Contains(expected, value) {
```
afiune, line 43: this is a big change for this function, It was never used but also, there is no unit tests so it is hard to tell if this is good or bad.

afiune (approved):

cli/cmd/vuln_host_test.go
afiune, line 432:
```suggestion
                " collector_type": "Agent"
```

PR #1309, opened 2023-06-30
feat(COD-1255): Enhancements to how telemetry can be used

edoardopirovano:
## Summary

Over in the Code Security team we want to add some observability in environments where we have the Lacework CLI and are running a bunch of command in sequence as well as doing some things outside of the CLI. We would like to land the two enhancements in this PR to support this:

- We now read a variable `LACEWORK_CONTEXT_ID` and report this. This will allow us to link up different invocations of the CLI into one unique context for the whole sequence of commands.
- We now support a hidden command `lacework telemetry upload` which accepts a JSON blob of telemetry to upload. We'll use this to report additional telemetry from the bits of our code running outside the CLI.

- One automated test was modified to check the new behaviour, and a new one was also added.
- I did manual end-to-end testing locally and observed data arriving to Honeycomb ([here](https://ui.honeycomb.io/lacework/datasets/lacework-cli-dev/result/da4Kaqvoyav?tab=raw)).

## Issue

https://lacework.atlassian.net/browse/COD-1255

### DIFF
--- a/cli/cmd/honeyvent.go
+++ b/cli/cmd/honeyvent.go
@@ -115,9 +115,10 @@ type Honeyvent struct {
 
 	// tracing data for multiple events, this is useful for specific features
 	// within the Lacework CLI such as daily version check, polling mechanism, etc.
-	TraceID  string `json:"trace.trace_id,omitempty"`
-	SpanID   string `json:"trace.span_id,omitempty"`
-	ParentID string `json:"trace.parent_id,omitempty"`
+	TraceID   string `json:"trace.trace_id,omitempty"`
+	SpanID    string `json:"trace.span_id,omitempty"`
+	ParentID  string `json:"trace.parent_id,omitempty"`
+	ContextID string `json:"trace.context_id,omitempty"`
 }
 
 // InitHoneyvent initialize honeycomb library and main Honeyvent, such event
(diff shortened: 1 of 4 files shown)


linpengzhang (approved):
Great! My only worry is whether sending separate trace_spans would not allow to correlate things about a component and this new telemetry in a single query. For instance, in the case of `sca`, we may not be able to have a dashboard with this new telemetry  grouped by `remote` information that `sca` is able to find

edoardopirovano:
@linpengzhang Hmm, that's interesting. Do you think it would make sense to have the env variable overrride the `trace_id` instead so that all the CLI invocations can end up as part of the same trace?

linpengzhang:
> @linpengzhang Hmm, that's interesting. Do you think it would make sense to have the env variable overrride the `trace_id` instead so that all the CLI invocations can end up as part of the same trace?

I think as long as the data is in two different rows, they can't be queried altogether. I doubt, but it might be that by overriding `trace_id`  and `trace_span`, Honeycomb aggregates the data

edoardopirovano:
> I think as long as the data is in two different rows, they can't be queried altogether. I doubt, but it might be that by overriding `trace_id` and `trace_span`, Honeycomb aggregates data sent in different instants. If that's not the case, the only solution is to send component specific data and other data all together

Hmm, okay. I think getting all the telemetry into one row is going to be too fiddly and potentially error prone so I think this new command should be fine. In most cases I reckon we probably won't need to query together but rather will want to see a single set of traces from a run that we're debugging.

PR #1312, opened 2023-07-04
feat: default compliance report to use report_name

dmurray-lacework:
## Summary

This change
- Changes the default parameters for `compliance aws|az|gcp get-report` to use `report_name` over `type`
- Fix report type not being updated in --csv output
- Mark the `--type` flag as deprecated

Report names are discoverable via `lacework report-definitions list` but types are not. This makes fetching reports by name preferable. Validation can be performed by fetching a valid list of reports from the api.  New and custom reports can be discovered.

- [x] run `compliance aws|az|gcp get-report`  without any flags
- [x] run `compliance aws|az|gcp get-report`  with --report_name flag
- [x] run `compliance aws|az|gcp get-report`  with --type flag
- [x] update compliance tests 

## Issue

https://lacework.atlassian.net/browse/LINK-1816

### DIFF
--- a/cli/cmd/compliance_aws.go
+++ b/cli/cmd/compliance_aws.go
@@ -21,7 +21,6 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
-	"log"
 	"strconv"
 	"strings"
 	"time"
@@ -40,7 +39,7 @@ var (
 	compAwsCmdState = struct {
 		Type       string
 		ReportName string
-	}{ReportName: "CIS Amazon Web Services Foundations Benchmark v1.4.0"}
+	}{ReportName: api.ComplianceReportDefaultAws}
 
 	// complianceAwsListAccountsCmd represents the list-accounts inside the aws command
 	complianceAwsListAccountsCmd = &cobra.Command{
@@ -574,14 +573,11 @@ func init() {
 valid types:%s`, prettyPrintReportTypes(api.AwsReportTypes())))
 
 	// mark report type flag as deprecated
-	err := complianceAwsGetReportCmd.Flags().MarkDeprecated("type", "use --report_name flag instead")
-	if err != nil {
-		log.Fatal("unable to deprecate flag '--type'")
-	}
+	errcheckWARN(complianceAwsGetReportCmd.Flags().MarkDeprecated("type", "use --report_name flag instead"))
 
 	// Run 'lacework report-definition --subtype AWS' for a full list of AWS report names
 	complianceAwsGetReportCmd.Flags().StringVar(&compAwsCmdState.ReportName, "report_name",
-		"CIS Amazon Web Services Foundations Benchmark v1.4.0",
+		api.ComplianceReportDefaultAws,
 		"report name to display, run 'lacework report-definitions list' for more information.")
 
 	complianceAwsGetReportCmd.Flags().StringSliceVar(&compCmdState.Category, "category", []string{},
--- a/cli/cmd/compliance_aws.go
+++ b/cli/cmd/compliance_aws.go
@@ -21,6 +21,7 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
+	"log"
 	"strconv"
 	"strings"
 	"time"
@@ -39,7 +40,7 @@ var (
 	compAwsCmdState = struct {
 		Type       string
 		ReportName string
-	}{Type: "AWS_CIS_14"}
+	}{ReportName: "CIS Amazon Web Services Foundations Benchmark v1.4.0"}
 
 	// complianceAwsListAccountsCmd represents the list-accounts inside the aws command
 	complianceAwsListAccountsCmd = &cobra.Command{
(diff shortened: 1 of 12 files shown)


jon-stewart (approved):

cli/cmd/compliance_aws.go
```diff
@@ -82,11 +83,15 @@ var (
...
 
-			if array.ContainsStr(api.AwsReportTypes(), compAwsCmdState.Type) {
-				return nil
-			} else {
-				return errors.Errorf(`supported report types are: %s'`, strings.Join(api.AwsReportTypes(), ", "))
+			if cmd.Flags().Changed("type") {
+				if array.ContainsStr(api.AwsReportTypes(), compAwsCmdState.Type) {
+					return nil
+				} else {
+					return errors.Errorf(`supported report types are: %s'`, strings.Join(api.AwsReportTypes(), ", "))
```
afiune, line 90:
```suggestion
					return errors.Errorf("supported report types are: %s", strings.Join(api.AwsReportTypes(), ", "))
```

cli/cmd/compliance_aws.go
```diff
@@ -39,7 +40,7 @@ var (
 	compAwsCmdState = struct {
 		Type       string
 		ReportName string
-	}{Type: "AWS_CIS_14"}
+	}{ReportName: "CIS Amazon Web Services Foundations Benchmark v1.4.0"}
```
afiune, line 43: yikes, this is such a long name and it is being used in multiple places, instead of hardcoding it, can we use a constant either from the `api` package, if any, or declare one here in the `cli` package

cli/cmd/compliance_aws.go
```diff
@@ -562,13 +568,20 @@ func init() {
...
-	//AWS_NIST_800-171_rev2'
+	//AWS_NIST_800-171_rev2
 	complianceAwsGetReportCmd.Flags().StringVar(&compAwsCmdState.Type, "type", "AWS_CIS_14",
 		fmt.Sprintf(`report type to display, run 'lacework report-definitions list' for more information.
 valid types:%s`, prettyPrintReportTypes(api.AwsReportTypes())))
 
+	// mark report type flag as deprecated
+	err := complianceAwsGetReportCmd.Flags().MarkDeprecated("type", "use --report_name flag instead")
+	if err != nil {
+		log.Fatal("unable to deprecate flag '--type'")
```
afiune, line 579: We do not use `log`, we have our own package `lwlogger`. use `cli.Log` instead.

Also, what are we trying to do with this log? 🤔 if you are just trying to check the error because lint tests are failing, then use `errcheckWARN()` instead.

cli/cmd/compliance_azure.go
```diff
@@ -40,7 +41,7 @@ var (
 	compAzCmdState = struct {
 		Type       string
 		ReportName string
-	}{Type: "AZURE_CIS_131"}
+	}{ReportName: "CIS Microsoft Azure Foundations Benchmark v1.5.0"}
```
afiune, line 44: same, constant please and use it in other places

cli/cmd/compliance_azure.go
```diff
@@ -139,11 +140,14 @@ Use the following command to list all Azure Tenants configured in your account:
...
-			if array.ContainsStr(api.AzureReportTypes(), compAzCmdState.Type) {
-				return nil
-			} else {
-				return errors.Errorf("supported report types are: %s", strings.Join(api.AzureReportTypes(), ", "))
+			if cmd.Flags().Changed("type") {
+				if array.ContainsStr(api.AzureReportTypes(), compAzCmdState.Type) {
+					return nil
+				} else {
+					return errors.Errorf("supported report types are: %s", strings.Join(api.AzureReportTypes(), ", "))
+				}
```
afiune, line 148: nit
```suggestion
			if cmd.Flags().Changed("type") &&
				!array.ContainsStr(api.AzureReportTypes(), compAzCmdState.Type) {
				return errors.Errorf("supported report types are: %s", strings.Join(api.AzureReportTypes(), ", "))
```

cli/cmd/compliance_azure.go
```diff
@@ -527,8 +531,15 @@ func init() {
 valid types:%s`, prettyPrintReportTypes(api.AzureReportTypes())),
 	)
 
+	// mark report type flag as deprecated
+	err := complianceAzureGetReportCmd.Flags().MarkDeprecated("type", "use --report_name flag instead")
```
afiune, line 535: same as above, search for `MarkDeprecated` in this repo for examples.

afiune (changes requested):

afiune:
This also doesn't feel like a feature, it feels more like a bug-fix.

dmurray-lacework:
> This also doesn't feel like a feature, it feels more like a bug-fix.

I'll agree there is a bug fix in this PR. But updating the default to the report_name is an enhancement not a bug fix.

afiune (approved):

PR #1321, opened 2023-07-12
fix(cli): compliance get-report blank CSV output

afiune:
## Summary

The platform today only returns a list of resources that are in violation, that means, “non-compliant” resources. We currently do not return the list of resources that are “compliant”, “requires-manual-assessment” or “could-not-assess”

- For those cases our CSV output will only show the number of resources.

Added integration tests and ran a few commands locally on my workstation.

## Issue

https://lacework.atlassian.net/browse/GROW-2316
https://lacework.atlassian.net/browse/LINK-1864

### DIFF
--- a/cli/cmd/compliance.go
+++ b/cli/cmd/compliance.go
@@ -273,6 +273,26 @@ func complianceCSVReportRecommendationsTable(details *complianceCSVReportDetails
 						""))
 			}
 		}
+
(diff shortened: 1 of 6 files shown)


ipcrm (approved):
Pulled down and tested locally; seems like the best solution we can have for when resources aren't returned from the API.

PengyuanZhao (approved):
LGTM!

PR #1331, opened 2023-07-18
feat(api): support new fields in alert_rules api

dmurray-lacework:
## Summary

Add support for missing fields in the alert_rules api

In the EventCategory property the values Registry, SystemCall, HostVulnerability & ContainerVulnerability
The Sources property has been added with the values Agent, Aws, Azure, Gcp, K8s
The AlertCategory property has been added with the values Policy, Anomaly, Composite

- [x] Run example `api/_examples/alert-rules/main.go` with [api changes](https://github.com/lacework/rainbow/pull/13216)

## Issue

https://lacework.atlassian.net/browse/GROW-2326

### DIFF
--- a/api/_examples/alert-rules/main.go
+++ b/api/_examples/alert-rules/main.go
@@ -39,6 +39,8 @@ func main() {
 		Severities:      api.AlertRuleSeverities{api.AlertRuleSeverityHigh},
 		ResourceGroups:  []string{"TECHALLY_000000000000AAAAAAAAAAAAAAAAAAAA"},
 		EventCategories: []string{"Compliance"},
+		AlertCategories: []string{"Policy"},
+		Sources: 		 []string{"Aws"},
 	}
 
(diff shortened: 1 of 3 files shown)


api/_examples/alert-rules/main.go
```diff
@@ -39,6 +39,8 @@ func main() {
 		Severities:      api.AlertRuleSeverities{api.AlertRuleSeverityHigh},
 		ResourceGroups:  []string{"TECHALLY_000000000000AAAAAAAAAAAAAAAAAAAA"},
 		EventCategories: []string{"Compliance"},
```
djmctavish, line 41: Should we have more tests of the newly added categories etc?

djmctavish (approved):

api/_examples/alert-rules/main.go
dmurray-lacework: updated alert_rule_test.go

Integration testing for new fields in the terraform-provider integrations too.

PR #1340, opened 2023-07-31
feat(CAD-360): Support Resource Group v2 in CLI

zekisherif:
## Summary

Adds support for the CRUD operations for Resource Groups V2.

1. Updated example code for resource groups and ensured it passed.
2. Manually ran the command line. See pictures below
3. ran all the commands listed to do before publishing a PR

List Resource Groups
Screenshot 2023-08-07 at 12 17 49 PM

Create an AWS RGv2
Screenshot 2023-08-07 at 12 18 41 PM

Create an AWS RGv1
Screenshot 2023-08-07 at 12 19 08 PM

List showing the RGv2 that was created
Screenshot 2023-08-07 at 12 19 30 PM

Delete the RGv1 created earlier
Screenshot 2023-08-07 at 12 22 22 PM

Calling GET for both an V1 and V2 resource
Screenshot 2023-08-07 at 12 34 41 PM

## Issue

https://lacework.atlassian.net/browse/CAD-360

### DIFF
--- a/api/resource_groups.go
+++ b/api/resource_groups.go
@@ -269,15 +269,7 @@ func (group ResourceGroupData) Status() string {
 }
 
 func (group ResourceGroupData) IsV2Group() bool {
-	if group.Props != nil {
-		return false
-	}
-
-	if group.Query != nil {
-		return true
-	}
-
-	return false
+	return group.Query != nil 
 }
 
 type ResourceGroupResponse struct {
--- a/api/resource_groups.go
+++ b/api/resource_groups.go
@@ -22,10 +22,10 @@ import (
 	"encoding/json"
 	"fmt"
 	"strconv"
-
-	"github.com/pkg/errors"
+	"time"
 
 	"github.com/lacework/go-sdk/lwtime"
+	"github.com/pkg/errors"
 )
 
 // ResourceGroupsService is the service that interacts with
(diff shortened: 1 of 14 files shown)


dmurray-lacework:
Hey @zekisherif it would great to `update api/_examples/resource-groups/main.go` while you're making changes to see the usage.

api/v2.go
```diff
@@ -113,7 +113,7 @@ func NewV2Endpoints(c *Client) *V2Endpoints {
 		AlertRules:              &AlertRulesService{c},
 		CloudAccounts:           &CloudAccountsService{c},
 		ContainerRegistries:     &ContainerRegistriesService{c},
-		ResourceGroups:          &ResourceGroupsService{c},
+		ResourceGroups:          NewResourceGroupsVersionService(c),
```
dmurray-lacework, line 116: I'm not 100% sure this is going to using the schemaService is going to fulfil what you need with this.

For context we used this schemaService when transitioning from the v1 api when cloudAccounts, alertChannels, containerRegistries were under `Integration` resource.  Only Get and Delete are exposed in the `V2Service` interface

zekisherif: Yeah I wasn't sure about the changes I actually needed in this file, but once I get to testing out the cli locally, I think I'll understand it better.

api/resource_groups_version_service.go
```diff
@@ -0,0 +1,367 @@
...
+		if createErr != nil {
+			err = createErr
+			return
+		}
+
+		err = castResourceGroupV2Response(createResponse, &response)
+		return
+	}
+
+	err = errors.New("Unable to create resource group")
```
dmurray-lacework, line 150: Is there any more context we can put in this error. What scenario would this be returned.

In the above code, 
  if props is not nil;then we use the v1 create and return
  if v2 is enabled and the query is not nil; attempt v2 create and cast response and return;
  
This error will return if the input is doesn't have props, doesn't have have v2 flag and a query. With this condition, maybe we can return eariler.

cli/cmd/resource_group_v2.go
```diff
@@ -0,0 +1,121 @@
...
+	cli.StartProgress(" Creating resource group...")
+	_, err = cli.LwApi.V2.ResourceGroups.Create(resourceGroup)
+	cli.StopProgress()
+	return err
+}
+
+func inputRGQueryFromEditor(resourceType string) *survey.Editor {
+	prompt := &survey.Editor{
+		Message:  fmt.Sprintf("Type a query for the new %s Resource Group", resourceType),
+		FileName: "resourceGroupQuery*.json",
```
dmurray-lacework, line 84: Editor input is great, perhaps we can help the user out with a link to docs to create a query. If there are no public docs at the minute we could put a placeholder and return to it.

The survey `Help` field may be useful for this

api/resource_groups_version_service.go
zekisherif: Ok I've refactored it a bit.

cli/cmd/resource_group_v2.go
zekisherif: Will add the doc to Help

api/resource_groups.go
```diff
@@ -304,6 +268,18 @@ func (group ResourceGroupData) Status() string {
...
+func (group ResourceGroupData) IsV2Group() bool {
+	if group.Props != nil {
+		return false
+	}
+
+	if group.Query != nil {
+		return true
+	}
+
+	return false
```
dmurray-lacework, line 280: nit: could reduce this the only condition that returns true is `group.Query != nil`

```suggestion
	return group.Query != nil 
```

dmurray-lacework (approved):
A couple of comments, but approved.

We still need to take a look and why the linting build step is failing.

cli/cmd/resource_groups.go
```diff
@@ -201,18 +253,42 @@ func parsePropsType(response api.ResourceGroupData) (api.ResourceGroupProps, err
...
+		"MACHINE",
+		"AWS(v2)",
+		"AZURE(v2)",
+		"GCP(v2)",
+		"CONTAINER(v2)",
+		"MACHINE(v2)",
+	}
+
+	isRGv2Enabled := false
+	ffResponse, _ := cli.LwApi.V2.FeatureFlags.GetFeatureFlagsMatchingPrefix("PUBLIC.rgv2.cli")
```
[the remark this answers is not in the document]
zekisherif: Fixed

PR #1350, opened 2023-08-08
fix(COD-1559): Add some retrying in `updater.go`

edoardopirovano:
## Summary

We've started seeing some intermitted failures with a 403 status code in this code. This PR adds some retrying so that if this is due to rate limiting as we suspect we will succeed after retrying.

Manual testing to confirm things are still working normally. Hard to confirm if this will fix things as the issue above is intermittent.

## Issue

https://lacework.atlassian.net/browse/COD-1559

### DIFF
--- a/lwupdater/updater.go
+++ b/lwupdater/updater.go
@@ -29,6 +29,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/cenkalti/backoff/v4"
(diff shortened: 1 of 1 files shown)


linpengzhang (approved):

jeremydubreil (approved):
Thanks!

PR #1375, opened 2023-09-12
fix(cli): Failures in Lacework CLI update checking

linpengzhang:
## Summary

`lacework version` is failing non deterministically, see https://github.com/lacework/code-security-action/actions/runs/5794782714/job/15707673444

This PR fails more silently,  by just logging an error instead of exiting.

Test: we are not exiting anymore, so it should just work. No need for tests as it is an easy change.

## Issue

[
](https://lacework.atlassian.net/browse/COD-1559)

### DIFF
--- a/cli/cmd/version.go
+++ b/cli/cmd/version.go
@@ -61,13 +61,13 @@ versions available for update.
 Set the environment variable 'LW_UPDATES_DISABLE=1' to avoid checking for updates.`,
 		Args: cobra.NoArgs,
 		Run: func(_ *cobra.Command, _ []string) {
-			componentVerionsOutput := &strings.Builder{}
+			componentVersionsOutput := &strings.Builder{}
 			if cli.LwComponents != nil {
 				for _, component := range cli.LwComponents.Components {
 					if component.IsInstalled() {
 						v, err := component.CurrentVersion()
 						if err == nil {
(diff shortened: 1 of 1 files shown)


jon-stewart (approved):

edoardopirovano:
Hmm, I don't think this is enough. `lacework version` isn't the only place we check for updates: there's also `dailyVersionCheck` which gets run the first time each day that we run a CLI command (it's invoked from `root.go`). I'm pretty sure I'd observed some failures there too.

linpengzhang:
> 

I ignored those calls since I was hoping they were not failing. Let me do another commit

linpengzhang:
> > 
> 
> I ignored those calls since I was hoping they were not failing. Let me do another commit

the only call of `versionCheck` is inside ` dailyVersionCheck()` which is called only in the snipped below. It is already failing silently
@edoardopirovano 
```
			// run the daily version check but do not fail if we couldn't check
			// this is not a critical part of the CLI and we do not want to impact
			// customers workflows or CI systems
			if err := dailyVersionCheck(); err != nil {
				cli.Log.Debugw("unable to run daily version check", "error", err)
			}
```

edoardopirovano (approved):
Gotcha, thanks for checking those calls. LGTM in that case.

PR #1384, opened 2023-09-21
feat: CDK component install telemetry

jon-stewart:
## Summary

A customer found that component installation would take ~30 minutes to complete.  This change provides more CLI information and honeycomb telemetry

Local testing:
- install iac
- update iac
- uninstall iac

## Issue

https://lacework.atlassian.net/browse/GROW-2446

### DIFF
--- a/cli/cmd/component.go
+++ b/cli/cmd/component.go
@@ -411,6 +411,8 @@ func runComponentsInstall(cmd *cobra.Command, args []string) (err error) {
 
 	params["verify_duration_ms"] = time.Since(start).Milliseconds()
 
+	start = time.Now()
+
 	cli.StartProgress(fmt.Sprintf("Configuring component %s...", component.Name))
 	// component life cycle: initialize
 	stdout, stderr, errCmd := component.RunAndReturn([]string{"cdk-init"}, nil, cli.envs()...)
--- a/cli/cmd/component.go
+++ b/cli/cmd/component.go
@@ -352,13 +352,15 @@ func componentsToTable() [][]string {
 
 func runComponentsInstall(cmd *cobra.Command, args []string) (err error) {
 	var (
-		component_name string
-		version        string
-		params         map[string]interface{}
-		start          time.Time
+		componentName string                 = args[0]
+		version       string                 = versionArg
+		params        map[string]interface{} = make(map[string]interface{})
+		start         time.Time
 	)
 
-	params = make(map[string]interface{})
+	cli.Event.Component = componentName
+	cli.Event.Feature = "install_component"
+	defer cli.SendHoneyvent()
 
 	cli.StartProgress("Loading components state...")
 	// @afiune maybe move the state to the cache and fetch if it if has expired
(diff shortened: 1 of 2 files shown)


PengyuanZhao (approved):

cli/cmd/component.go
```diff
@@ -349,7 +350,16 @@ func componentsToTable() [][]string {
 	return out
 }
 
-func runComponentsInstall(_ *cobra.Command, args []string) (err error) {
+func runComponentsInstall(cmd *cobra.Command, args []string) (err error) {
+	var (
+		component_name string
```
afiune, line 355: Nit- Not the standard way to name variables
```suggestion
		componentName string
```

cli/cmd/component.go
```diff
@@ -349,7 +350,16 @@ func componentsToTable() [][]string {
...
+		version        string
+		params         map[string]interface{}
+		start          time.Time
+	)
+
+	params = make(map[string]interface{})
```
afiune, line 361: Nit- this variable was just defined, let's move this initialization up there.

cli/cmd/component.go
```diff
@@ -359,21 +369,49 @@ func runComponentsInstall(_ *cobra.Command, args []string) (err error) {
...
+		err = errors.New(fmt.Sprintf("component %s not found. Try running 'lacework component list'", component_name))
 		return
 	}
 
+	cli.OutputChecklist(successIcon, fmt.Sprintf("Component %s found\n", component_name))
+
+	if versionArg == "" {
+		version = component.LatestVersion.String()
+	} else {
+		version = versionArg
```
afiune, line 385: What do we gain having this if statement all the way here? Should we instead initialize version with the context of versionArg and then, if version is empty then get the latest version? Another nit but this if confused me a bit.

cli/cmd/component.go
```diff
@@ -389,6 +427,13 @@ func runComponentsInstall(_ *cobra.Command, args []string) (err error) {
 	cli.OutputChecklist(successIcon, "Component configured\n")
 	cli.OutputHuman("\nInstallation completed.\n")
 
+	params["configure_duration_ms"] = time.Since(start).Milliseconds()
+
+	cli.Event.Component = component_name
+	cli.Event.Feature = "install_component"
```
afiune, line 433: Let's move this assignations (feature and component name) at the beginning of the function, that way if we error out, we will send the Honeyvent with the information already (plus the error information)

lwcomponent/component.go
```diff
@@ -293,11 +288,38 @@ func (s State) Install(name string, version string) error {
...
+	// APIs out there. In particular, 0.2.0 is the version of the API result before
+	// https://github.com/lacework/rainbow/pull/10874 which added the capability for
+	// the API to return multiple versions. At the time of writing, that PR is merged
+	// but not fully rolled out. In the old version of the API result, we would only
+	// return the latest version, and wouldn't set the version field of its artifacts.
+	// So, if we're looking for the latest version we should set what we're looking
+	// for to the empty string. If we're looking for anything else, we can leave it
+	// set as it is: we'll fail to find it and report an error which is the right
+	// behaviour as only the latest version can be installed in this case.
+	if s.Version == "0.2.0" && version == component.LatestVersion.String() {
```
afiune, line 305: @edoardopirovano do we still need this?

edoardopirovano: I think we can drop this, and also the case here: https://github.com/lacework/go-sdk/pull/1384/files#diff-b20e74d5abf181172ebf9b793963628be0345afb5ab1f7492b090df0e8082d66L245

PR #1397, opened 2023-09-29
Support sast tables in cds

sergiogiro:
Add the parameter to specify the url type for sast tables

## Summary

We need to prepare the API for uploading sast tables.

Unit tests

## Issue

https://lacework.atlassian.net/browse/COD-1156

### DIFF
--- a/api/component_data.go
+++ b/api/component_data.go
@@ -15,11 +15,17 @@ type ComponentDataService struct {
 	client *Client
 }
 
+const URL_TYPE_DEFAULT = "Default"
+const URL_TYPE_SAST_TABLES = "SastTables"
+
+var URL_TYPES = []string{URL_TYPE_DEFAULT, URL_TYPE_SAST_TABLES}
+
 type ComponentDataInitialRequest struct {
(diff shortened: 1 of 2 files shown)


api/component_data.go
```diff
@@ -53,8 +60,19 @@ type ComponentDataCompleteResponse struct {
 	Guid string `json:"guid,omitempty"`
 }
 
-func (svc *ComponentDataService) UploadFiles(name string, tags []string, paths []string) (string, error) {
-	initialRequest, err := buildComponentDataInitialRequest(name, tags, paths)
+func (svc *ComponentDataService) UploadFiles(
```
linpengzhang, line 63: this is a breaking change right? this functions is used by other components and we are changing signature. I'd prefer to avoid doing that and making the change backward compatible

linpengzhang (changes requested):

api/component_data.go
sergiogiro: Made it backwards compatible now

linpengzhang (approved):
lgtm

PR #1398, opened 2023-10-04
chore(CAD-508): Remove team-members from go cli

zekisherif:
## Summary

Since RBAC v2 has been enabled for all customers for quite a while. We don't need this API in documentation anymore as it will not work.

This is also being down now because it was raised as an issue in https://lacework.atlassian.net/browse/LINK-1972

No testing needed. This Pr is all about deleting files!

## Issue

 https://lacework.atlassian.net/browse/LINK-1972 https://lacework.atlassian.net/browse/CAD-508


afiune (changes requested):
We can't just remove commands as a "CHORE" - You need to

1. Hide this command and mark it as DEPRECATED
2. Wait until we do a `2.0.0` release and in there remove it completely
3. The command should continue to work for a period of time (3-6 months)

We can decide the time but definitely the removal must be a major version bump

PR #1407, opened 2023-10-12
feat: Components daily version check

linpengzhang:

### DIFF
--- a/cli/cmd/version.go
+++ b/cli/cmd/version.go
@@ -131,17 +131,75 @@ func versionCheck() (*lwupdater.Version, error) {
 	return sdk, nil
 }
 
-// dailyVersionCheck will execute a version check on a daily basis, the function uses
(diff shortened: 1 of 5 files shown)


cli/cmd/version.go
```diff
@@ -131,17 +131,75 @@ func versionCheck() (*lwupdater.Version, error) {
...
+		return component.Status() == lwcomponent.UpdateAvailable, nil
+	} else {
+		return false, nil
+	}
+}
+
+// dailyVersionCheck will execute a version check on a daily basis, the function uses
+// the file ~/.config/lacework/version_cache to track the last check time
+func dailyVersionCheck() error {
+	if !isCheckEnabled() {
```
afiune, line 200: I didn't know that we were not checking for JSON output. 🤕 

Could we also check that here?
```suggestion
	if cli.JSONOutput() || !isCheckEnabled() {
```

afiune (approved):

PR #1409, opened 2023-10-12
refactor(state.Install): add ability to install Python oneFolder archives

tmac1973:
add internal/archive to detect and process tgz files and refactored state.Install to support downloading tgz archives

## Summary

- Add the ability to install python oneFolder components. 
- Additionally refactor the state.Install method to stage component downloads before install. 
- Add internal/archive to support untar/gunzip functionality along with the ability to detect tar and gz files without relying on file extensions. 

- I created tests for the internal/archive functions. 
- I built the lacework binary and tested installing a python oneFolder archive from s3 as well as single file components such as "iac". 

Note: I have not added a mock python oneFolder component yet 

## Issue

https://lacework.atlassian.net/browse/GROW-1602


internal/archive/detect.go
```diff
@@ -0,0 +1,89 @@
...
+func FileIsTar(file string) (isTar bool, err error) {
+	mtype, err := detectFileType(file)
+	if err != nil {
+		return
+	}
+	isTar = mtype == "application/x-tar"
+	return
+}
+
+func DetectTGZAndUnpack(filePath string, targetDir string) (err error) {
```
jon-stewart, line 50: This works, but a cool idea would be to have a recursive `Unpack` function instead, with a parameter to limit the recursion depth (and a default value).

```
Unpack(filepath, targetDir string) {
  switch TYPE {
    case GZ: {
      file = Gunzip()
      Unpack(file, targetDir)
    }
    case Tar: {
      file = Untar()
      Unpack(file, targetDir)
    }
  }
}
```

jon-stewart (approved):

PR #1411, opened 2023-10-13
fix: CDK bugs

jon-stewart:
## Summary

1) No components returned from the API would lead to a panic.
Fix:  safe array access

2) No `~/.config/lacework/components` directory would lead to an error
Fix: create the dir if it doesn't exist

Local manual testing


PengyuanZhao (approved):

afiune (commented):
Can we make sure next changes come with tests?

PR #1412, opened 2023-10-13
feat: check if cdk component is tainted

PengyuanZhao:
## Summary

- For the install and update command, if the specified version is tainted, then the command will fail by default. Nothing is changed.
- For the list and show command, set the component status to "Tainted", displayed as "Tainted(Please update)", to prompt users to update.

lacework components list
Screenshot 2023-10-16 at 9 10 26 AM

lacework components show component-example
Screenshot 2023-10-16 at 9 10 36 AM

## Issue

https://lacework.atlassian.net/browse/GROW-2367

### DIFF
--- a/lwcomponent/catalog.go
+++ b/lwcomponent/catalog.go
@@ -259,13 +259,23 @@ func NewCatalog(client *api.Client, stageConstructor StageConstructor) (*Catalog
 
 	cdkComponents := make(map[string]CDKComponent, len(rawComponents)+len(localComponents))
 
+	allVersions := []semver.Version{}
+
+	for _, c := range rawComponents {
+		ver, err := semver.NewVersion(c.Version)
+		if err != nil {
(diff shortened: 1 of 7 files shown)


lwcomponent/catalog.go
```diff
@@ -259,13 +259,23 @@ func NewCatalog(client *api.Client, stageConstructor StageConstructor) (*Catalog
 
 	cdkComponents := make(map[string]CDKComponent, len(rawComponents)+len(localComponents))
 
+	allVersions := []semver.Version{}
+
+	for _, c := range rawComponents {
```
jon-stewart, line 264: This doesn't have all of the available versions.  We need to call API `ListComponentVersions`.

Like in https://github.com/lacework/go-sdk/blob/main/lwcomponent/catalog.go#L80

An API call for every component installed.  Thankfully we will cache this.

jon-stewart (approved):
👍

PR #1419, opened 2023-10-18
fix(cdk): place Python component binary in component root

tmac1973:
## Summary

The python pyinstaller build command was creating the component binary in subdirectory, not the component root directory. This prevented the lacework cli from executing the component. 

1. compiled the lacework cli
2.  tested the creation of a new python component with scaffolding 
3. tested that the cli could execute the component after build

## Issue

https://lacework.atlassian.net/browse/GROW-2491


PengyuanZhao (approved):
LGTM

PR #1429, opened 2023-11-01
chore: use AlertRule subCategory in stead of eventCategory

PengyuanZhao:
## Summary

- `eventCategory` is deprecated. A new field `subCategory` is added to the AlertRules API in [#14700](https://github.com/lacework/rainbow/pull/14700).
- There are 3 eventCategory values that are not in sync with subCategories: `App`, `Cloud` and `K8sActivity`. To support them for existing users, they are converted to corresponding subCategory values.

`make test`
`make integration-context-tests`

## Issue
https://lacework.atlassian.net/browse/GROW-2490

### DIFF
--- a/api/alert_rules.go
+++ b/api/alert_rules.go
@@ -58,19 +58,6 @@ var AlertRuleSubCategories = []string{
 	"K8sActivity",
 }
 
-// var AlertRuleEventCategories = []string{
-// 	"Compliance",
-// 	"App",
-// 	"Cloud",
-// 	"File",
-// 	"Machine",
-// 	"User",
-// 	"Platform",
-// 	"K8sActivity",
-// 	"Registry",
-// 	"SystemCall",
-// }
-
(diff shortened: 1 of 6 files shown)


api/alert_rules.go
```diff
@@ -40,15 +40,22 @@ var AlertRuleCategories = []string{"Anomaly", "Policy", "Composite"}
 // Valid inputs for AlertRule SubCategories property
 var AlertRuleSubCategories = []string{
```
jon-stewart, line 41: Pity these are not enum.  Lines 153+ could refer to a type instead of repeating the strings.

jon-stewart (approved):

api/alert_rules.go
PengyuanZhao: Yes. The strings will be used in terraform to enforce valid values. Will leave it to keep it simple.

PR #1431, opened 2023-11-02
fix(GROW-2556): enable azure storage account infra encryption

ipcrm:
## Summary

Enable Infrastructure Encryption by default when creating a new Azure Storage Account for activity log integration.

* updated automated tests
* executed manually

## Issue

https://lacework.atlassian.net/browse/GROW-2556


afiune (approved):
Awesome. 

Does this "encryption feature" requires more permissions? I am thinking about our Deployment Role for Azure and how will we codify this into a Checklist (in the future)

ipcrm:
re: permissions, not to my knowledge because it's CSP managed (in this case)

PR #1435, opened 2023-11-06
feat(RAIN-74824): support queryLanguage when creating query

qdfuxiao:


afiune:
We will cover it with https://github.com/lacework/go-sdk/pull/1467

PR #1438, opened 2023-11-07
fix(generate): load AWS_PROFILE & AWS_REGION environment variables

afiune:
We were not loading AWS environment variables before this change.

Wrote integration tests and ran command locally.

## Issue

N/A

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,6 +158,10 @@ run-api-example: ## Run an API example like 'make run-api-example example=api/_e
 		LW_SUBACCOUNT=$(shell lacework configure show subaccount) \
 		go run $(example)
 
+.PHONY: build
+build: ## Compiles binary for the running workstation
+	go build -o bin/lacework -ldflags=$(GO_LDFLAGS) github.com/lacework/go-sdk/cli
+
 .PHONY: build-cli-cross-platform
 build-cli-cross-platform: ## Compiles the Lacework CLI for all supported platforms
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -158,6 +158,10 @@ run-api-example: ## Run an API example like 'make run-api-example example=api/_e
 		LW_SUBACCOUNT=$(shell lacework configure show subaccount) \
 		go run $(example)
 
+.PHONY: build
+build: ## Compiles binary for the running workstation
+	go build -o bin/lacework -ldflags=$(GO_LDFLAGS) github.com/lacework/go-sdk/cli
```
ipcrm, line 163: 🎉

ipcrm (commented):
I'm wondering if this is always the right thing to do. 

I think this is the first place that we’ve deviated from go-sdk controlled settings? I think there’s an interesting question being raised here; e.g. for example if I have AWS_PROFILE set and we're respecting that why aren't we reading the region from that profiles setting? Why not the role (shared config), etc.

I feel like if we want to process these two vars we may need to go a step further and load the SDK so that _all_ host-configurations are picked up. It feels a bit like this is putting ourselves 'half-in' the pool.

Would love to discuss a bit further (or if other reviewers feel this is a non-issue, then I'm happy to be overruled).

afiune:
@ipcrm I appreciate your feedback. Let's hold on this change for now.

PR #1441, opened 2023-11-07
feat(cdk): expose cache to components

ipcrm:
## Summary

Enable CDK components to use the Lacework CLI cache.

Integration test, manually using `go-component`

## Issue

https://lacework.atlassian.net/browse/GROW-2498


proto/v1/cdk.proto
```diff
@@ -52,3 +60,22 @@ message HoneyventRequest {
...
+}
+
+message ReadCacheResponse {
+  bool miss = 1;
+  optional bytes data = 2; // body is empty if cache miss
+}
+
+message WriteCacheResult {
+  bool success = 1;
+  optional string msg = 2; // error message is empty if success true
```
afiune, line 80: nit: In Go, we are used to do `if err != nil` so, could we flip this to be `error` so we can do:
* `if response.Error` instead of
* `if !response.Success`
```suggestion
  bool error = 1;
  optional string message = 2; // error message is empty if error false
```

proto/v1/cdk.proto
```diff
@@ -52,3 +60,22 @@ message HoneyventRequest {
...
+  google.protobuf.Timestamp expires = 3;
+}
+
+message ReadCacheRequest {
+  string key = 1;
+}
+
+message ReadCacheResponse {
+  bool miss = 1;
+  optional bytes data = 2; // body is empty if cache miss
```
afiune, line 75: nit: same
```suggestion
  bool hit = 1;
  optional bytes data = 2; // body is empty if cache miss (hit == false)
```

proto/v1/cdk.proto
ipcrm: Will revise.

proto/v1/cdk.proto
ipcrm: Will revise.

proto/v1/cdk.proto
ipcrm: Fixed in cf85c3009a3e40c0df3c2d526721c3f99b8a56be

proto/v1/cdk.proto
ipcrm: Fixed in cf85c3009a3e40c0df3c2d526721c3f99b8a56be

Makefile
```diff
@@ -148,7 +148,7 @@ fmt-check: ## Lists formatting issues
 
 .PHONY: imports-check
 imports-check: ## Lists imports issues
-	@test -z $(shell goimports -l $(shell go list -f {{.Dir}} ./... | grep -v proto))
+	@test -z $(shell goimports -l $(shell find . -type f -name '*.go' -not -path './vendor/*' -not -path './cli/cdk/go/proto/*'))
```
afiune, line 151: not platform agnostic but it's fine, I don't think we run this test in CI on windows...

afiune (approved):
Outstanding change! 🙌🏽

Makefile
ipcrm: Please take a look at the commit message on e043b66030a59f5ad1052afbd55977e914a087c2

It doesn't seem like go-list has a better way to solve this, but I'm open to ideas!

PR #1444, opened 2023-11-09
fix(lwtime): fix issues related to daylight savings

hazedav:
## Summary
If a particular (week / day) relative time offset ends up spanning a daylight savings time event, it will be off by an hour (as evident in the startTime and endTimes in the example below).

```
❯ lacework alert ls --range "last 7 days" --debug                            
...
["application/json"],"Method":["POST"],"User-Agent":["Command-Line/1.33.3"]},"body":"{\"timeFilter\":{\"startTime\":\"2023-11-02T17:16:09.613018Z\",\"endTime\":\"2023-11-09T18:16:09.613018Z\"}}\n"}
...
ERROR unable to list alerts: 
  [POST] https://dev7.dev7.corp.lacework.net/api/v2/Alerts/search
  [400] Difference between startTime and endTime should not be greater than 7 Days
```

We can mitigate this by normalizing week and day offsets back to a constant number of hours.

## Limitations
Due to months and years having a non-deterministic number of hours, these will not be normalized at this time.

Unit testing

## Issue
https://lacework.atlassian.net/browse/LINK-2308


lwtime/nattime_internal_test.go
```diff
@@ -367,3 +368,32 @@ func TestNaturalFromTime(t *testing.T) {
...
+	}
+
+	startTime, endTime, err := parseNaturalFromTime("Last 7 days", refTime)
+	if err != nil {
+		assert.FailNow(t, err.Error())
+	}
+	fmt.Println(endTime.UTC())
+
+	diff := endTime.Sub(startTime)
+	fmt.Println(diff.Hours())
```
ipcrm, line 396: still needed?

ipcrm (approved):
👏 awesome!


Also tested locally and confirm the change resolved the issue.

lwtime/nattime_internal_test.go
hazedav: If the test fails, this will give some output.

hazedav: I can remove.

PR #1462, opened 2023-11-27
feat(GROW-2590): CDK .dev file info

jon-stewart:
## Summary

CDK v1 code path now using `.dev` file data.

We only need to compile for our platform when we install the CLI using make.  This is a lot faster.

1. `lacework component switch techally`
2. `lacework component dev tester`
3. `lacework component switch qan`
4.

```
> lacework component list
       STATUS                 NAME              VERSION                   DESCRIPTION
-------------------+-------------------------+-----------+--------------------------------------------
  Installed          tester                    0.0.0-dev   (dev-mode) testing stuff

Components version: 0.3.0
```

## Issue

https://lacework.atlassian.net/browse/GROW-2590

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -212,16 +212,8 @@ protoc-python: install-tools
 		--grpc_python_out=./cli/cdk/python proto/v1/*.proto
 
 .PHONY: install-cli
-install-cli: build-cli-cross-platform ## Build and install the Lacework CLI binary at /usr/local/bin/lacework
-ifeq (x86_64, $(shell uname -m))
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -171,6 +171,14 @@ build-cli-cross-platform: ## Compiles the Lacework CLI for all supported platfor
             -ldflags=$(GO_LDFLAGS) \
             github.com/lacework/go-sdk/cli
 
+.PHONY: build-cli
+build-cli: ## Compiles the Lacework CLI for your platform
```
afiune, line 175: what is the difference with `build` directive above? Should we replace it in favor of this one?

jon-stewart: I didn't see that.  Yeah, we should use `build` instead.

PengyuanZhao (approved):

PR #1466, opened 2023-11-29
feat: go-resty component download

jon-stewart:
## Summary

- Using `go-resty`
- Download progress
  - progress shown in CLI spinner
  - prototype components show progress in MB (we do not have access to total file size)
  - v1 components show process in %
  - CLI spinner updated if value changes
- No request timeout
  - can be set with environmental variable `CDK_DOWNLOAD_TIMEOUT_MINUTES`

Unit tests + manual

https://github.com/lacework/go-sdk/assets/5750305/8dd3bc7b-b45d-47e8-ad0c-086bc4cbd4f5

## Issue

https://lacework.atlassian.net/browse/GROW-2596


lwcomponent/http.go
```diff
@@ -19,58 +19,89 @@
...
+			if totalSize == 0 {
+				fmt.Printf("Downloaded: %.0fkb\n", float64(size)/(1<<10))
+			} else {
+				var percent float64 = float64(size) / float64(totalSize) * 100
+				fmt.Printf("Downloaded: %.0f", percent)
+				fmt.Println("%")
+			}
+		}
+
+		time.Sleep(time.Second)
```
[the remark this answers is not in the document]
ClassicStocksy: Arguably it's fine as is - 30 seconds would be pretty slow, and would only be 30 lines of logged output...

jvogt:
Is the 5 minute timeout a hard limit? My expectation is the download will not timeout as long as there is progress.

Example: I was on conference wifi, roughly 3mbit, and it took about 15 minutes to download the `sast` component manually with curl.

I would expect a sane timeout if the download does not start (something like 15 seconds), **no timeout** if there is progress, and maybe a longer timeout (1-5 minutes) if no data has been received.

lwcomponent/http.go
```diff
@@ -19,58 +19,89 @@
...
 	"time"
+
+	"github.com/go-resty/resty/v2"
 )
 
 const (
-	defaultTimeout  = 30 * time.Second
-	defaultMaxRetry = 2
+	DefaultMaxRetry = 3
+	defaultTimeout  = 5 * time.Minute
```
jon-stewart, line 32: environmental variable

lwcomponent/http.go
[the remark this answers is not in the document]
jon-stewart: Thinking - show unique percentages.  Show MB's downloaded and unique MB progress.

lwcomponent/http.go
```diff
@@ -19,58 +19,89 @@
...
+				log.Fatal(err)
+			}
+
+			size := info.Size()
+			if size == 0 {
+				size = 1
+			}
+
+			if totalSize == 0 {
+				fmt.Printf("Downloaded: %.0fkb\n", float64(size)/(1<<10))
```
afiune, line 97: The `lwcomponent` should not interact with STDOUT or STDERR

afiune (changes requested):

lwcomponent/http.go
jon-stewart: How should we indicate progress?

jon-stewart: I am now using a virtual terminal - is this the approach you had in mind?

lwcomponent/http.go
```diff
@@ -19,58 +19,109 @@
...
+					_, _ = vterm.Write([]byte(fmt.Sprintf("Downloaded: %.0fmb\n", mb)))
+
+					previous = mb
+				}
+			} else {
+				percent := float64(size) / float64(totalSize) * 100
+
+				if percent > previous {
+					_, _ = vterm.Write([]byte(fmt.Sprintf("Downloaded: %.0f", percent)))
+					_, _ = vterm.Write([]byte("%"))
```
ClassicStocksy, line 118: Not familiar with `vterm` but I think the previous commit for this had `println` so there'd be a new line after the percent sign - will there still be a new line?

dmurray-lacework (approved):

afiune (approved):

PR #1467, opened 2023-12-01
feat(cli): Select query language only on query creation

incredibler:
## Summary

For lacework query create command, if a customer doesn’t run it with a -f/--file or -u/--url option, it will be in EDITOR mode and a template query text will be in a editor for the customer to edit.

Rego query needs a different query text template than LQL, so we need to add a selector to it so it’s known which query template to provide.  The supported language can be either `Rego` or `LQL`. By default it's `LQL`. This selector will be only displayed to customers for whom Rego is enabled.  

* build and install the CLI from source:  `make prepare && make install-cli`
* Run query create `lacework query create`, it enters interactive mode and the survey works as expected.

### DIFF
--- a/cli/cmd/lql.go
+++ b/cli/cmd/lql.go
@@ -212,11 +212,14 @@ func setQuerySourceFlags(cmds ...*cobra.Command) {
 				"file", "f", "",
 				fmt.Sprintf("path to a query to %s", action),
 			)
-			cmd.Flags().StringVarP(
-				&queryCmdState.Language,
-				"language", "l", "",
-				fmt.Sprintf("language of the query to %s, can be LQL or Rego, required in EDITOR mode", action),
-			)
+			// enable --language option for query create
+			if action == "create" {
+				cmd.Flags().StringVarP(
+					&queryCmdState.Language,
+					"language", "l", "LQL",
+					fmt.Sprintf("language of the query to %s, can be LQL or Rego", action),
+				)
+			}
 			// url flag to specify a query from url
 			cmd.Flags().StringVarP(
 				&queryCmdState.URL,
--- a/cli/cmd/lql.go
+++ b/cli/cmd/lql.go
@@ -44,6 +44,7 @@ var (
 		Range         string
 		Start         string
 		URL           string
+		Language      string
 		ValidateOnly  bool
 		FailOnCount   string
 		EmptyTemplate bool
@@ -211,6 +212,11 @@ func setQuerySourceFlags(cmds ...*cobra.Command) {
 				"file", "f", "",
 				fmt.Sprintf("path to a query to %s", action),
 			)
+			cmd.Flags().StringVarP(
+				&queryCmdState.Language,
+				"language", "l", "",
+				fmt.Sprintf("language of the query to %s, can be LQL or Rego, required in EDITOR mode", action),
+			)
 			// url flag to specify a query from url
 			cmd.Flags().StringVarP(
 				&queryCmdState.URL,
@@ -248,7 +254,7 @@ func inputQuery(cmd *cobra.Command) (string, error) {
 	if !queryCmdState.ValidateOnly {
 		action = strings.Split(cmd.Use, " ")[0]
 	}
-	return inputQueryFromEditor(action)
+	return inputQueryFromEditor(action, queryCmdState.Language)
 }
 
 func inputQueryFromLibrary(id string) (string, error) {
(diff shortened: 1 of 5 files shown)


cli/cmd/lql.go
```diff
@@ -296,14 +302,19 @@ func inputQueryFromURL(url string) (query string, err error) {
 	return
 }
 
-func inputQueryFromEditor(action string) (query string, err error) {
+func inputQueryFromEditor(action string, language string) (query string, err error) {
+	if language != "Rego" && language != "LQL" {
```
gspofford-lw, line 306: Don't do this, please. Let the server decide.

cli/cmd/lql.go
```diff
@@ -296,14 +302,19 @@ func inputQueryFromURL(url string) (query string, err error) {
...
+		return
+	}
 	prompt := &survey.Editor{
 		Message:  fmt.Sprintf("Type a query to %s", action),
 		FileName: "query*.yaml",
 	}
 
 	if (action == "create" || action == "run") && !queryCmdState.EmptyTemplate {
-		prompt.Default = `queryId: YourQueryID
+		if language == "LQL" {
```
gspofford-lw, line 316: I wonder if a `switch (language)` would be more idiomatic. The `default` clause might be used to create an error too.

incredibler: Yes switch looks better. Has updated it.

cli/cmd/lql.go
```diff
@@ -44,6 +44,7 @@ var (
 		Range         string
 		Start         string
 		URL           string
+		Language      string
```
hazedav, line 47: Fundamentally I don't think we need to add `--language` as a flag at all given that it's part of the query markup (for the same reason that queryId and queryText are not flags...

hazedav: Ah.  I think I see what's going on.  This is for us to be able to help pre-populate if using the editor.  I think it might be cleaner to simply use a survey.Select with `LQL` and `Rego` options, only if and when we're in the `inputQueryFromEditor` flow...

I think adding a flag is confusing when creating query from file, url, etc. when this will already have the `queryLanguage` specified.

incredibler: We are adding this option to decide which query template to provide in EDITOR mode - Rego query is different from LQL query.

incredibler: Can you give me an example of how a survey is defined? @hazedav

hazedav: Sure thing bud.  We use `survey.Select` a bunch in go-sdk.  Here is one such example:
https://github.com/lacework/go-sdk/blob/06b930715167c526ce9f8301d9a0cae23c8e9bdc/cli/cmd/container_registry.go#L160

Basically before opening the editor we'll prompt the user to select their desired language.

cli/cmd/lql.go
```diff
@@ -310,8 +296,22 @@ func inputQueryFromURL(url string) (query string, err error) {
...
-func inputQueryFromEditor(action string, language string) (query string, err error) {
+func inputQueryFromEditor(action string) (query string, err error) {
 	var queryTextTemplate string
+	language := "LQL"
+	if action == "create" {
+		languageSelect := &survey.Select{
+			Message: "Choose query language to create: ",
+			Options: []string{
+				"LQL",
+				"Rego",
```
afiune, line 306: As much as I like the use of Survey here, what will happen if users select Rego but they are not part of the Private Preview customers that have Rego enabled in the backend?

incredibler: There is a feature flag `lpp_rego_enabled` telling whether a customer can use Rego or not. However, it seems we cannot call `GET /api/v2/FeatureFlags/:tag` endpoint for FF check in a public repo like https://github.com/lacework/go-sdk

varming: > what will happen if users select Rego but they are not part of the Private Preview customers that have Rego enabled in the backend?

They should get a suitable 400 error.

integration/lql_create_test.go
```diff
@@ -38,8 +38,10 @@ func TestQueryCreateHelp(t *testing.T) {
 
 func TestQueryCreateEditor(t *testing.T) {
 	out, err, exitcode := LaceworkCLIWithTOMLConfig("query", "create")
-	assert.Contains(t, out.String(), "Type a query to create")
-	assert.Contains(t, out.String(), "[Enter to launch editor]")
+	assert.Contains(t, out.String(), "Choose query language to create")
```
varming, line 41: I would not expect this type of prompt to be released until Lacework officially supports Rego for all customers.

Can we make the argument an optional string that we don't prompt the customer for as long as we are in private preview? This way customers that are not part of the preview are not impacted by Rego support and customers that are part of the preview can use the appropriate flag.

incredibler: I've added a FF check to ensure the prompt to choose a query language is only displayed to customers for whom Rego is enabled.

afiune (approved):

hazedav:
This looks good to me.  Should we add an integration test for this?  It looks like perhaps we had one at some point?

hazedav (approved):

incredibler:
> This looks good to me. Should we add an integration test for this? It looks like perhaps we had one at some point?

If there's an established way of enabling a FF for integration test, I can try to add one.

PR #1468, opened 2023-12-04
refactor(cli): deprecate report-definitions command

afiune:
## Summary

We will be deprecating Report Definitions APIs, we are hidding this command so that it does not
get used any further and the documentation gets removed from https://docs.lacework.net/cli/commands/lacework_report-definition

N/A

## Issue

N/A


dmurray-lacework:
Help tests need updated

ipcrm (approved):

PR #1472, opened 2023-12-06
feat(RAIN-91461): support Rego in query validate

incredibler:
## Summary

This is to support an extra field queryLanguage in the input for query validation command.  It’s optional, by default it should be LQL

* build and install the CLI from source: make prepare && make install-cli
* Create a file containing a Rego query:
```
$ cat /Users/kchen/proj/opa/rego_query.yaml
---
queryId: TestRegoQueryViaCli2
queryLanguage: Rego
queryText: |-
  package lacework.policies.lacework_global_134
  import future.keywords
  import data.lacework
  import data.lacework.assessment
  source := lacework.aws.cfg.list("s3", "list-buckets")
  assess := assessment.violation(input, "just because")
```
*  Query validate command can be executed successfully against the file:
```
$ /usr/local/bin/lacework query validate  -f /Users/kchen/proj/opa/rego_query.yaml
Query validated successfully. Time for 🍕
```

### DIFF
--- a/api/lql_validate.go
+++ b/api/lql_validate.go
@@ -19,7 +19,8 @@
 package api
 
 type ValidateQuery struct {
-	QueryText string `json:"queryText"`
+	QueryText     string  `json:"queryText"`
+	QueryLanguage *string `json:"queryLanguage,omitempty" yaml:"queryLanguage,omitempty"`
 }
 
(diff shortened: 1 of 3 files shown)


hazedav:
Should we also add an integration test for this?

PengyuanZhao (approved):

PR #1481, opened 2023-12-12
feat(cli): Add time filter to vulnerability host list-cve command

kendall-choy-lw:
## Summary
The `lacework vulnerability host list-cves` command is timing out for some customers with a large amount of data. The command defaults to a 24 hour window. We can replicate this by calling the API with the same 24h window and it times out. Enable the start time, end time, and range flags to enable querying smaller amounts of data. 

Passed unit tests
Passed integration tests
Build CLI locally and manually tested the flags

## Issue
https://lacework.atlassian.net/browse/LINK-2415
https://lacework.atlassian.net/browse/RAIN-91480

### DIFF
--- a/cli/cmd/vuln_host.go
+++ b/cli/cmd/vuln_host.go
@@ -87,6 +87,7 @@ func init() {
 	)
 
 	setTimeRangeFlags(
+		vulHostListCvesCmd.Flags(),
 		vulHostListHostsCmd.Flags(),
 	)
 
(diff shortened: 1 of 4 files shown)


afiune (commented):
Could we please add a test like this one https://github.com/lacework/go-sdk/blob/main/integration/host_vulnerability_test.go#L57 with the new flags and a short time range? 2 hours or so

afiune (approved):

PR #1508, opened 2024-01-10
fix: component arg parse

jon-stewart:
## Summary

This is the problem https://github.com/lacework/go-sdk/blob/main/cli/cmd/component_args.go#L83-L86.

```
		// We're actually a bit stuck here as we don't know if this flag
		// takes an argument or not, so we don't know whether or not to consume
		// the next arg.  What we'll do is peek ahead, and if the next arg does
		// not start with - then we'll take it.
```

Cobra strips the surrounding double quotes and when the go-sdk sees a `-` prefix, it treats the argument as a 'shorthand' (e.g. -vvvv).

## The Fix

Simple, we don't need to parse the flags.  Use `--` to terminate flag parsing and pass them as arguments to the component, and let the component handle the parsing.

We benefit by having less code to maintain and instead lean upon the Cobra library.  We make use of the flag terminator `--`, like the K8 CLI.

Manually.  Create a component using `lacework component dev test --scaffolding Golang --type CLI_COMMAND`.  Add the following patch to the component code.

```
diff --git a/cmd/placeholder.go b/cmd/placeholder.go
index 10bc5ee..75e41ae 100755
--- a/cmd/placeholder.go
+++ b/cmd/placeholder.go
@@ -30,12 +30,24 @@ var (
                Long:  `This is just an example of a command.`,
                RunE: func(_ *cobra.Command, _ []string) (err error) {
                        fmt.Println("Your new component 'jontest' is ready!")
+
+            fmt.Println(TestState.time)
+
                        return nil
                },
        }
+    TestState = struct {
+        time string
+    }{}
 )

 func init() {
        // add the placeholder command
        rootCmd.AddCommand(placeholderCmd)
+
+        placeholderCmd.Flags().StringVar(
+        &TestState.time,
+        "time", "-t",
+        "start time for query",
+    )
 }
```

```
> bin/lacework --nocache test --json -- placeholder --time -24h
Your new component 'test' is ready!
-24h
```

## Issue

https://lacework.atlassian.net/browse/GROW-2677


afiune (changes requested):
This is a phantasmic way to solve this problem but it delegates the problem of
knowing CLI vs Component flags to the user. Which essentially means 1) a major
version bump, 2) a lot of documentation about the new arg parsing, and 3) one or
more communications to internal and external users about this change.

I am holding this change up until we exhaust all other options to solve this problem.

- Can we take a stab at others ways to solve this problem?

afiune:
We are going to go with https://github.com/lacework/go-sdk/pull/1512

PR #1512, opened 2024-01-12
fix: component args

jon-stewart:
## Summary

If a component flag is followed by an argument that has a `-` prefix, we treat it as an argument, and not as a shorthand flag.

I added further tests based on we need to [support](https://github.com/spf13/pflag?tab=readme-ov-file#command-line-flag-syntax).

There is a problem.  We are unable to handle both shorthand expansion like `-xyz` -> `-x -y -z` and shorthand assignment such as `-n1234`, `-n1234` is equivalent to `-n=1234` and `-n 1234`.  We cannot lookup the component flags to know that `-n1234` should not be expanded.  Just pass through to the component and let the component parse.

Inputs based on https://github.com/spf13/pflag?tab=readme-ov-file#command-line-flag-syntax

Unit testing and manual

## Issue

https://lacework.atlassian.net/browse/GROW-2677

### DIFF
--- a/cli/cmd/component_args.go
+++ b/cli/cmd/component_args.go
@@ -83,8 +83,8 @@ func (p *componentArgParser) parseLongArg(flags *pflag.FlagSet, s string, args [
 		// We're actually a bit stuck here as we don't know if this flag
 		// takes an argument or not, so we don't know whether or not to consume
 		// the next arg.  What we'll do is peek ahead, and if the next arg does
-		// not start with - then we'll take it.
-		if len(args) > 0 && len(args[0]) > 0 && args[0][0] == '-' {
+		// not start with '--' then we'll take it.
+		if len(args) > 0 && len(args[0]) > 0 && (args[0][0] == '-' && args[0][1] == '-') {
 			// This component flag does not take an arg
 			return args
 		}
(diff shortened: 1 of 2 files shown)


cli/cmd/component_args.go
```diff
@@ -126,14 +126,31 @@ func (p *componentArgParser) parseShortArg(flags *pflag.FlagSet, s string, args
 		if flag == nil {
 			// Not our flag, pass to the component.  Like the long form above we
 			// don't know whether to consume an extra arg, so we'll do the same
-			// thing: if the next arg does not start with - then pass it along
-			p.componentArgs = append(p.componentArgs, fmt.Sprintf("-%s", shorthand))
-			if len(shorthands) == 1 && (len(args) > 0 && len(args[0]) > 0 && args[0][0] == '-') {
+			// thing: if the next arg does not start with `--`` then pass it along
```
afiune, line 129:
```suggestion
			// thing: if the next arg does not start with `--` then pass it along
```

cli/cmd/component_args_test.go
```diff
@@ -75,6 +75,45 @@ func TestComponentArgs(t *testing.T) {
...
+			},
+			[]string{
+				"iac", "negative-values", "--time", "-24h",
+			},
+			[]string{},
+		},
+		{
+			// We do not have access to the component flag configuration and therefore cannot
+			// support the shorthand expansion of `-xyz` -> `-x -y -z`.  We have to treat it as
+			// `x=yz` and pass through to the component.
```
afiune, line 90: Dropping support of shorthand expansion is fine. It just limits our "expert" users to chain flags. If we get
complains, we have a way forward, and maybe in the future we switch to something similar to https://github.com/lacework/go-sdk/pull/1508

jon-stewart:
We may not be handling:
- `-abc`  => `-a=true -b=true -c=true`
- `-abcs "hello"` => `-a=true -b=true -c=true -s="hello"`
- `-absd="hello"` => `-a -b -s="hello" -d="hello"`
- `-abcs1234` => `-a=true -b=true -c=true -s=1234`

afiune (approved):
Added one more tests that I found interesting from our Telemetry via Honeycomb.

This looks good to go! Thank you, @jon-stewart

PR #1515, opened 2024-01-16
chore: set account details for honeycomb metrics

dmurray-lacework:
Set additional properties for honeycomb metrics

### DIFF
--- a/api/metrics.go
+++ b/api/metrics.go
@@ -18,14 +18,39 @@
 
 package api
 
+import "runtime"
+
 // MetricsService is a service that sends events to Lacework APIv2 Server metrics endpoint
 type MetricsService struct {
 	client *Client
 }
 
 func (svc *MetricsService) Send(event Honeyvent) (response HoneyEventResponse, err error) {
+	event.setAccountDetails(*svc.client)
 	err = svc.client.RequestEncoderDecoder("POST", apiV2HoneyMetrics, event, &response)
 	return
+}
+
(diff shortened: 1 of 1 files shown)


api/metrics.go
```diff
@@ -18,16 +18,41 @@
...
 	client *Client
 }
 
 func (svc *MetricsService) Send(event Honeyvent) (response HoneyEventResponse, err error) {
+	event.setAccountDetails(*svc.client)
 	err = svc.client.RequestEncoderDecoder("POST", apiV2HoneyMetrics, event, &response)
 	return
 }
 
+func NewHoneyvent(version, feature, dataset string) Honeyvent {
```
ipcrm, line 34: This doesn't seem to be used?

ipcrm (commented):
Question on the `NewHoneyvent` function

api/metrics.go
dmurray-lacework: At the moment it's not used in the go-sdk and cli. But it is used by terraform provider.

dmurray-lacework: See this PR for an example -> https://github.com/lacework/terraform-provider-lacework/pull/611/files#diff-145f8ae7d249239fe57b8eb4e7bd92129c5606ae1a30a2bd1ee9615ff7b26093R47

ipcrm: ty!

ipcrm (approved):

PR #1522, opened 2024-01-24
refactor: use v2/metrics for cli honeycomb events

dmurray-lacework:
- [x] Run CLI commands observe honeycomb dataset
- [x] Ensure Trace with multiple spans works as expected

## Issue
https://lacework.atlassian.net/browse/GROW-2705


jon-stewart (approved):

cli/cmd/cache.go
```diff
@@ -132,7 +132,7 @@ func (c *cliState) EraseCachedToken() error {
 }
 
 func (c *cliState) ReadCachedToken() {
-	if c.noCache {
+	if c.noCache || c.Cache == nil {
```
afiune, line 135: I feel you catch some issues plugin this in, I am now against these additions but I want to understand why we need to check for the cache here and also, initialize the api somewhere else that is not the `root.go`?

dmurray-lacework: Prior to this we could send Honeycomb events without the client being initialized.

In the persistent prerun we always sent a Honeycomb event, say for example `lacework help`.
We then got an event in honeycomb with the cmd,version but no account/subaccount or any client data.

Now we cannot do that. We need Lacework auth to send a honeycomb event. This also means those honeycomb events with no client data will not be sent.

PR #1530, opened 2024-02-06
chore(cli): print new component version available message

PengyuanZhao:
## Summary
- Use the existing func `dailyComponentUpdateAvailable` to check if the component version was checked 1 day ago.
- If yes, print message if component installed version is less than the latest version on the ApiInfo

Screenshot 2024-02-12 at 5 55 43 PM

## Issue

[](https://lacework.atlassian.net/browse/GROW-2589)


afiune (changes requested):
Are we going to print this message on EVERY execution? If so, this is a no-no, it will be too invasive for the customer.

We need to do this message only once a day. Just like our daily version check. And if it is possible, I would recommend printing this message at the end of the command execution, not at the beginning, otherwise it gets lost.

PengyuanZhao:
Updated to only print once a day, leveraging the existing function.

afiune (approved):
Much better, still I wish we could write some tests. Not a blocker, for now.

PR #1585, opened 2024-03-07
fix(telemetry): record silent error

afiune:
## Summary

We should not fail (in Honeycomb) when we can't check for updates, instead we are going to
send a silent error.

Built it and ran CLI locally, saw Honeycomb dev dataset.

### Before

Screenshot 2024-03-07 at 9 13 41 AM

### After

Screenshot 2024-03-07 at 10 28 47 AM

### DIFF
--- a/cli/cmd/version.go
+++ b/cli/cmd/version.go
@@ -181,7 +181,7 @@ func dailyComponentUpdateAvailable(componentName string) (bool, error) {
 		err := versionCache.StoreCache(cacheFile)
 
 		if err != nil {
-			cli.Event.Error = err.Error()
+			cli.Event.FeatureData = map[string]interface{}{"silent_error": err.Error()}
 			return false, err
 		}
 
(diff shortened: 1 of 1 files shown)


cli/cmd/version.go
```diff
@@ -193,12 +193,40 @@ func dailyComponentUpdateAvailable(componentName string) (bool, error) {
...
+		err            error
+		currentVersion *lwupdater.Version
+	)
+	defer func() {
+		cli.Event.Feature = featDailyVerCheck
+		if err != nil {
+			cli.Event.FeatureData = map[string]interface{}{"silent_error": err.Error()}
+		}
+		cli.SendHoneyvent()
+	}()
```
ipcrm, line 207: Will the CLI always stay running long enough for this to send? Looks like the daily check is called in the post-execution of the root command.

afiune: Yup. We have workers in place to wait for this. https://github.com/lacework/go-sdk/blob/main/cli/cmd/honeyvent.go#L115-L124

ipcrm: Perfect!

ipcrm (approved):

PR #1597, opened 2024-03-20
fix(cli): avoid panic while updating progress

afiune:
## Summary

While installing new component, we are seeing a panic similar to:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x2 addr=0x40 pc=0x10534999c]

goroutine 11 [running]:
github.com/lacework/go-sdk/cli/cmd.downloadProgress(0x14000820060?, {0x14000888120?, 0x0?}, 0x0)
	/codefresh/volume/go-sdk/cli/cmd/component.go:1177 +0x27c
github.com/lacework/go-sdk/cli/cmd.prototypeRunComponentsInstall.func1({0x14000888120?, 0x0?}, 0x1400005c8a0?)
	/codefresh/volume/go-sdk/cli/cmd/component.go:925 +0x34
created by github.com/lacework/go-sdk/lwcomponent.State.Install in goroutine 1
	/codefresh/volume/go-sdk/lwcomponent/component.go:279 +0x594
```

This is due to the fact that we were updating the progress `Suffix` unsafely. The fix is to use the
CLI functions to start a new progress every time we have an update inside the download func.

Additional things in this PR:
* Fixed infinite loop when trying to install a component that is already installed
* Added a make directive `test-only` to run only unit tests without `prepare`
* Enabled HTTP Tracing to output more information about the `EOF` error we are seeing
* A few styling improvements

Updated tests and ran multiple `install` and `uninstall`.

## Issue

https://lacework.atlassian.net/browse/GROW-2765

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -61,7 +61,10 @@ help:
 prepare: git-env install-tools go-vendor ## Initialize the go environment
 
 .PHONY: test
-test: prepare ## Run all go-sdk tests
+test: prepare test-only ## Run all go-sdk tests
+
+.PHONY: test-only
+test-only: ## Run all go-sdk tests only (without prepare)
 	$(eval PACKAGES := $(shell go list ./... | grep -v integration))
 	gotestsum -f testname --rerun-fails=3 --packages="$(PACKAGES)" \
 		-- -v -cover -run=$(regex) -coverprofile=$(COVERAGEOUT) $(PACKAGES)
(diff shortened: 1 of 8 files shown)


PengyuanZhao (approved):

PengyuanZhao:
LGTM except the `TestDownloadFile` error.

Issue #1615, opened 2024-04-22
feat: Add Team and Owner to MachineTags


vallieres:
# Feature Request

**Describe the Feature Request**
We are using the information from those field in our internal tooling, however, these two fields are present in API response, but not in the Go struct that used to unmarshal the response.

**Is your feature request related to a problem? Please describe**
A problem for us since we need it, but it's not a bug or error.

**Describe Preferred Solution**
Adding the two fields would be sufficient.

**Additional Context**
List any other information that is relevant to your issue. Stack traces, related issues, suggestions on how to add, use case, Stack Overflow links, forum links, screenshots, OS if applicable, etc.

ipcrm:
Some discussing here on this in #1616, see (https://github.com/lacework/go-sdk/pull/1616#issuecomment-2077067274). Will raise another PR to add a method that returns all the tags in a `map[string]interface{}`.

ipcrm:
@vallieres please take a look at #1618 and confirm that'll get what you need, thx!

PR #1616, opened 2024-04-22
feat: added owner and team fields

vallieres:
## Summary

We are using these two fields in our internal tool (Team and Owner), and they were missing from the Client.

I pulled the information from those fields in our local tooling by using a go.mod replace statement to override our lacework client with a local version.

## Issue

https://github.com/lacework/go-sdk/issues/1615

### DIFF
--- a/api/v2_vulnerabilities.go
+++ b/api/v2_vulnerabilities.go
@@ -527,7 +527,9 @@ type VulnerabilityHostMachineTags struct {
 	InternalIP                            string `json:"InternalIp"`
 	LwTokenShort                          string `json:"LwTokenShort"`
 	Name                                  string `json:"Name"`
+	Owner                                 string `json:"Owner"`
 	SubnetID                              string `json:"SubnetId"`
+	Team                                  string `json:"Team"`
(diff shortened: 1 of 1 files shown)


ipcrm:
@vallieres thank you for raising this!

In looking into this a bit it seems the `owner` and `team` fields are specific to your use case so I'm hesitant to add them globally like this. 

That said, would it solve your use case if we added a new method like `GetMachineTagsRaw` (or some better name) that just returns `map[string]interface{}` with all the tag values so you can use them as needed?

vallieres:
> @vallieres thank you for raising this!
> 
> In looking into this a bit it seems the `owner` and `team` fields are specific to your use case so I'm hesitant to add them globally like this.
> 
> That said, would it solve your use case if we added a new method like `GetMachineTagsRaw` (or some better name) that just returns `map[string]interface{}` with all the tag values so you can use them as needed?

Hi! Yeah that could work for sure. I figured adding the two fields was the simplest solution 😄

ipcrm:
No worries! I'll comment on the original issue and can raise that PR today for your review, if that works? We'd just close this PR.

PR #1637, opened 2024-06-04
feat(GROW-2928): Upgrade the gcp terraform version

leijin-lw:
## Summary

Upgrade the gcp terraform version to use gcp 5.0 module.


ipcrm (changes requested):
Thanks for the updates, for now let's _just_ update the module you had problems with. We need to plan how to test all of these since multiple are making major version jumps.

leijin-lw:
> Thanks for the updates, for now let's _just_ update the module you had problems with. We need to plan how to test all of these since multiple are making major version jumps.

Sure. I will only change the self-deployment after merge this PR.

ipcrm (approved):

PR #1639, opened 2024-06-06
feat: prosast component data upload

jon-stewart:
## Summary

ProSAST must upload with the `UrlType` of `ProSast`.

Improved resourceGroup integration test by telling user how to recover.

Unit test

## Issue

https://lacework.atlassian.net/browse/COD-3355

### DIFF
--- a/api/component_data.go
+++ b/api/component_data.go
@@ -17,8 +17,9 @@ type ComponentDataService struct {
 
 const URL_TYPE_DEFAULT = "Default"
 const URL_TYPE_SAST_TABLES = "SastTables"
+const URL_TYPE_PROSAST = "ProSast"
 
-var URL_TYPES = []string{URL_TYPE_DEFAULT, URL_TYPE_SAST_TABLES}
+var URL_TYPES = []string{URL_TYPE_DEFAULT, URL_TYPE_SAST_TABLES, URL_TYPE_PROSAST}
 
 type ComponentDataInitialRequest struct {
 	Name             string          `json:"name"`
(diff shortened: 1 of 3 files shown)


api/component_data.go
```diff
@@ -70,6 +71,10 @@ func (svc *ComponentDataService) UploadSastTables(
 	return svc.doUploadFiles(name, []string{"sast"}, paths, URL_TYPE_SAST_TABLES)
 }
 
+func (svc *ComponentDataService) UploadProSast(name string, paths []string) (string, error) {
+	return svc.doUploadFiles(name, []string{"sast"}, paths, URL_TYPE_PROSAST)
```
hemanthgk10, line 75: Sounds like you are using the same tag as the other one. Will it not have any impact?

```
func (svc *ComponentDataService) UploadSastTables(
	name string, paths []string) (string, error) {
	return svc.doUploadFiles(name, []string{"sast"}, paths, URL_TYPE_SAST_TABLES)
}
```

Do you need a separate tag?

jon-stewart: Thanks for reviewing this @hemanthgk10. It is strange that there are 2 'type' fields - legacy code.

The tags are used [here](https://github.com/lacework/rainbow/blob/main/iris/src/main/java/net/lacework/iris/componentdata/ComponentDataHelper.java#L246-L250) to distinguish between SCA/IAC and SAST RedPanda topics.

The `UrlType` is used [here](https://github.com/lacework/rainbow/blob/main/iris/src/main/java/net/lacework/iris/componentdata/ComponentDataHelper.java#L424-L433) to select a S3 key prefix.

I hope you can see from this [PR](https://github.com/lacework/rainbow/pull/18306/files) that if I want to choose a different S3 key prefix I need to use the `UrlType`. The `sast` tag doesn't need to change.

ipcrm (approved):

PR #1648, opened 2024-07-16
feat(GROW-2949): Provide organization_id for project level agentless integration

leijin-lw:
Provide organization_id for project level agentless integration.

Tested using `lacework generate cloud-account gcp` command line to generate agentless terraform.

### DIFF
--- a/lwgenerate/gcp/gcp.go
+++ b/lwgenerate/gcp/gcp.go
@@ -141,10 +141,6 @@ type GenerateGcpTfConfigurationArgs struct {
 
 	Projects []string
 
-	// GCP organization id for agentless integration. Agentless integration requires an organization id
-	// even for project level integration
-	AgentlessOrganizationId string
-
 	// Default GCP Provider labels
 	ProviderDefaultLabels map[string]interface{}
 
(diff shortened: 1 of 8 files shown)


lwgenerate/gcp/gcp.go
```diff
@@ -632,6 +643,8 @@ func createAgentless(args *GenerateGcpTfConfigurationArgs) ([]*hclwrite.Block, e
 			if args.OrganizationIntegration {
 				attributes["integration_type"] = "ORGANIZATION"
 				attributes["organization_id"] = args.GcpOrganizationId
```
ipcrm, line 645: Do we need both of these? Looks like AgentlessOrganizationId always overwrites GcpOrganizationId. Couldn't we just use the one?

ipcrm (changes requested):
Question on the new `AgentlessOrganizationId`

lwgenerate/gcp/gcp.go
leijin-lw: I don't want break the existing behavior by using existing GcpOrganizationId. It was used by config and pubsub integration as well. AgentlessOrganizationId is specially used in agentless integration.

ipcrm (changes requested):
@leijin-lw latest change looks good!

We also need to update the `lacework generate cloud-account gcp` command. Currently, if you use that command and enable agentless it doesn't require a org-id to be supplied. As I understand it that will prevent the agentless TF module from functioning correctly (apologies if I've got that wrong).

leijin-lw:
> @leijin-lw latest change looks good!
> 
> We also need to update the `lacework generate cloud-account gcp` command. Currently, if you use that command and enable agentless it doesn't require a org-id to be supplied. As I understand it that will prevent the agentless TF module from functioning correctly (apologies if I've got that wrong).

I don't think we have support for agentless integration in CLI.  As least agentless team didn't realize we have CLI to generate terraform last time I talked with them. 

https://github.com/lacework/go-sdk/blob/main/cli/cmd/cloud_account.go#L294

ipcrm (approved):

PR #1657, opened 2024-09-18
Allow user to define cloudtrail name

leijin-lw:

### DIFF
--- a/lwgenerate/aws/aws.go
+++ b/lwgenerate/aws/aws.go
@@ -1050,9 +1050,11 @@ func createCloudtrail(args *GenerateAwsTfConfigurationArgs) (*hclwrite.Block, er
 			attributes["consolidated_trail"] = true
 		}
 		// S3 Bucket attributes
+		if args.CloudtrailName != "" {
+			attributes["cloudtrail_name"] = args.CloudtrailName
+		}
 		if args.CloudtrailUseExistingTrail {
 			attributes["use_existing_cloudtrail"] = true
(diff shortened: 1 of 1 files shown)


lwgenerate/aws/aws.go
```diff
@@ -1050,6 +1050,9 @@ func createCloudtrail(args *GenerateAwsTfConfigurationArgs) (*hclwrite.Block, er
 			attributes["consolidated_trail"] = true
 		}
 		// S3 Bucket attributes
+		if args.CloudtrailName != "" {
```
PengyuanZhao, line 1053: Why adding this? Will the following `if args.CloudtrailUseExistingTrail` work?

leijin-lw: Because we want the user to define the *new* cloudtrail name. Existing go-sdk is using "lacework-cloudtrail" as default new cloudtrail name and we cannot change it. 

There is a bug happen when user failed to delete the old "lacework-cloudtrail" and the new cloudtrail has to be a different name

PengyuanZhao: Isn't the code below (line 1054) `attributes["cloudtrail_name"] = args.CloudtrailName` using the passed in `args.CloudtrailName`?

leijin-lw: I get your point. I'll remove the line. `attributes["cloudtrail_name"]` should not depends on`args.CloudtrailUseExistingTrail `

PR #1670, opened 2024-10-30
chore: Change go-sdk to v2

leijin-lw:


zekisherif:
@leijin-lw "line 74 in cli/cmd/resource_group_v2.go" looks like linting is referring to the old version of this function. It is actually the correct number of parameters

PR #1673, opened 2024-11-01
fix: Fix casing during reading of resource group

zekisherif:
## Summary

Bug in CLI causing us to output wrong casing for filter names.

Added a unit test. Also ran cli locally to visually confirm that the filter names where not getting changed when a camel case filter name is provided.


jonfranco224:
Can I be removed from email notifications for this please?

PR #1705, opened 2025-04-25
fix(LINK-2046: stop using integrations for get registries

kendall-choy-lw:
## Summary
The vulnerability container list-registries command is currently missing data. It relies on container registry integration data to get a list of scanned registries. However, proxy scanner and inline scanner integrations don't contain the registry domain, and we are only able to get this data by looking at the container assessment. We can use the container assessment API here to get a full list of scanned registries in the last 7 days

The vulnerability container list-assessments command is also missing data. It calls the container assessment API while filtering for registry domains coming from the integration information. This only include platform scans, and would also consistently hit the API 500,000 total row limitation. We can improve this by breaking up the API requests further by scanner type. Breaking it up into four separate requests by scanner type prevents us from missing non-platform scan assessments. However, we are still hitting the API row limitation.

Tested list-registries, list-assessments, and container scan commands with local built CLI

## Issue
https://lacework.atlassian.net/browse/LINK-2046

### DIFF
--- a/cli/cmd/vuln_container_scan.go
+++ b/cli/cmd/vuln_container_scan.go
@@ -109,7 +109,7 @@ func userFriendlyErrorForOnDemandCtrVulnScan(err error, registry, repo, tag stri
 		"Could not find vulnerability integrations",
 	) {
 
-		registries, errReg := getContainerRegistries()
+		registries, errReg := getPlatformScannerIntegrationContainerRegistries()
 		if errReg != nil {
 			cli.Log.Debugw("error trying to retrieve configured registries", "error", errReg)
 			return errors.Errorf("container registry '%s' not found", registry)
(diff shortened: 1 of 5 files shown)


cli/cmd/vuln_container_scan.go
```diff
@@ -109,7 +109,7 @@ func userFriendlyErrorForOnDemandCtrVulnScan(err error, registry, repo, tag stri
 		"Could not find vulnerability integrations",
 	) {
 
-		registries, errReg := getContainerRegistries()
+		registries, errReg := getPlatformScannerIntegrationContainerRegistries()
```
kendall-choy-lw, line 112: Since this cli command is to trigger a platform on-demand scan, we do only want registry domains from integrations

PR #1716, opened 2025-05-28
feat(preflight): build preflight package for AWS

PengyuanZhao:
## Summary

This PR creates the package which serves functions to do preflight checks for AWS integrations including Agentless, Config and CloudTrail.

The following tasks will be run sequentially:
- Fetch caller info
- Fetch all policies
- Extract permission from policies
- Find out the missing required permissions.
- Fetch other account details including:
  - Enabled regions
  - Existing CloudTrail
  - EKS clusters
  - If the call can access organization
    - Fetch management account ID
    - Fetch all account IDs and org unit IDs

Run integration test in `preflight_aws_test.go`

### DIFF
--- a/lwpreflight/aws/util.go
+++ b/lwpreflight/aws/util.go
@@ -7,8 +7,11 @@ import (
 )
 
 // ParseResourceName extracts the resource name from the caller identity Arn
-// for example if the Arn is arn:aws:iam::123456789012:user/MyUser
-// MyUser string is returned
+// Examples:
+//   - arn:aws:iam::123456789012:root -> root
+//   - arn:aws:iam::123456789012:user/MyUser -> MyUser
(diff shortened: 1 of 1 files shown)


lwpreflight/aws/util.go
```diff
@@ -0,0 +1,27 @@
...
+		return "", err
+	}
+
+	parts := strings.Split(arnObj.Resource, ":")
+	lastStr := parts[len(parts)-1]
+	paths := strings.Split(lastStr, "/")
+
+	if strings.Contains(lastStr, "assumed-role") {
+		return paths[len(paths)-2], nil
+	}
```
jeffreynglw, line 24: I think it would be useful to add an example comment on how this assumed-role Arn might look like.

PR #1718, opened 2025-06-03
feat(preflight): build preflight package for GCP

PengyuanZhao:
## Summary

This PR creates the package which serves functions to do preflight checks for GCP integrations including Agentless, AuditLog and Config

Depend on https://github.com/lacework/go-sdk/pull/1716

Run integration test: `preflight_gcp_test.go`

Run the following and check the output:
```
package main

import (
	"fmt"
	"os"

	"github.com/lacework/go-sdk/v2/lwpreflight/gcp"
)

func main() {
	preflight, err := gcp.New(gcp.Params{
		Agentless: true,
		Config:    true,
		AuditLog:  true,
		Region:    "us-west2",
		ProjectID: "abc-demo-project-123",
		CredentialsFile: "/Users/pengyuan/.config/gcloud/application_default_credentials.json",
	})
	if err != nil {
		fmt.Println(err)
		return
	}

	result, err := preflight.Run()
	if err != nil {
		fmt.Println(err)
		return
	}

	fmt.Printf("%#v\n", result)
}
```


lokesh-vadlamudi (approved):
LGTM

PR #1719, opened 2025-06-06
chore: update metrics service to use v2/telemetry/OtelMetrics

PengyuanZhao:
## Summary

We are migrating from Honecomb to Elastic. This PR updates go-sdk to use the new endpoint `v2/telemetry/OtelMetrics`.

Run a lacework command such as `lacework cloud-account list`, then verify the metric event is sent to Elastic succeffuly:

Screenshot 2025-06-06 at 4 02 41 PM

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -35,7 +35,7 @@ jobs:
 
       - name: Build Artifacts
         env:
-          HONEYDATASET: ${{ secrets.PROD_HONEYDATASET }}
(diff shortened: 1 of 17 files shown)


heidihe2024:
should we keep both sending honeycomb event and sending metric event before we verify the data in elastic? then we can deprecate honeycomb as soon as the data validation completed

PengyuanZhao:
@heidihe2024 As it shows above, I've tested and verified that the data can be sent to Elastic successfully and shows up correctly. Would that be adequate?

PR #1728, opened 2025-06-19
chore(lwpreflight): add verbose writer to show preflight progress messages

PengyuanZhao:
## Summary

Add `verbosewriter` to print preflight progress messages. It can be overwritten so that the consumer of `lwpreflight` can relay the progress messages to another writer.

Screenshot 2025-06-19 at 6 06 10 PM


lokesh-vadlamudi (approved):
LGTM :-)

PR #1738, opened 2025-07-01
chore: check wildcard permissions in Azure preflight package

lokesh-vadlamudi:
## Summary

Adding wildcard permissions check in azure preflight package, also fine tuning admin check during initialization.

Using multiple Service principles with different custom role and owner RBAC role. 

Test case 1:
SP with wildcard permissions in the custom role. - passed
 
Test case 2:
SP with owner role assigned - The IsAdmin is set to true. Passed
Screenshot 2025-07-01 at 2 44 16 PM

Test case 3:
SP with regular permissions instead of wildcard permissions.  - Passed

### DIFF
--- a/lwpreflight/azure/azure.go
+++ b/lwpreflight/azure/azure.go
@@ -16,10 +16,11 @@ type azureConfig struct {
 }
 
 type Preflight struct {
-	azureConfig      azureConfig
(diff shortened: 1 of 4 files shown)


PengyuanZhao (approved):
LGTM.

PR #1744, opened 2025-07-21
feat(lwpreflight): support org level checks

PengyuanZhao:
## Summary

- Add org level permissions for AWS and GCP
- Add org policy constraint check for GCP
  - GCP Organizations can optionally disable the creation of service account keys. If the `disableServiceAccountKeyCreation ` policy is enabled, our integrations will fail.

Screenshot 2025-07-21 at 2 06 58 PM


lokesh-vadlamudi (approved):
LGTM :-)

PR #1844, opened 2026-07-14
feat: Add lacework cloud-account backup/delete --bulk/restore/cleanup commands

PengyuanZhao:
## Jira/Github ticket

https://lacework.atlassian.net/browse/LINK-4430

## Summary

Adds four composable `lacework cloud-account` subcommands for backing up, bulk-deleting, restoring, and cleaning up cloud account integrations. These are general-purpose primitives (keyed off the existing `--type` registry, so they work for any cloud-account type); together they support migrating from many single-account AWS Config (`AwsCfg`) integrations to a single org-level integration, but they're reusable beyond that scenario.

## New / changed commands

- **`cloud-account backup -t <Type> [--account-ids | --account-ids-file] [-f <file>]`**
  Writes the integrations of a given type to a local JSON file. For AWS types, `--account-ids` filters by the account parsed from each role ARN. Prompts for a save directory when `-f` isn't given (auto-generates the filename); non-interactive runs use the current directory.
- **`cloud-account delete --bulk --file <backup> [--dry-run]`**
  Extends the existing `delete` command to remove every integration listed in a backup file. `delete <guid>` still works.
- **`cloud-account restore --file <backup> [--dry-run]`**
  Re-creates the integrations from a backup (undoes `delete --bulk`). Because Create is server-assigned, restored integrations get **new** GUIDs — the command prints an old→new GUID mapping table.
- **`cloud-account cleanup --file <backup> [-t <Type>] [--dry-run] [--aws-profile] [--aws-region] [--assume-role-name] [--no-assume-role]`**
  Deletes the AWS resources a deleted AWS Config integration leaves behind: the IAM role and its customer-managed/inline policies (AWS-managed policies such as `SecurityAudit` are only detached, never deleted), and the owning CloudFormation stack.

## `cleanup` details

- **Credentials:** assumes `--assume-role-name` (default `OrganizationAccountAccessRole`) into each target account. Auto-detects when the base credentials are already in the target account (e.g. the management account's own integration) and uses them directly; `--no-assume-role` forces direct credentials.
- **CloudFormation stack deletion:** locates the owning stack via `DescribeStackResources(PhysicalResourceId=<role>)`, searching the account's regions (base/`--aws-region` first, then the rest) since CloudFormation is regional. Deletes the stack in its own region and waits for completion. IAM cleanup and stack deletion are independent — a failure of one does not block the other.
- **Safety:** the role is taken from each backup record's role ARN (no name guessing); everything is idempotent (already-deleted resources are treated as done), non-CFN roles get IAM-only cleanup, and per-account failures are reported without aborting the run.

## Suggested migration workflow

1. `cloud-account backup -t AwsCfg --account-ids-file accounts.txt`
2. `cloud-account delete --bulk --file <backup>`
3. Deploy the org integration with CloudFormation config_org template
4. Verify org coverage; if something's wrong, `cloud-account restore --file <backup>`
5. `cloud-account cleanup --file <backup>`

> Note: `restore` depends on the old AWS IAM role still existing (Create triggers live role validation), so run `cleanup` **last**.

## Implementation notes

- All commands default to a dry-run/confirmation-guarded flow; destructive actions require explicit `--apply`/confirmation and honor `--noninteractive` and `--json`.
- Adds the `github.com/aws/aws-sdk-go-v2/service/cloudformation` dependency (vendored).

Manually test all new commands.


lokesh-vadlamudi (approved):
LGTM

PR #1855, opened 2026-08-13
feat(CAD-2252): target both OUs and single accounts for AWS org-level Agentless

PengyuanZhao:
### Linked JIRA issue(s) - Required
https://lacework.atlassian.net/browse/CAD-2252 (Mantis 1279468)

### Description

An org-level Agentless integration can monitor the organization root, OUs, or individual accounts. Individual accounts were rejected: `Validate()` required a profile/region for every 12-digit ID, and `createAgentless` dropped anything not `ou-`/`r-` prefixed from the StackSet deployment targets.

That rule exists because the **CLI** provisions each monitored account's snapshot role through a per-account provider alias, which needs credentials for every account. A `SERVICE_MANAGED` **StackSet does not** — it can narrow an ancestor OU to a specific account list via `account_filter_type`, and the organization root is an ancestor of every account. So callers that can supply the root (self-deployment, which holds only org-level credentials) can now target individual accounts without per-account credentials.

Note `organization.monitored_accounts` already accepted all three forms — the [module's own validation](https://github.com/lacework/terraform-aws-agentless-scanning/blob/main/variables.tf) is `^ou-…$|^[0-9]{8,32}$|^r-…$`. It is forwarded to the Lacework SaaS integration to say *what to scan* and needs no credentials. Only snapshot-role provisioning was the blocker.

### Changes

**`lwgenerate/aws`**
- Add `AgentlessOrganizationRootID` + `WithAgentlessOrganizationRootID`.
- Relax the profile/region requirement **only** when that root is set, so existing CLI callers keep identical behavior and the identical error string.
- Emit a second stack set instance `snapshot_role_accounts` targeting the root with an `INTERSECTION` account filter. When OUs are also selected, the first instance gains a `DIFFERENCE` filter so the two target provably disjoint sets, and `depends_on` serializes them.

**`lwpreflight/aws`**
- Discover root/OU IDs for Agentless, not just Config. The permissions (`organizations:ListRoots`, `ListOrganizationalUnitsForParent`) were already in `RequiredPermissionsForOrg[Agentless]` and already granted by the customer-facing policy JSON — no policy change needed.
- Paginate `ListAccounts` and `ListOrganizationalUnitsForParent`, which silently truncated at 20 entries.

### Why `depends_on` is load-bearing

CloudFormation permits one operation per stack set at a time and the AWS provider does **not** retry `OperationInProgressException` on create. Without it the mixed case fails nearly every time, not intermittently.

### Backwards compatibility

Output is unchanged for every existing caller:
- `TestGenerationAgentlessOrganization` and its golden are **untouched** and still pass.
- `integration/aws_generation_test.go` is untouched.
- `TestGenerationAgentlessOrganizationRootNoChurnForOrgUnitTargets` generates the OU-only case twice, with and without the root, and asserts the outputs are **byte-identical** — this is what protects existing deployments, whose `deployment_targets` are all `ForceNew`.

New coverage: full goldens for the mixed and accounts-only cases, plus a five-case validation table.

### How to test

`go test ./lwgenerate/aws/... ./lwpreflight/...` — all green, along with `gofmt`, `go build ./...` and `go vet`.

Consumed by self-deployment (services) and the Automated Configuration wizard (rainbow) in follow-up PRs under the same ticket.

### DIFF
--- a/lwgenerate/aws/aws.go
+++ b/lwgenerate/aws/aws.go
@@ -1349,57 +1349,10 @@ func createAgentless(args *GenerateAwsTfConfigurationArgs) ([]*hclwrite.Block, e
 			blocks = append(blocks, monitoredModule)
 		}
 
-		autoDeploymentBlock, err := lwgenerate.HclCreateGenericBlock(
-			"auto_deployment",
-			nil,
-			map[string]interface{}{"enabled": true, "retain_stacks_on_account_removal": false},
-		)
-		if err != nil {
-			return nil, err
-		}
-		lifecycleBlock, err := lwgenerate.HclCreateGenericBlock(
-			"lifecycle",
-			nil,
-			map[string]interface{}{
-				"ignore_changes": lwgenerate.CreateSimpleTraversal([]string{"[administration_role_arn]"}),
-			},
-		)
-		if err != nil {
-			return nil, err
-		}
-
-		stacksetResource, err := lwgenerate.NewResource(
-			"aws_cloudformation_stack_set",
-			"snapshot_role",
-			lwgenerate.HclResourceWithAttributesAndProviderDetails(
-				map[string]interface{}{
-					"capabilities":     lwgenerate.CreateSimpleTraversal([]string{"[\"CAPABILITY_NAMED_IAM\"]"}),
-					"description":      "Lacework AWS Agentless Workload Scanning Organization Roles",
-					"name":             "lacework-agentless-scanning-stackset",
-					"permission_model": "SERVICE_MANAGED",
-					"template_url": "https://agentless-workload-scanner.s3.amazonaws.com" +
-						"/cloudformation-lacework/latest/snapshot-role.json",
-					"parameters": lwgenerate.CreateMapTraversalTokens(map[string]string{
-						"ExternalId":         "module.lacework_aws_agentless_scanning_global.external_id",
-						"ECSTaskRoleArn":     "module.lacework_aws_agentless_scanning_global.agentless_scan_ecs_task_role_arn",
-						"ResourceNamePrefix": "module.lacework_aws_agentless_scanning_global.prefix",
-						"ResourceNameSuffix": "module.lacework_aws_agentless_scanning_global.suffix",
-					}),
-				},
-				[]string{"aws.main"},
-			),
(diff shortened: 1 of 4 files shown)


lwgenerate/aws/aws.go
```diff
@@ -340,7 +350,10 @@ func (args *GenerateAwsTfConfigurationArgs) Validate() error {
 			if len(args.AgentlessMonitoredAccountIDs) == 0 {
 				return errors.New("must specify monitored account ID list for Agentless organization integration")
 			}
-			if len(args.AgentlessMonitoredAccounts) == 0 {
```
lokesh-vadlamudi, line 343: nit: The existing check only looks at the AgentlessOrganizationRootID field, but "Entire organization" sends the root inside the MonitoredAccountIDs list instead — a different variable that check never reads.

So the new regex needs the r- branch, or it would reject every legitimate whole-org deployment while still letting a hostile r-…"], evil = ... string in the list flow straight into the generated HCL.


```
// Mirrors the monitored-account validation in the terraform-aws-agentless-scanning module and in
// the wizard (AWSMonitoredAccountRegex), so all three accept the same IDs.
     var agentlessOrgUnitIDFormat = regexp.MustCompile(`^(r-[0-9a-z]{4,32}|ou-[0-9a-z]{4,32}-[0-9a-z]{8,32})$`)      

      // OU and root entries reach the StackSet deployment_targets as raw HCL tokens whenever
      // one is present, with no dependency on AgentlessOrganizationRootID, so their format is
      // checked here rather than inside the root-only block below.
      for _, id := range args.AgentlessMonitoredAccountIDs {
        if isAgentlessOrgUnitID(id) && !agentlessOrgUnitIDFormat.MatchString(id) {
          return errors.New("monitored organizational unit IDs must be ou-*" +
            " and the organization root must be r-*")
        }
      }
```

lokesh-vadlamudi (approved):
Approving with a nit

lwgenerate/aws/aws.go
PengyuanZhao: Updated with https://github.com/lacework/go-sdk/pull/1855/changes/6b25c6b0dfaa6b253049b10e58fc1fffac786517

PR #1861, opened 2026-09-14
feat(api,lwgenerate): add FortiDSPM-managed DSPM accounts and modules

cstrace-ftnt:
## Jira/Github ticket

None. Part of the FortiDSPM ↔ FortiCNAPP integration (phase 2): FortiDSPM-managed
DSPM scan engines deployed from the Cloud Accounts wizard or from a customer's own
Terraform.

### What

- `api`: `AwsFortiDspmData` / `AzureFortiDspmData` for DSPM cloud accounts that
  FortiDSPM manages itself. They reuse the existing `AwsDspm` / `AzureDspm` cloud
  account types but carry only the account identity (AWS account id, or Azure
  tenant id and optional subscription id) and the regions. There is no customer
  bucket and no cross-account credentials: FortiDSPM pushes scan results to
  Lacework itself.
- `api`: the create response carries `fortiDspmDeployment`, the deployment
  api-server obtained from FortiDSPM: deployment id, token expiry and one
  `regions[]` entry per region with the single-use activation token and the image
  reference (AMI id on AWS; image VHD URL plus Hyper-V generation on Azure).
  `ActivationTokens()`, `Images()` and `HypervGenerations()` key these by region,
  which is the shape Terraform `for_each` needs at plan time.
- `api`: `ReportFortiDspmDeploymentStatus` (`PUT
  /api/v2/FortiDspm/deployments/{intgGuid}/status`) forwards the Terraform
  outcome for one integration to FortiDSPM through api-server.
- `lwgenerate`: `WithDspm` on the AWS and Azure generators emits one
  `terraform-{aws,azure}-fortidspm` module per region.

### How

The generated layout follows agentless scanning: the first region's module is
`global = true` and owns the `lacework_integration_{aws,azure}_fortidspm`
resource; the other regions take `global_module_reference` and read the
activation token and image for their region from it. AWS DSPM regions get their
own `aws` provider aliases. The module source is pinned to a git ref
(`?ref=v0.2.0`) until the modules are on the registry.

### Notable

- Existing `AwsDspm` / `AzureDspm` types, the `lacework_integration_*_dspm`
  resources and the agentless/config/cloudtrail generators are unchanged; all
  existing lwgenerate golden tests pass as they were.
- The new `lacework_integration_{aws,azure}_fortidspm` provider resources and the
  `v0.2.0` module tags are in follow-up PRs on terraform-provider-lacework and
  the two module repositories; this PR only has to merge and release first.

- `go test ./api/ ./lwgenerate/...`: new tests cover the AWS and Azure create
  request/response shape (including that no bucket field is sent), a read that
  has no `fortiDspmDeployment`, the status report request, and the generated HCL
  for a two-region AWS layout and a single-region Azure layout.
- End to end against a local api-server and self-deployment built from this
  branch: the Cloud Accounts wizard deployed an AWS scan engine in `us-west-2`
  that activated against the FortiDSPM dev environment, and an Azure scan engine
  in `westus2` built from the returned VHD URL. A customer-style `terraform
  apply` / `destroy` of the Azure module using these types also succeeded.


lwgenerate/aws/aws.go
```diff
@@ -330,9 +340,12 @@ func (args *GenerateAwsTfConfigurationArgs) IsEmpty() bool {
 
 // Ensure all combinations of inputs our valid for supported spec
 func (args *GenerateAwsTfConfigurationArgs) Validate() error {
-	if !args.Agentless && !args.Cloudtrail && !args.Config {
+	if !args.Agentless && !args.Cloudtrail && !args.Config && !args.Dspm {
 		return errors.New("Agentless, CloudTrail or Config integration must be enabled")
```
PengyuanZhao:
```suggestion
		return errors.New("Agentless, CloudTrail, Config or DSPM integration must be enabled")
```

lwgenerate/aws/aws.go
```diff
@@ -1245,6 +1277,65 @@ func isAgentlessOrgUnitID(id string) bool {
 	return strings.HasPrefix(id, "ou-") || strings.HasPrefix(id, "r-")
 }
 
+func dspmProviderAlias(region string) string {
+	return strings.ReplaceAll(region, "-", "_")
+}
+
+// DspmModuleName is the name of the FortiDSPM module generated for a region;
+// callers use it to address the module's outputs.
+func DspmModuleName(region string) string {
```
PengyuanZhao, line 1286: Should this be private? 


```suggestion
func dspmModuleName(region string) string {
```

lwgenerate/azure/azure.go
```diff
@@ -850,6 +884,63 @@ func createAgentless(args *GenerateAzureTfConfigurationArgs) ([]*hclwrite.Block,
 	return blocks, nil
 }
 
+// DspmModuleName is the name of the FortiDSPM module generated for a region;
+// callers use it to address the module's outputs.
+func DspmModuleName(region string) string {
```
PengyuanZhao, line 889: Should this be private?

```suggestion
func dspmModuleName(region string) string {
```

PengyuanZhao (approved):
LGTM except for a nit.

lwgenerate/aws/aws.go
cstrace-ftnt: Kept exported on purpose: self-deployment uses it to build the output traversal `module.<name>.lacework_integration_guid` for the first region's module. The name is derived from the region (`-` → `_` on AWS, lowercased with spaces removed on Azure), so exporting it keeps that normalization in one place instead of duplicating it in services. Happy to make it private if you'd rather services hardcode the name like it does for `aws_config` / `main_cloudtrail`.

lwgenerate/azure/azure.go
cstrace-ftnt: Kept exported on purpose: self-deployment uses it to build the output traversal `module.<name>.lacework_integration_guid` for the first region's module. The name is derived from the region (`-` → `_` on AWS, lowercased with spaces removed on Azure), so exporting it keeps that normalization in one place instead of duplicating it in services. Happy to make it private if you'd rather services hardcode the name like it does for `aws_config` / `main_cloudtrail`.

lwgenerate/aws/aws.go
PengyuanZhao: Got it. That makes sense.

PR #1862, opened 2026-09-14
fix(CAD-2352): stop creating VPC flow logs for org Agentless

PengyuanZhao:
### Linked JIRA issue(s) - Required
https://lacework.atlassian.net/browse/CAD-2352

### Description

An AWS **organization**-level Agentless deployment fails during `terraform apply`:

> `Error: creating Flow Log (...): api error 400: Access Denied for LogDestination: lacework-agentless-scanning-bucket-186cd561. Please check LogDestination permission.`

**Why it fails.** The scanning VPC's flow logs are written to the module's own S3 bucket, and that bucket's policy grants the log delivery service nothing — `terraform-aws-agentless-scanning/main.tf` has no `delivery.logs.amazonaws.com` principal and no `s3:GetBucketAcl` anywhere, only two `Principal: "*"` **Deny** statements. Per [AWS](https://docs.aws.amazon.com/vpc/latest/userguide/flow-logs-s3-permissions.html), the delivery grant is auto-attached only if the flow-log creator owns the bucket and can rewrite its policy — *"This policy overwrites any existing policy attached to the bucket"* — and otherwise *"flow log creation fails."* We do grant `s3:PutBucketPolicy`, so the outcome is environment- and ordering-dependent: it succeeds in some accounts and denies in others.

**Why only org.** Single-account deployments have not created flow logs since GROW-3001. That fix (`25d43ec1e`, #1684) set `use_aws_flow_log = false` on the two modules the **single-account** branch emits and left the **organization** branch untouched, where the module's `default = true` still applies. The org branch sets `regional = true` on both its global and per-region modules, and `aws_flow_log` is gated on `var.regional && var.use_aws_flow_log && !var.use_existing_vpc` — so every org deployment creates them.

### Changes

Set `use_aws_flow_log = false` on the org branch's global and per-region modules, carrying the same GROW-3001 comment the single-account branch has. The two snapshot-role modules are deliberately untouched: they pass neither `global` nor `regional`, so the module's own guard already prevents a flow log there.

These are `traffic_type = "REJECT"` logs on a VPC that exists solely to run the scanner, and single-account onboarding has gone without them since December 2024 — so this loses nothing a customer asked for and makes the two paths consistent.

### Testing

`go build`, `go vet`, `gofmt` and `go test ./lwgenerate/...` all clean.

- Three org goldens gain the attribute; the **single-account golden is untouched**, which is the proof the change is confined to the org branch.
- New `TestGenerationAgentlessOrganizationCreatesNoFlowLogs` asserts the org HCL disables flow logs on both modules and never enables them, matched by regexp since `hclwrite` pads `=` to the widest key per block. Confirmed it fails without the fix.

### Follow-ups (not this PR)

- **The module's bucket policy is the deeper problem.** Flow logs cannot work as shipped, and when AWS auto-attaches its grant it *overwrites* the policy — silently deleting the `ForceSSLOnlyAccess` and `ForceSSEOnlyUploads` deny statements. The module team should either grant delivery properly or default `use_aws_flow_log` to `false`.
- CAD-2352 also reports the **rollback** failing with `DependencyViolation` on the subnet; that is a separate defect in self-deployment's `stopAWSECSTasks` and is fixed in the paired services PR.


lokesh-vadlamudi (approved):
Looks good.

PR #1865, opened 2026-09-16
fix(CAD-2352): require ecs:DescribeTasks for Agentless preflight

PengyuanZhao:
### Description

Caught in review of [services#34837](https://github.com/lacework/services/pull/34837).

The Agentless rollback waits for Fargate tasks to reach `STOPPED` before terraform destroys the subnet their ENIs are in — otherwise the destroy fails with `DependencyViolation`. That waiter polls `DescribeTasks`, and `ecs:DescribeTasks` is in neither `RequiredPermissions[Agentless]` nor `RequiredPermissionsForOrg[Agentless]`. Neither list uses a wildcard, so the specific action has to be named.

The result today: a customer holding exactly the published `aws_agentless_policy.json` gets `AccessDenied` during the rollback — the point at which it matters most — and preflight told them nothing beforehand. The error is returned, not swallowed, so it fails the rollback.

The matching change to the two downloadable policy documents is [rainbow#24422](https://github.com/lacework/rainbow/pull/24408).

**Deliberately not added here:** `events:DisableRule` and `events:ListRuleNamesByTarget`, which [services#34904](https://github.com/lacework/services/pull/34904) uses to suppress the scan schedule before draining. That call is best effort and the deployment degrades without it, so making it a *blocking* preflight failure would reject deployments over a permission we already tolerate lacking. They are in the published policy so new customers get them.

### Tests and additional notes

`go build ./lwpreflight/...` and `go test ./lwpreflight/...` pass. Nothing pins these lists — no golden file, snapshot or test asserts their contents (I checked `permission_simulate_test.go`, `cli/cmd/preflight_test.go`, `integration/preflight_aws_test.go`, and the `preflight_aws` help golden), so this is the only change needed.

Note this is a **blocking** check: `CheckPermissions` appends to `p.errors`, rainbow gates the Deploy button on an empty error set, and the CLI exits non-zero. So customers whose policy lacks `ecs:DescribeTasks` will now be told at preflight instead of at rollback — which is the intent, but it is a behaviour change for them.

Drive-by: the non-org list had `events:TagResource` ahead of `events:RemoveTargets`, where the org list and both JSON documents are alphabetical.


lokesh-vadlamudi (approved):
LGTM

